PR #34, opened 2016-12-09
Change thank you text to bold and make the words thank you orange

### LINKED ISSUE #33: Change thank you text to bold and make the words thank you orange

jkup:
Separated the words Thank you into their own strings and removed the form element HTML, replacing it with an h4 and some text.

Fix #33

This is what it looks like now!
screen shot 2016-12-08 at 9 18 24 pm


bsclifton:
Matches spec, looks good :smile:

PR #35, opened 2016-12-11
removed txt option and new verified copy

### LINKED ISSUE #26: Remove text file option
We've decided to go DNS validation only.

### LINKED ISSUE #30: Update email verified page
We need to remove the web server option, add the new language that says wait times can be up to 3 days and then the skip the line copy before the DNS record button.

jkup:
Fix #26 fix #30

Auditors @ayumi @bsclifton 

I updated the email verified copy and removed (what I think) are all traces of the TXT verification now that we're not using it.

Would love if someone that knows rails a bit better could take a look and make sure I didn't do something stupid! :)

This is what it looks like now:

screen shot 2016-12-10 at 10 52 39 pm


bsclifton:
cc @bradleyrichter for the copy; how's it look?

bsclifton:
LGTM! :smile:

ayumi:
sweet!

PR #36, opened 2016-12-11
update dns copy

### LINKED ISSUE #29: Update copy on DNS page
There is new copy for the DNS page as well as some additional colors to communicate to the user that they need to interact with the fields before hitting verify.

jkup:
Fix #29

Auditors @ayumi @bsclifton 

Updated the copy on the DNS page to look like the spec. @bradleyrichter how does the design look?

screen shot 2016-12-10 at 11 43 24 pm


bsclifton:
@jkup can you rebase?

I had some nitpicks with the copy for @bradleyrichter; my suggestions (updated parts bolded):
- "Add a new **TXT** record **for your domain**" for the title (more accurate + has great Google results too, in case the user wanted to Google for help)
- Update the next part to be: "In order to verify your site ownership, you need to add a new DNS record to your **domain, which you can do using the account for your** DNS host"
- The customer could also put @ instead of specifying the domain name. I mention this because I could see them accidentally copying the whole field, including the "(Root record)".
- For "Name", maybe we could have alternatives in there too- like:
"Name **(or Host / Hostname):**"

Longer term, it would be cool to add a copy/paste button next to each field :smile:

bsclifton:
I went through several of the docs you provided, @bradleyrichter and did some touch ups on the words. Important things (in addition to the above) to call out:
- I replaced "your domain" with the actual domain :smile:
- mentioning the zone file (which is usually mentioned when trying to locate the host's DNS manager tool)
- Value field has "(or TXT Value / Data)" since it varies between hosts

Here's what it looks like now:
screen shot 2016-12-11 at 2 20 17 am

PR #37, opened 2016-12-12
DNS help popover

ayumi:
screen shot 2016-12-11 at 22 23 28

added [css-modal](https://github.com/drublic/css-modal) and restyled like a popover

@jkup is this helpful? it's not exactly bootstrap-popovers, but I think it could work for getting a dismissable info box.


jkup:
Awesome thanks! Gonna move the PR to my name so I can get all the feedback sent to me!

PR #39, opened 2016-12-12
New verification pending page

bsclifton:
Will fix https://github.com/brave/publishers/issues/31

Here's the new flow:

-----------

1) You click the big orange button:
screen shot 2016-12-14 at 2 47 19 am

-----------

2) It shows an in progress message (shows for at least 2 seconds before calling):
inprogress

-----------

3) If verification fails, a new screen is shown (which has the link back to the DNS verify page):
screen shot 2016-12-14 at 10 00 36 am

3a) If failure happened because ID was already taken, this error is shown (same as what was in alert bar)
screen shot 2016-12-14 at 9 53 32 am

3b) If failure happened because API had an outage or similar, this error is shown (same as what was in alert bar- with link to retry):
screen shot 2016-12-14 at 10 00 06 am


jkup:
We've gone back and forth on this a fair amount but I sort of think we should ship this as is. This approach works with or without JavaScript and I think it's probably best if we don't try to keep them on the page.

We might want to change the copy of the error message to be more like "Assuming you added the record we'll email you as soon as it's verified!"

What do you think @bradleyrichter @bsclifton?

bradleyrichter:
The conflicting messages here aren't going to work. 

For failure, we should have 2 screens. 

1. Verifying...please wait. (it needs to spin long enough to cover the time needed for instant verification. )

2a. Success!

2b. Verification could not be completed at this time. The Brave server will keep checking for up to 24 hours. You will receive an email notification after verification is complete, or has timed out. 

If you feel that you have reached this message in error, please edit your DNS record and try verification again. (verification is linked)

bsclifton:
PR updated!

- Clicking submit button: 
  - hides the DNS record fields
  - shows the "in progress" message; giving the appearance of a new page
  - uses setTimeout to submit the form (which is now a hidden form)

- Verify works almost the same
  - success goes to verified screen
  - failure goes to new failure screen

I know we need to get this out there and I did test this locally.  @bradleyrichter if it looks good, please go ahead and accept :smile: Merging will auto-deploy to production and will close things out :smile:

bsclifton:
Updated to remove alert bar; ready for re-review, @bradleyrichter

app/views/publishers/verification_dns_record.html.slim
```diff
@@ -1,8 +1,22 @@
 - content_for(:navbar_content) do
   span Step 2 of 4: #{I18n.t("publishers.new_header")}
 
+javascript:
+  function submit_for_verification () {
```
ayumi: minor: we should probably use camelCase for javascript

app/controllers/publishers_controller.rb
```diff
@@ -76,19 +73,26 @@ def verify
...
   rescue Faraday::Error
-    flash.now[:alert] = I18n.t("shared.api_error")
-    render(publisher_verification_action)
+    @try_again = true
+    @failure_reason = I18n.t("shared.api_error")
+    render(:verification_failed)
+  end
+
+  # Verification is still in progress
+  def verification_failed
```
ayumi: i noticed you render this action's view above.
can the user ever visit this action? otherwise you could remove this action, and still use the view above.

app/views/publishers/verification_dns_record.html.slim
```diff
@@ -1,8 +1,22 @@
...
+    var dns_records = document.getElementById('dns_records');
+    var verification_in_progress = document.getElementById('verification_in_progress');
+    var submit = document.getElementById('begin_verification');
+
+    dns_records.style.display = 'none';
+    verification_in_progress.style.display = 'block';
+
+    setTimeout(function () {
```
ayumi: does this verification page still work without JS?
i think it's important the site doesn't require JS.

bsclifton: Good catch :smile: I created an issue where this can be fixed: https://github.com/brave/publishers/issues/43

PR #49, opened 2016-12-23
Add back verification with a file (not shown to users)

bsclifton:
Add back verification with a file; it's not exposed though- requires you to manually visit the URL.

Fixes https://github.com/brave/publishers/issues/48

Auditors: @mrose17, @bradleyrichter


bsclifton:
Once this is deployed, you can tell folks to manually visit:
https://publishers.brave.com/publishers/verification_public_file

and they'll see the file option again:
screen shot 2016-12-23 at 12 33 14 pm

PR #69, opened 2017-02-24
Save LegalForm document fields to S3

### LINKED ISSUE #66: Save LegalForm Docusign field data
right now, the filled and signed filled PDF files are encrypted and stored in S3, which is good; however, it'd be helpful to have the actual fields available in encrypted form as well.

ayumi:
Also refactor Docusign envelope getter classes a little bit.

Fix #66


app/services/docusign_base_envelope_getter.rb
```diff
@@ -0,0 +1,33 @@
+# Get requests are rate limited, 1 per 15m per unique envelope endpoint:
+# https://docs.docusign.com/esign/guide/appendix/resource_limits.html
+# https://www.docusign.com/developer-center/api-overview#go-live
+# TODO: Use webhooks.
+class DocusignBaseEnvelopeGetter < DocusignBaseService
+  RATE_LIMIT_DELAY = 15.minutes
```
bsclifton: ❤️ how Ruby has syntax like this

ayumi:
@bsclifton good question about existing publishers! they will be okay.
- Their `form_fields_s3_key` field will be blank, but that's okay because it's only used right after docusign status validation to send a Slack notification.
- I'll make a follow-up task to backfill it on existing publishers.

A thing I did separately was to create updated Docusign templates, because existing template fields aren't named (e.g. "page1.text1"), and for downstream use they should be semantic (e.g. "residenceCountry").

bsclifton:
@ayumi awesome 😄 I'll leave the merging up to you, since it'll auto-deploy after merge and I you may want to manually test afterwards

Issue #51, opened 2016-12-27
Login tokens should expire


ayumi:
Every publisher has a secret URL which lets them manage their publisher registration. We added this to make it easy for publishers to bookmark and return to the management site.

[ThePythonicCow has noted in a Community post](https://community.brave.com/t/security-of-brave-publishers-payment-page-inadequate/378) that these private URLs don't expire, so if a third party discovers the link (for example, by forwarding a Publishers email, or posting it publicly) the publisher account would be compromised. This is a legit concern.

We'd like to:

0. Expire the secret URL token after use
*suggestion*: After verifying the login token and logging the publisher in, empty the token field.

1. Expire the secret URL token after some interval
*suggestion*: Add a new field, Publisher `#authentication_token_expires_at` and when we verify the token also verify the expires_at date.

2. For emails, ensure secret URL tokens are freshly generated right before each email is sent.

3. Add option to "Email me a login link"

@diracdeltas @mrose17

diracdeltas:
that sounds good. also in emails and on the website, it should have a disclaimer to never publish or share the token.

ThePythonicCow:
As noted above, I originated this concern a couple of months ago in a Community post.

"Secret" URL's are not an appropriate long term authentication mechanism, in my view, for any potentially high value resource.  They can be useful, for single-use, one-time, quickly expiring (say a day or less) checking that a newly provided email address works.  They can be useful for low value resources within controlled limitations, such as handing out read-only links to Amazon AWS S3 objects, which can be expired automatically, and which can be limited to the IP address of the user.

Some sort of real authentication method, classically involving at a minimum passwords that are never transmitted in the clear, never show plainly in a URL, and rarely sent via email, is needed to protect financial assets and identity (such as W-9 tax forms.)

Either I am seriously mistaken here, or else this "security" apparatus was never seen a competent security audit.  That seriously needs to happen.

PR #71, opened 2017-03-01
Expire login tokens; add email login link

ayumi:
Fix: #51

### Home
screen shot 2017-03-01 at 02 06 02

### Form
screen shot 2017-03-01 at 02 03 05

### Login email
screen shot 2017-03-01 at 02 02 29


config/locales/en.yml
```diff
@@ -73,6 +76,18 @@ en:
     log_out: "Exit"
     logged_out: "Your session has been cleared."
     new: "Submit contact information"
+    new_auth_token: "Log in with email"
+    new_auth_token_header: "Existing publisher login"
+    new_auth_token_body: "If you've already started or completed verification, we can send a login link to the email address associated with your publisher registration."
+    new_auth_token_body_2: "If you've forgotten the email:"
+    new_auth_token_body_3: "1) For verifications completed, leave email blank and we'll use the most recent email."
+    new_auth_token_body_4: "2) For verifications in progress, please "
```
bsclifton: While the programmer in me loves how the two entries above line up (For verifications X, Y) this sounds a little unnatural- maybe it could be:
`1) If you've already completed the verification process, you can leave email blank and we'll use the most recent one on file`
`2) If you were still working through verification, please include email or `
`start a new verification`

bsclifton: With the comment above too, I think it's worth suggesting (as I did in the text) that you can provide the email and domain if the person has already started (but not finished) the process... BUT they can just start the process over again 😄   The wording makes the email field look useless since it's telling you to:
- not put email if finished
- if not finished, don't even use this- start the process over

app/services/publisher_login_link_emailer.rb
```diff
@@ -0,0 +1,65 @@
...
+        return false
+      end
+    end
+
+    publishers_not_verified = Publisher.where(
+      brave_publisher_id: brave_publisher_id,
+      verified: false
+    )
+    if publishers_not_verified.none?
+      @error = I18n.t("publishers.new_auth_token_publisher_not_found")
```
bsclifton: Is disclosing that a domain is registered or not OK?

For example, if I wanted to be a creep and see if nytimes.com, reddit.com, etc have already started the process, this would let me know if records for that domain were found.

I don't know if there's a clean work-around, other than perhaps not allowing the person to provide email and then providing a "Email was sent if any was associated to that domain. If you'd like to start the process over (with a different email), click here" kind of link

ayumi: This is a good thought; I considered this but decided it was okay because browser-laptop shows the verified icon next to domains. It's also simple to creep on domains there: open a new profile, enable Payments, turn down min visit criteria, then browse a few websites.
@mrose17 @BrendanEich

bsclifton: @ayumi that's true 😄  I guess the only difference may be that you can tell if someone started the process (versus the ones showing next to the site, which have finished). Looking back, I think it's OK

config/locales/en.yml
ayumi: ++ great suggestions

app/controllers/publishers_controller.rb
```diff
@@ -139,15 +173,34 @@ def authenticate_via_token
...
 
+  # Used by #create_auth_token
+  # Kinda copied from Publisher #normalize_brave_publisher_id
+  def normalize_brave_publisher_id(brave_publisher_id)
+    require "faraday"
+    PublisherDomainNormalizer.new(domain: brave_publisher_id).perform
+  rescue PublisherDomainNormalizer::DomainExclusionError
+    "#{I18n.t('activerecord.errors.models.publisher.attributes.brave_publisher_id.exclusion_list_error')} #{Rails.application.secrets[:support_email]}"
+  rescue PublisherDomainNormalizer::OfflineNormalizationError => e
+    e.message
```
ayumi: if these errors are uncaught it will result in a 500, for which there's a [catch-all error page](https://publishers.brave.com/blabla404).

ayumi:
@bsclifton thanks for the feedback! i've addressed the comments.

separately i'll fix the dev localhost:3000 vs :5000 and add CI to auto run tests.

bsclifton:
cc: @ThePythonicCow

PR #79, opened 2017-04-13
Allow publishers to make their verification status public (opt-in)

bsclifton:
Add checkbox to setup process allowing publishers to opt-in to making their verification status public

Fixes https://github.com/brave/publishers/issues/68

Auditors: @ayumi


bsclifton:
I merged to staging- we'll just need to run migrations there and then we should be able to test 😄

PR #95, opened 2017-08-18
Allow publishers to link accounts to Uphold

dgeb:
This replaces the usage of Bitcoin addresses with linkage to Uphold accounts.

During publisher onboarding, the following steps are now taken:

1. Instead of being asked to enter a Bitcoin address, publishers are now shown a "Connect with Uphold" button. This button links to an authorization endpoint at uphold.com, and includes a `state` token that `brave/publishers` uses to validate any response from Uphold.

2. Following the link to uphold.com brings users to an authorization page where they can either create an Uphold account or sign in to an existing account. 

3. Once a publisher has authorized an account to link with Brave, they should be redirected to `/publishers/uphold_verified`. The query parameters should include the `state` token passed to Uphold, as well as a fresh `code`. The `state` token received is compared with the one that was sent.

4. The `code` is immediately exchanged with Uphold using the new `ExchangeUpholdCodeForAccessTokenJob`, which is performed synchronously. 

5. On success, a publisher's account is assigned `uphold_access_parameters`. The `publishers/finished` page is rendered, indicating success to the user.  

6. The new `UploadUpholdAccessParametersJob` is kicked off to asynchronously send the `uphold_access_parameters` to the new v2 wallet endpoint at `brave/eyeshade`. Once this is complete, the new `uphold_verified` flag is set on the Publisher model, finishing the Uphold authorization and linking process.

The following environment variables are introduced:

* `UPHOLD_AUTHORIZATION_ENDPOINT` - The Uphold authorization endpoint. Should include the tokens `<UPHOLD_CLIENT_ID>`, `<UPHOLD_SCOPE>`  and `<STATE>`, which will be replaced at runtime.  
Example: `https://sandbox.uphold.com/authorize/<UPHOLD_CLIENT_ID>?scope=<UPHOLD_SCOPE>&intention=signup&state=<STATE>`
  
* `UPHOLD_CLIENT_ID` - The Uphold client ID.

* `UPHOLD_CLIENT_SECRET` - The Uphold client secret.

* `UPHOLD_API_URI` - The API endpoint for Uphold. Example: `https://api-sandbox.uphold.com`

* `UPHOLD_PROVIDER` - The domain for Uphold (passed to eyeshade). Example: `api-sandbox.uphold.com`

* `UPHOLD_SCOPE` - The scope for Uphold authorization. Example: `cards:read,user:read,transactions:transfer:application`

The following related work is beyond the scope of this PR, but should be tracked going forward:

* The exchange of an Uphold code for a token must occur within the five minute lifetime of the code. If this exchange is not immediately successful, the exchange could be repeated in the background while the user is returned to a page that polls for the successful completion of the task. We could allow X retries with a Y backoff period. This process will only work seamlessly with JS enabled.

* Eyeshade needs to be updated with a new v2 wallet endpoint that accepts the Uphold parameters. This endpoint should be: `/v2/publishers/#{publisher.brave_publisher_id}/wallet` and should accept a PUT request with the body `"{\"parameters\": #{publisher.uphold_access_parameters}, \"verificationId\": \"#{publisher.id}\"}"`

* How can publishers change their linked Uphold account?

* How are active publishers whose accounts are already linked to Bitcoin addresses asked to link their accounts to Uphold? 

* All code related to legal forms / Docusign / Bitcoin should be removed eventually.

* The onboarding "steps" displayed (e.g. 1 of 4) are no longer accurate and need to be re-examined.


app/jobs/upload_uphold_access_parameters_job.rb
```diff
@@ -0,0 +1,11 @@
+class UploadUpholdAccessParametersJob < ApplicationJob
```
bsclifton: Rails specific question: Is there a UI available for viewing job status? ex: job log, including params passed and status (did an exception occur, etc)

bsclifton:
Clicking the `Create an account` link unfortunately trashes the query string ☹️  Any way to keep the return URL in the query string? I'd imagine a lot of folks would go down this code path
screen shot 2017-08-21 at 10 37 37 pm

bsclifton:
Actually- it seems my last comment was a bit premature. Following the link takes you into a process that leaves you in uphold (ex: you have to verify email, etc)

Maybe we should have a confirmation for the user that lets them know what's about to happen

Something like:
"You're about to leave the Brave publishers website. If you already have an uphold account, you'll be returned. If you don't, you'll want to finish the sign up process and then return to publishers.brave.com again to continue"

bsclifton:
What name should be shown in the `authorize application` screen? Since we're just using the sandbox, it has a simple name... but we may want to consider something like "Brave Publishers system" that the user will recognize
screen shot 2017-08-21 at 10 49 13 pm

app/jobs/upload_uphold_access_parameters_job.rb
dgeb: @bsclifton the Sidekiq web interface is mounted at `/magic` by default in [routes.rb](https://github.com/brave/publishers/blob/master/config/routes.rb#L67)

app/controllers/publishers_controller.rb
```diff
@@ -129,6 +130,40 @@ def verify
   # Shown after verification is completed to encourage users to submit
   # payment information.
   def verification_done
+    @publisher = current_publisher
+    @publisher.uphold_state_token = SecureRandom.hex(64)
+    @publisher.save!
+  end
+
+  def uphold_verified
```
ayumi: Minor: Running in dev without HTTPS, when I get redirect here after Uphold oauth it's HTTP.

test/controllers/publishers_controller_test.rb
```diff
@@ -1,5 +1,6 @@
 require "test_helper"
 require "shared/mailer_test_helper"
+require 'webmock/minitest'
```
ayumi: Minor: this project prefers double quotes

app/views/publishers/home.html.slim
```diff
@@ -18,34 +18,15 @@
         - else
           div= I18n.t("publishers.status_to_receive_payments")
           ul
-            - if current_publisher.bitcoin_address.blank?
-              li= I18n.t("publishers.status_bitcoin_address_required")
-            - if !current_publisher.legal_form_completed?
-              li= I18n.t("publishers.status_legal_form_required")
+            - if current_publisher.uphold_status != :verified #ToDo: Further breakdown
+              li= I18n.t("publishers.status_connect_to_uphold_required")
```
ayumi: Could we make this link, for example to the publisher uphold authorize path? It would help clarify where to proceed.

app/services/publisher_wallet_setter.rb
```diff
@@ -2,32 +2,28 @@
 class PublisherWalletSetter < BaseApiClient
```
ayumi: Minor: This job doesn't directly set the wallet anymore, so maybe we should add a comment up top clarifying that it sets uphold params.

dgeb: Good point - I will correct the comment at the top. We should consider renaming the eyeshade endpoint as well if the name "wallet" is no longer applicable.

app/controllers/publishers_controller.rb
dgeb: It is pretty inelegant to have to modify the redirect URL manually. However, Uphold requires that the redirect URL be HTTPS. One solution would be to locally run a proxy, but that's a hassle in its own way.

dgeb:
@ayumi @bsclifton @mrose17 Thanks for your reviews. Between our discussion yesterday and some tweaks today, I _think_ I've taken care of the blockers. 

> I think there's 1 blocker, which is the manual DNS verification page not hitting Publishers#verification_done to update the token, leading to a 500 – more details in the specific comment.

@ayumi Please see [this commit](https://github.com/brave/publishers/pull/95/commits/5cdc3c736d2a7d447fc818beb42f8c0bbe6b3b16) to confirm that this has been addressed.

There are some points that definitely deserve more discussion, such as:

> Could we make this link, for example to the publisher uphold authorize path? It would help clarify where to proceed.

There are a couple states in which that link will be shown, and we may want to provide actions to handle each separately. If this isn't a blocker, then I'd suggest we flesh this out more in a separate issue.

app/controllers/publishers_controller.rb
```diff
@@ -131,8 +131,9 @@ def verify
   # payment information.
   def verification_done
     @publisher = current_publisher
-    @publisher.uphold_state_token = SecureRandom.hex(64)
-    @publisher.save!
+    if @publisher.uphold_state_token.nil?
```
ayumi: I think since we moved this into PublisherVerifier this bit is no longer needed

dgeb: @ayumi I'm not attached to this code, but we need some way to initialize the `uphold_state_token` for publishers who are already in the system, have already been verified, yet still need to connect with Uphold. This is related to the discussion we need to have about allowing publishers to switch Uphold accounts (and probably with newly generated state tokens).

dgeb:
I've just rebased and pushed the following changes:

* The `UPHOLD_SCOPE` env var example is now set to `cards:read,user:read,transactions:transfer:others` 

* The `UPHOLD_PROVIDER` env var has been introduced. This domain is passed to eyeshade as the value for `provider` during the uphold upload (i.e. wallet setter) process. For dev, we should use a value of `api-sandbox.uphold.com`

* The `Publisher#generate_uphold_state_token` method has been refactored as `prepare_uphold_state_token`. This method will not regenerate a token if one is already set (which is consistent with its only usages).

PR #96, opened 2017-08-22
Introduce API endpoint to delete publishers

### LINKED ISSUE #47: API endpoint to delete publishers
Will be used by an admin to delete junk publishers.

@mrose17

dgeb:
[Resolves #47]

### DIFF
--- a/app/controllers/api/publishers_controller.rb
+++ b/app/controllers/api/publishers_controller.rb
@@ -49,6 +49,11 @@ class Api::PublishersController < Api::BaseController
     end
   end
 
+  def destroy
+    Publisher.where(brave_publisher_id: params[:brave_publisher_id], verified: false).destroy_all
+    render(json: { message: "success" })
+  end
+
   private
 
   def publisher_create_params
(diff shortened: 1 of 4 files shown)


app/controllers/api/publishers_controller.rb
```diff
@@ -65,6 +75,16 @@ def require_valid_legal_form
     end
   end
 
+  def require_publisher
+    @publisher = Publisher.find_by(brave_publisher_id: params[:brave_publisher_id])
```
ayumi: There can be multiple publishers per `brave_publisher_id` which have `verified: false`. For example, many curious people have tried to register `google.com`, creating many unverified Publisher records. The only constraint is there can only be 1 verified publisher.

Depending on what @mrose17 feels I think this should destroy all Publisher records matching the brave_publisher_id.

mrose17: @ayumi - agreed. get rid of all of them!

dgeb: Ok - will do 👍

dgeb:
The `destroy` endpoint now destroys all publishers that match the specified `brave_publisher_id` AND have `verified: false`. `200` is returned regardless of whether any publishers actually match these criteria.

One edge case to consider is that the `verified` column in the `publishers` table does not explicitly disallow `NULL` values. Should we introduce a migration to add that constraint?

ayumi:
@dgeb migration to add Publisher #verified NOT NULL sounds all right to me. The column defaults to `false` so it's not a problem, but it'd be a good practice thing.

Issue #53, opened 2017-01-03
Protocol on domain leads to error


jonathansampson:
While attempting to register a domain, one user reported that the presence of a protocol in the domain field yields a somewhat unhelpful error message. Perhaps we should detect protocols, in addition to having the example format in the input element's placeholder.

Report: https://twitter.com/OmgImAlexis/status/816163321711710208

mrose17:
cf., https://github.com/brave/ledger-publisher#syntax for the syntax of a publisher identity

dgeb:
Stripping the protocol could be done either entirely at the `ledger` endpoint (`/v2/publisher/identity`) or through a normalization step prior to sending the domain to `ledger`.

See the ["ledger offline"](https://github.com/brave/publishers/blob/master/app/services/publisher_domain_normalizer.rb#L30) code for an example of the latter.

mrose17:
@dgeb - please do the latter... what gets sent to the ledger/eyeshade should conform to the syntax above...

dgeb:
Resolved via #100.

PR #98, opened 2017-08-25
Email brave_publisher_id as link

### LINKED ISSUE #46: the publisherID in all e-mails should to be in HTML
see the two images below. the first one contains the email address as text, and the mail reader recognizes it as a URL and underlines it. the second one contains the email address as markup.

the issue is that some mail readers don't recognize all TLDs, so in some cases, the first one doesn't have the address automatically made into a link.

in all cases, for all messages, the publisherID should be a link, not a string

lgebhardt:
Updates the uphold_account_changed and welcome emails to use HTML links to the `brave_publisher_id` .

Adds email previews for all emails in the system to aid in testing. These are not necessary, but they are helpful for quickly viewing the emails. As part of the rails testing framework they will not be available in production. 

I made an assumption that the link should use `http` and not `https` in order to match the probable behavior of the email clients' rendering of found links (for known top level domains).

There are a few emails that appear to not be used any more. I don't see an issue filed to remove them, though it should be considered.

Fixes #46

### DIFF
--- a/test/helpers/publishers_helper_test.rb
+++ b/test/helpers/publishers_helper_test.rb
@@ -1,0 +1,9 @@
+require 'test_helper'
+
+class PublishersHelperTest < ActionView::TestCase
+  test "should render brave publisher id as a link" do
+    publisher = publishers(:default)
+    assert_dom_equal %{<a href="http://#{publisher.brave_publisher_id}">#{publisher.brave_publisher_id}</a>},
+                     link_to_brave_publisher_id(publisher)
+  end
+end
--- a/app/helpers/publishers_helper.rb
+++ b/app/helpers/publishers_helper.rb
@@ -16,6 +16,11 @@ module PublishersHelper
     "https://#{publisher.brave_publisher_id}"
   end
 
+  def link_to_brave_publisher_id(publisher)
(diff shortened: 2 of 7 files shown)


test/helpers/publishers_helper_test.rb
```diff
@@ -0,0 +1,9 @@
+require 'test_helper'
+
+class PublishersHelperTest < ActionView::TestCase
+  test "should render brave publisher id as a link" do
+    publisher = publishers(:default)
+    assert_dom_equal %{<a href="http://#{publisher.brave_publisher_id}">#{publisher.brave_publisher_id}</a>},
```
ayumi: just a minor note, wish we could know when it's https:// vs not so we could show https:// when appropriate

lgebhardt: The protocol will sometimes be known (as we're fixing in #53). We could capture it when it's provided. Is that something that is worth the effort, given it won't be 100% (or would require the entry of a full url)? Even with the protocol we might be missing the other parts of a complete and resolvable URL like the hostname, path, and parameters.

ayumi: you are correct it's probably not worth the effort

PR #99, opened 2017-08-28
Add locale string for `phone_normalized`

lgebhardt:
Adds a locale string for `phone_normalized` so the validation error field matches the display name for the field.

```
Phone normalized: is an invalid number
```
I encountered this field name mismatch on the `/publishers/new` endpoint. I hope this is a desirable change.


ayumi:
works great! thanks :cactus:

PR #101, opened 2017-08-29
Fix flow for returning publishers who've completed Uphold verification

lgebhardt:
Publishers were being directed to the `/publishers/verification_done` endpoint. After this fix they are directed to the `/publishers/home` endpoint.

### DIFF
--- a/app/helpers/publishers_helper.rb
+++ b/app/helpers/publishers_helper.rb
@@ -49,12 +49,14 @@ module PublishersHelper
   end
 
   def publisher_next_step_path(publisher)
+    return home_publishers_path if publisher.uphold_complete?
+
     return verification_publishers_path if !publisher.verified?
+
     # ToDo: Polling page for exchanging uphold_code for uphold_access_parameters
     # return authorize_uphold_path if publisher.uphold_code && publisher.uphold_access_parameters.blank?
-    return verification_done_publishers_path if publisher.uphold_access_parameters.blank?
 
-    home_publishers_path
+    verification_done_publishers_path
   end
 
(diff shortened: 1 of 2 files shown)


app/helpers/publishers_helper.rb
```diff
@@ -49,12 +49,13 @@ def publisher_verification_file_url(publisher)
...
   def publisher_next_step_path(publisher)
+    return home_publishers_path if publisher.uphold_complete?
+
     return verification_publishers_path if !publisher.verified?
+
     # ToDo: Polling page for exchanging uphold_code for uphold_access_parameters
     # return authorize_uphold_path if publisher.uphold_code && publisher.uphold_access_parameters.blank?
     return verification_done_publishers_path if publisher.uphold_access_parameters.blank?
-
-    home_publishers_path
```
dgeb: It seems like this method should still have a default path to fallback to.

lgebhardt:
@dgeb Thanks, good catch. Fixed it so the `verification_done_publishers_path` is now the default since we exit early if the uphold path is complete and if the publisher isn't verified.

app/models/publisher.rb
```diff
@@ -84,6 +84,10 @@ def verify_uphold
     save!
   end
 
+  def uphold_complete?
+    self.uphold_verified || !self.uphold_access_parameters.blank?
```
ayumi: minor: could the second condition also be `self.uphold_access_parameters.present?`

lgebhardt: Good catch. I'll fix that up.

lgebhardt:
@ayumi all set

PR #102, opened 2017-08-29
Refactor JSON sent to wallet endpoint

dgeb:
These changes are based on changes requested by @mrose17.

A `server` member has been added inside the `parameters` that come from Uphold.

The `UPHOLD_PROVIDER` env var is no longer needed, since we’re always going to just send `uphold` for this value.


mrose17:
LGTM.

PR #103, opened 2017-08-29
Remove bitcoin and legal forms

lgebhardt:
This cleans up the now unused PublisherLegalForms and Bitcoin related components. There's a lot of changes, almost all deletions. I've tried to group the changes by functionality into two commits, in the hopes that it makes it easier to review. There is a third commit, "Filter uphold_code and uphold_access_parameters from logs", which could be a standalone PR. I left it in here to avoid merge conflicts since we're also removing a bitcoin reference on the same line.

**Note**: Data is being removed from the database by dropping the PublisherLegalForms table and the encrypted bitcoin column from the Publishers table. If this data isn't stored elsewhere I can rework the migrations and/or code to preserve this, if needed.


app/controllers/api/publishers_controller.rb
```diff
@@ -1,8 +1,4 @@
 class Api::PublishersController < Api::BaseController
-  before_action :require_verified_publisher,
-    only: %i(notify update_legal_form)
```
ayumi: should this filter still apply to `#notify`?

config/locales/en.yml
```diff
@@ -205,7 +195,7 @@ en:
     save_and_continue: "Save and continue"
   static:
     landing_header: "Welcome to Brave Payments"
-    landing_body: "Brave users are already making contributions to thank you for your efforts. The first step toward collecting those contributions is to verify ownership of your web address so Brave users can see that your site is verified. Once fully verified, you will begin to receive daily micropayments in your Bitcoin wallet."
+    landing_body: "Brave users are already making contributions to thank you for your efforts. The first step toward collecting those contributions is to verify ownership of your web address so Brave users can see that your site is verified. Once fully verified, you will begin to receive daily micropayments in your Uphold account."
```
ayumi: Maybe we should prefer "Basic Attention Token wallet" – cc @mrose17

app/controllers/api/publishers_controller.rb
lgebhardt: Sorry, this one slipped by.

lgebhardt: @ayumi ok, fixed

config/locales/en.yml
ayumi: Tracking this at https://github.com/brave/publishers/issues/104

mrose17: @ayumi - agreed. i think we can use 'BAT wallet"

PR #105, opened 2017-08-30
Allow DNS verify form to work with JS disabled

### LINKED ISSUE #43: DNS verify form should work without JavaScript
Per feedback https://github.com/brave/publishers/pull/39#pullrequestreview-12979047

The "Verify DNS Record" button uses JavaScript to show an interstitial "verifying..." page while the call is in progress.  If folks have JavaScript disabled, this form submission will not work.

We need to add a `<noscript>` block on this page which falls back to the previous behavior if JavaScript is not enabled (show a button which blocks until call is done). The JS enabled button can be set to display:none and have it's display set to "block" on page_load().

dgeb:
The prior approach used a hidden form triggered by a `click` event on an
independent button. This did not work with JS disabled.

This approach converts the independent button to be a submit button for
a standard (non-hidden) form that will submit regardless.

The display of the interstitial should still happen with JS enabled.

Further exploration of an XHR-based alternative that uses polling is 
recommended.

[Resolves #43]

### DIFF
--- a/app/views/publishers/verification_dns_record.html.slim
+++ b/app/views/publishers/verification_dns_record.html.slim
@@ -2,18 +2,26 @@
   span Step 2 of 4: #{I18n.t("publishers.new_header")}
 
 javascript:
-  function submitForVerification () {
-    var dns_records = document.getElementById('dns_records');
-    var verification_in_progress = document.getElementById('verification_in_progress');
-    var submit = document.getElementById('begin_verification');
+  function submitForVerification() {
+    var dnsRecords = document.getElementById('dns_records');
+    var verificationInProgress = document.getElementById('verification_in_progress');
 
-    dns_records.style.display = 'none';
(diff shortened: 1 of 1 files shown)


app/views/publishers/verification_dns_record.html.slim
```diff
@@ -2,19 +2,27 @@
...
-      begin_verification.click();
+    setTimeout(function() {
+      var verifyForm = document.getElementById('verify_form');
+      verifyForm.submit();
     }, 2000);
   }
 
+  window.addEventListener('load', function() {
+    var verifyButton = document.getElementById('begin_verification');
+    verifyButton.addEventListener("click", function(event) {
```
ayumi: minor: single quotes

dgeb: Good catch! Just fixed.

PR #106, opened 2017-09-04
Use general publisher update endpoint for wallet updates

dgeb:
As per @mrose17’s [comments](https://github.com/brave/publishers/issues/74#issuecomment-326138778), wallet updates can now be made via `PUT /v1/publishers/{publisher}`. 

The `/wallet` suffix has been deprecated.


mrose17:
LGTM.

Issue #74, opened 2017-03-08
publisher dashboard should include link for contribution settlement statement


mrose17:
when a contribution settlement occurs, the publisher should receive an email and the publisher dashboard should contain a link to download the statement.

alexwykoff:
mr : @ayumi what is your take on the dashboard?

ay : i think we probably need only a single page with a few links. we probably don't need a framework per say. let's keep it simple & pragmatic

mr : i agree

bc : ditto

mr : i will close this and take action on #76 

to summarize : @mrose17 will handle

mrose17:
the "ask": when a verified publisher is at https://publishers.brave.com/publishers/home, they will see something like this:
![0](https://user-images.githubusercontent.com/400552/29220407-cc003edc-7e6f-11e7-882e-2e53207d5f75.png)
in addition to what we see here, there should be a form for "Publisher Statements". it should include a "Time range" drop-down with these choices:
- Past 7 days
- Past 30 days
- Month-to-date
- Past month
- Year-to-date
- Last year
- All

there should also be an indicator as to whether the publisher is on on the "auto-exclude" list, and if so, a button that will open up a mailto: to discuss that.

so, first step is for @bradleyrichter to decide on the UI... then i'll put together the API summary.

mrose17:
![0](https://user-images.githubusercontent.com/400552/29896472-fef9baa8-8d91-11e7-84be-349fc3a8123e.png)
![1](https://user-images.githubusercontent.com/400552/29896474-ff0291d2-8d91-11e7-8d2b-2e6d1b9dfc6c.png)
![2](https://user-images.githubusercontent.com/400552/29896473-fefb4724-8d91-11e7-8e73-8eb55d84d3aa.png)

mrose17:
API changes:

1. when calling `publisher_token_requester.rb::perform` include a `query` parameter of `?show_verification_status=true|false`

2. to update `show_verification_status` use `PUT /v1/publishers/{publisher}` with a payload of `{ "show_verification_status": true|false }`

dgeb:
@mrose17 A few questions:

> when calling `publisher_token_requester.rb::perform` include a `query` parameter of `?show_verification_status=true|false`

How does the additional query parameter affect the result? Currently just `{ token: value }` is returned. Will the addition of `?show_verification_status=true` also return a `show_verification_status` member as a sibling to `token`? Is `?show_verification_status=false` the same as not including any query params?

[...]

Issue #50, opened 2016-12-23
Add a route that presents both options (DNS & text file)


mrose17:
here is what the publisher sees after doing the email verification:
![0](https://cloud.githubusercontent.com/assets/400552/23190709/65d9790e-f84e-11e6-95be-05a93726752b.png)
which is nice, but we need to do a bit of fine tuning.

first, we should look at the SOA for the publisher domain to see who the DNS provider is. if we have an entry for the provider in the list of "easy to use DNS provider sites", then the text above should be "flipped" to ask them to try the DNS approach, and the big orange button should take them to a page with the current overview PLUS the provider-specific instructions.

second, if they can't figure out how to do it, then they should get a button that does a customized `mailto:` to zendesk.

alexwykoff:
aw : seems like creating the .well-known/text file would make sense 

mr : does this need more spec/sprucing?

bc : we can expose routes and show users the options

mr : is the wording sufficent ?

bc : explain what needs to be displayed on the screen

to summarize : we need specs to clarify the layout and copy @bradleyrichter @alexwykoff

bsclifton:
Here's the route for the text file option:
https://github.com/brave/publishers/blob/6fb3377aa1f8334f0897acb98d3cd49de0f1f776/config/routes.rb#L15

mrose17:
@bradleyrichter - before we can proceed on this, we need to review the wording.

@dgeb & @lgebhardt - when we start on this issue, the first thing is to restore that route and then @bradleyrichter can figure out what other changes we need to make...

alexwykoff:
Chatted with Brad and he mentioned a missing 3rd option which is the wordpress extension made by Marshall.

alexwykoff:
Following up on this, perhaps worthy of a separate issue, we should add detection on domain to know if they're using wordpress and offer the option conditionally.

alexwykoff:
Here are a few updated ideas. I'll try to present them in order as they would appear in a flow:

1. Updated landing page :

In this example, we're looking to do a blue background with a fair amount of transparency, background image tbd. User options on the right.

Most notably, the shift to requiring an email upfront for validation for new, unverified publishers.

<img width="962" alt="screen shot 2017-09-06 at 8 56 03 am" src="https://user-images.githubusercontent.com/1434040/30121732-486cd5fe-92e1-11e7-85ab-f73ed56259c7.png">

2. Notification of verification email sent : 

[...]

PR #109, opened 2017-09-13
Enable in-place edits of publisher from dashboard

dgeb:
This PR addresses the bulk of the requirements from #74. The exception is the generation of publisher statements, which can be performed with largely the same techniques employed here.

This PR enables a verified publisher to update their `name`, `email`, and `show_verification_status` preference from the dashboard.

Vanilla js to show in place forms, submit those forms via `fetch` (which is polyfilled), and show progress via a spinner (chosen by @alexwykoff).

All backend code is tested. If desired, we would add full-stack integration tests to test the client-side as well. To go this route, we may wish to update to Rails 5.1 and use the new Rails system tests built on Capybara.


ayumi:
A small potential problem is users' emails aren't verified through this new in-place edit flow. They might make a typo or enter a defunct email.

mrose17:
@dgeb - i think that updating the email should require another roundtrip. i really don't want to be left hanging ... is that doable.

also, a `<noscript>` tag sounds like a good idea!

dgeb:
Thanks for the review and feedback.

I can integrate a confirmation email into the process. I will coordinate this change with @lgebhardt, since his work in #50 is related.

Also, I will add a `<noscript>` tag that lets users know they'll need to enable JS to make edits.

Issue #110, opened 2017-09-14
Verify email changes


dgeb:
Emails can now be edited from the dashboard (see #109). Before these changes become effective, we should verify that the new email address is valid.

dgeb:
I'm guessing that we'll want to send two emails:

* one to the current address - notifying the user that an email change has been requested.

* one to the new address - requesting that the user follow a link to verify their new address.

@mrose17 @alexwykoff does that sound right? If so, I'll need text for both (although I can improvise for now).

mrose17:
yes, both addresses. @alexwykoff - please provide the copy. i suggest that the text for the initial  verification email be used as a starting point.

dgeb:
Resolved by #114

mrose17:
Fixed by #114

PR #114, opened 2017-09-21
Require verification of emails

dgeb:
This PR addresses three different issues:

* @lgebhardt has tweaked the signup flow to require just an email, as specified in #50. This involves a new layout of the home page and a new form for updating unverified publishers (i.e. adding a name, domain, etc.). Note that some visual elements still need tweaking on the home page.

* As per #110, whenever a user updates their email from the dashboard, they are required to verify it. Emails are sent to the old and new addresses. The dashboard handles pending email addresses via a special notice.

* CVE-2017-5029 has recently been issued against the Nokogiri gem. This vulnerability has been addressed in 1.7.2.

* CVE-2017-9050 for Nokogiri 1.7.2 requires an upgrade to >= 1.8.1, but this will involve some dependency conflict resolution. For now, the bundler vulnerability audit is skipped. A tracking issue will be opened to resolve this.

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -19,7 +19,9 @@ class PublishersController < ApplicationController
              new
              new_auth_token)
   before_action :require_unverified_publisher,
-    only: %i(verification
+    only: %i(email_verified
+             update_unverified
+             verification
              verification_dns_record
              verification_public_file
              verification_failed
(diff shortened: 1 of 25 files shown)


app/views/publisher_mailer/verify_email.html.slim
```diff
@@ -0,0 +1,21 @@
...
+- content_for(:title) do
+  h3= I18n.t("publisher_mailer.verify_email.title")
+
+.notice= I18n.t("publisher_mailer.verify_email.private_email_notice")
+
+p= I18n.t("publisher_mailer.verify_email.intro")
+
+p
+  strong.attribute= "#{Publisher.human_attribute_name("email")}:"
+  = @publisher.email
```
ayumi: Should this be `@publisher.pending_email`?

app/controllers/publishers_controller.rb
```diff
@@ -64,9 +62,30 @@ def create_done
 
   def update
     publisher = current_publisher
+    update_params = publisher_update_params
+
+    current_email = publisher.email
```
ayumi: minor: my intuition tells me this conditional email confirmation flow should be extracted into a service class

app/views/publisher_mailer/verify_email.html.slim
dgeb: Good catch! Fixed.

app/controllers/publishers_controller.rb
dgeb: It did become complex enough that I can see that a service would be warranted 👍

dgeb:
@ayumi thanks for your review! I fixed the minor blocker and would be glad to follow up with your other suggestions after this lands (since @lgebhardt needs to rebase his work sooner than later).

app/controllers/publishers_controller.rb
```diff
@@ -35,22 +37,18 @@ class PublishersController < ApplicationController
...
         verify_recaptcha(model: @publisher)
         : true
     if throttle_legit && @publisher.save
-      # TODO: Change to #deliver_later ?
-      PublisherMailer.welcome(@publisher).deliver_later!
-      PublisherMailer.welcome_internal(@publisher).deliver_later if PublisherMailer.should_send_internal_emails?
+      PublisherMailer.verify_email(@publisher).deliver_later!
+      PublisherMailer.verify_email_internal(@publisher).deliver_later if PublisherMailer.should_send_internal_emails?
+      session[:created_publisher_id] = @publisher.id
       session[:created_publisher_email] = @publisher.email
```
ayumi: sorry I didn't catch this – this should probably be `@publisher.pending_email`

feel free to update this in a small patch after

PR #121, opened 2017-09-22
Bring the landing page appearance in line with #50

lgebhardt:
Restructures the landing page content to place the buttons to the right of the input fields and text.

### DIFF
--- a/app/assets/stylesheets/application/static.sass
+++ b/app/assets/stylesheets/application/static.sass
@@ -1,6 +1,6 @@
 body[data-controller="static"]
   padding-top: $navbar-margin-bottom
-  background-color: rgba(145, 157, 190, 0.85)
+  background-color: rgba(darken($bravePurple, 6%), 0.85)
   color: white
   .centered
     text-align: center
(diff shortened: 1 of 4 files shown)


ayumi:
The left alignments look a little off:
screen shot 2017-09-23 at 00 15 48

lgebhardt:
@ayumi Fixed the alignment, email field width and rebased against staging to test with the larger logo.

@alexwykoff I should note that the button colors do not match the screenshot in the issue. Specifically the "Let's Begin" is set as the primary button (orange), and "Log in" is a secondary (white). In the screen shot the "Log In" button looks to be a secondary and the "Log in" appears to be rendered behind the overlay. Would you like  a new style for the "Log in" button, along with changing the "Let's Begin" to secondary?

alexwykoff:
If we look at the landing page in https://github.com/brave-intl/publishers/issues/50
 https://user-images.githubusercontent.com/1434040/30121732-486cd5fe-92e1-11e7-85ab-f73ed56259c7.png , the landing page shows that the log in button has no contrast and the 'Let's Begin' button is white.

lgebhardt:
@alexwykoff Ok, here's the latest.
screen shot 2017-09-25 at 1 08 03 pm

And mobile:
screen shot 2017-09-25 at 1 07 26 pm

alexwykoff:
Much better!  Are you using Arial or Open Sans?

lgebhardt:
It's using Lato, with fallback to Arial. I haven't changed the fonts from the current site, which is defined here: https://github.com/brave-intl/publishers/blob/master/app/assets/stylesheets/variables/fonts.scss

app/assets/stylesheets/application/static.sass
```diff
@@ -19,3 +19,30 @@ body[data-controller="static"]
...
     width: 128px
+  hr
+    border-color: $braveLightOrange
+  .form-control
+    background: $bravePurple
+    color: white
+  .control-label
+    width: 100%
+    color: white
+  .btn
```
ayumi: Once we standardize colors, I think it's important to move button styles into the shared button styles file to ensure we're consistent across the app. cc @alexwykoff

PR #124, opened 2017-09-25
only Publisher#generate_verification_token if brave_publisher_id present

### LINKED ISSUE #122: Publisher.generate_verification_token should only run after brave_publisher_id is set
right now it's running `after_create` since we changed signup to only require an email (rather than also brave_publisher_id) in https://github.com/brave-intl/publishers/pull/114

ayumi:
Fix #122

cc @dgeb @lgebhardt


ayumi:
tried it on staging

PR #126, opened 2017-09-27
New Verification Flows

lgebhardt:
This PR implements the remaining changes specified in #50.

As of now there are a few outstanding issues that will block this from being accepted:
- Missing the progress header contents. This will be added shortly.
- The graphics submitted should be reviewed and replaced with official ones

Needs clarification:
- Do we need checks to prevent a user from trying to verify a site if their site is not accessible or running HTTPS? The flow will only support enabling the verify process for the Support Queue and the DNS option. However a user would be able to work around this. What level of enforcement does publishers need here?


lgebhardt:
Some screen shots:

Step 2
step 2

Step 3
step 3

Step 4.a
step 4 a

Step 4.b
step 4 b

Step 4.b.II
step 4 b ii

Step 4.c
step 4 c

Step 4.d
step 4 d

Step 4.e
step 4 e

Step 4.f
step 4 f

Step 5
step 5

Step 6.a
step 6 a

dgeb:
This PR now also includes the progress header, which appears as follows:

screen shot 2017-09-26 at 5 24 44 pm

Each "progress circle" is drawn dynamically with an SVG and can support any fill percentage.

@alexwykoff and I have discussed centering the progress header and other page-specific content in the navbar in a separate PR.

mrose17:
let's fire this up!

PR #130, opened 2017-09-27
Move class from form to button

### LINKED ISSUE #129: Button in a button on `create_done`
Button inception:

screen shot 2017-09-27 at 3 23 26 pm

Looks to be button styles on the form.

lgebhardt:
Fixes #129


ayumi:
This page looks kind of awkward because the envelope is off center. This is better than what we had previously, but we should fix it.

screen shot 2017-09-27 at 21 26 32

dgeb:
@ayumi I also talked with @lgebhardt about this. He said this image and a few others were just cut out of the spec screenshots provided in #50. They are just serving as placeholders until the finals are provided (perhaps by @alexwykoff).

PR #163, opened 2017-10-12
Adds i8n-tasks development gem

lgebhardt:
Will allow management of localizations
Uses the default setup, which may need some customization for a best fit

Documentation can be found at https://glebm.github.io/i18n-tasks/
Along with a host of useful features, it can be used to find unused and missing keys:

```bash
i18n-tasks health
```

Might be useful to integrate with travis to ensure we don't launch with missing strings. This is TBD.


mrose17:
@lgebhardt - nice!
@ayumi - wdyt?

ayumi:
looks handy. thanks!

PR #169, opened 2017-10-12
Disable all scripts related to statement_period

dgeb:
Because the element was commented out, the event listeners could not be
added.


ayumi:
:tomato:

PR #174, opened 2017-10-16
added analytics for other verification methods, contact info added, c…

alexwykoff:
added analytics for other verification methods, contact info added, connect to uphold, and support queue

### DIFF
--- a/app/assets/javascripts/analytics.js
+++ b/app/assets/javascripts/analytics.js
@@ -15,27 +15,8 @@ if (document.head.dataset.piwikHost) {
     e = e || window.event;
     var target = e.target || e.srcElement;
 
-    // A switch would feel better, but would eval on target.dataset actually work?
-    if (target.dataset.trackWordpress) {
-      _paq.push(['trackEvent', 'WordpressVerificationClicked', 'Clicked', 'WordpressFlow']);
-    }else if (target.dataset.trackPublicFile) {
-      _paq.push(['trackEvent', 'PublicFileVerificationClicked', 'Clicked', 'PublicFileFlow']);
-    }else if (target.dataset.trackGithub) {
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/analytics.js
```diff
@@ -15,9 +15,27 @@ if (document.head.dataset.piwikHost) {
     e = e || window.event;
     var target = e.target || e.srcElement;
 
-    // Track if a user clicked on the verify button for a Wordpress flow
+    // A switch would feel better, but would eval on target.dataset actually work?
```
ayumi: yes I think a switch would be more legit

ayumi: another way you could do this is just to bind more data attributes, like one for `WordpressVerificationClicked`, one for `Clicked`, one for `WordpressFlow`. Then for all elements which have these data attributes, send the data the element contains.

ayumi: other nit is please don't use comments for code questions, in case we merge the code the funny comments will get stuck. it's better to leave a Github comment on the PR .

if u must write FIXME or XXX or HACK in the comment

app/assets/javascripts/analytics.js
```diff
@@ -15,9 +15,27 @@ if (document.head.dataset.piwikHost) {
...
     if (target.dataset.trackWordpress) {
       _paq.push(['trackEvent', 'WordpressVerificationClicked', 'Clicked', 'WordpressFlow']);
+    }else if (target.dataset.trackPublicFile) {
```
ayumi: `space after }`

mrose17: @alexwykoff ^^^

alexwykoff:
thanks y'all. will def update per feedback. looking to do data-track.id and switch on values.

alexwykoff:
sorry, missed a few easy errors. will push again w /fixes

PR #185, opened 2017-10-19
Tune up

lgebhardt:
Fixes style issues reported by Jenn:

- Buttons based on A tags no long show underlined style on hover
- Multiple font weights adjusted
- Minor spacing adjustments
- Info type pages are now slightly wider

Also, the Brave logo removed from "Now verifying your domain..." page. A spinner was asked for, but I have reservations about that issue since this page will never stop the spinner. It would be nice to have this page poll, with a spinner, for verification and then navigate to the dashboard when complete.

Still to do: look at installing additional font weights.


mrose17:
LGTM

PR #186, opened 2017-10-19
Install full set of Open Sans fonts

lgebhardt:
Update from v14 to v15
Not using the italics so they are not being loaded to improve load times


mrose17:
LGTM.

PR #205, opened 2017-10-24
Add statement generation

### LINKED ISSUE #74: publisher dashboard should include link for contribution settlement statement

dgeb:
Allows publishers connected to uphold to download statements from eyeshade.

The flow is as follows:

1. A publisher requests a statement for a given period from the dashboard.
2. A request is made to eyeshade's `/v1/publishers/#{publisher.brave_publisher_id}/statement` endpoint. A `reportURL` is returned.
3. The `reportURL` endpoint is polled until eyeshade returns the report as CSV data.
4. The report is saved to the `PublisherStatement` record's `contents`, an encrypted `text` field.
5. An email is sent out to notify the publisher that the statement is ready to view.
6. The dashboard continues to poll a `statement_ready` endpoint, which checks for the presence of `contents`. While polling, `Generating...` is shown with varying ellipsis.
    a. When polling succeeds, a `View statement` link is presented on the dashboard.
    b. If the statement is not ready when polling ends, a notice is displayed saying that an email will be sent out when the statement is ready.

Todo: Statements expire after 24 hours. We need a task to clear them out. Tracked via #206.

Closes #74

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,8 +7,6 @@ gem "active_model_serializers", "~> 0.10.0"
 # Encrypt DB data at rest
 gem "attr_encrypted", "~> 3.0.0"
 
-gem "aws-sdk", "~> 2", require: false
-
 gem "bootstrap-sass", "~> 3.3.6"
 
 # Authentication
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -50,12 +50,6 @@ GEM
       encryptor (~> 3.0.0)
     autoprefixer-rails (6.5.0.2)
       execjs
-    aws-sdk (2.6.10)
-      aws-sdk-resources (= 2.6.10)
-    aws-sdk-core (2.6.10)
-      jmespath (~> 1.0)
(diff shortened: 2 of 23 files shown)


dgeb:
Some followup questions:

* Should all non-expired statements be shown on the dashboard?

* After a statement has been generated, how should a user request another statement? Should selecting a new period immediately allow for a new statement to be generated? Currently, the user must refresh the page to request another statement, but this doesn't feel very dynamic.

* Is the following email text acceptable?

screen shot 2017-10-24 at 5 22 18 pm

mrose17:
answers:

- Should all non-expired statements be shown on the dashboard? **yes**
- After a statement has been generated, how should a user request another statement? **i would keep the button in the same place and just have the unexpired statements listed below the button
- Is the following email text acceptable? **i'll defer to @alexwykoff as to the template to use**

PR #220, opened 2017-10-27
Default currency

### LINKED ISSUE #172: Ability to set/change deposit currency (BAT conversion) on dashboard
Allow users (publishers) to define which currency they'd want the BAT to be converted and deposited in their Uphold wallet.

screen shot 2017-10-13 at 3 15 08 pm

lgebhardt:
Adds the default currency in the dashboard. This was implemented as a standard form select instead of a custom form connected to a change button for the sake of expediency. The setting is saved to the database when changed and the contribution balance and converted balance are updated.

Minor changes were made to the tests for robustness.

Fixes #172 

screen shot 2017-10-27 at 10 49 24 am


mrose17:
thanks! merging in process.

Issue #231, opened 2017-11-03
Dashboard JS error when Uphold has not been connected


dgeb:
When Uphold has not been connected, the `publisher_default_currency` element is not visible on the page, which causes a JS error that prevents some of the dashboard's dynamic functionality.

PR #233, opened 2017-11-03
Add support for new content providers

### LINKED ISSUE #225: Update publisher site to support creators
1.
Remove email form from landing page (though contrary to image, align 'get started' button right.
screen shot 2017-10-30 at 1 04 49 pm

2.
Give users a choice between traditional site flow and new Oauth flow. If they choose 'site operator' they go to 3, otherwise they go to 4.
screen shot 2017-10-30 at 1 05 53 pm

3. 
Update traditional flow to validate email after the user has selected that flow. Then send them to the contact information page and the rest of the flow.
screen shot 2017-10-30 at 3 06 59 pm

4. 
Send users to Oauth for the provider.

5. 
Post-Oauth, send users to Dashboard.

6. For users returning to log in, the page needs to supply an option for OAuth users.
screen shot 2017-10-31 at 2 37 16 pm

dgeb:
Introduces support for new creators in addition to site creators, as described in #225 (with the agreed upon amendment that sign up will still require email verification as a first step for either account type).

Pieces that are still a WIP:

* UI / UX from #225 should be considered preliminary.
* Dashboard appearance should also be considered preliminary.
* New services need tests.
* We're experiencing problems with the new `v1/owners` endpoint on eyeshade:
  - We need to sync up our phone number validation rules between publishers and eyeshade.
  - Running eyeshade locally on docker is returning a 404 for that endpoint
* The Google auth token has a limited lifetime (perhaps 1 hour). As a result, we're currently limiting the channel check to immediately post-auth. We should consider using refresh tokens and storing them in the DB (per Google's recommendations).

Closes #225, #231, #232

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


docs/publishers-secrets.example.sh
```diff
@@ -29,4 +29,13 @@ export UPHOLD_AUTHORIZATION_ENDPOINT="https://sandbox.uphold.com/authorize/<UPHO
...
 export UPHOLD_DASHBOARD_URL="https://sandbox.uphold.com/dashboard"
-export TERMS_OF_SERVICE_URL="https://basicattentiontoken.org/publisher-terms-of-service/"
\ No newline at end of file
+export TERMS_OF_SERVICE_URL="https://basicattentiontoken.org/publisher-terms-of-service/"
+
+# Get these from the Google Web application setup. During setup:
+# - make sure the authorized redirect url goes to the endpoint /publishers/auth/google_oauth2/callback
+export GOOGLE_CLIENT_ID=""
+export GOOGLE_PROJECT_ID=""
+export GOOGLE_AUTH_URI=""
```
ayumi: where do i find GOOGLE_AUTH_URI, GOOGLE_TOKEN_URI and GOOGLE_AUTH_PROVIDER_X509_CERT_URL

config/locales/en.yml
```diff
@@ -334,3 +339,8 @@ en:
     landing_login_body_html: "<strong>Already verified?</strong> Log in to see your publisher dashboard."
     landing_login_button: "Log in"
     landing_begin_button: "Get Started"
+  youtube:
+    authenticate_with_youtube: "Authenticate with YouTube"
+    account_not_found: "Account not found"
```
ayumi: This looks to be emitted when a matching Publisher isn't found. Could we change the copy to something like "Couldn't find a matching publisher record"?

dgeb:
@ayumi Thanks for your review. We've made the requested changes, including:

* @lgebhardt added Google API setup instructions to the README in [this commit](https://github.com/brave-intl/publishers/pull/233/commits/53a48cfd327c8236c8251aaaef153703a235c00f). 

* It turns out that a number of those environment vars were used during experiments but are no longer needed, so we just [removed them](https://github.com/brave-intl/publishers/pull/233/commits/9eabed28b4c0703e1d546187b631e4a0f02bcbd1). The only two that are currently used are `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET`, which are described in the setup instructions.

* I took your suggestion to add validations to the `Publisher` model in [this commit](https://github.com/brave-intl/publishers/pull/233/commits/8010d8baae04a8f0651725865b7fd992db11021c).

Please give it another look and let us know how it goes.

ayumi:
I finally understand the error I was getting previously. I had been trying to Login with OAuth **before** creating an account with OAuth! So as a minor UX improvement, when users try to OAuth login it could automagically create the Publisher account; or just say "No publisher registration found. Click 'Get started' to create an account."

PR #243, opened 2017-11-07
Balance to Wallet

### LINKED ISSUE #242: Update balance from eyeshade wallet instead of balance
Remove `balance` (`/v2/publishers/{brave_publisher_id}/balance`) and `legacy_balance` (`/v1/publishers/{brave_publisher_id}/balance`) api calls and replace with wallet calls:

`/v1/owners/{owner}/wallet`
or
`/v2/publishers/{owner}/wallet`

lgebhardt:
Resolves #242, #244

Tests are still needed for the WalletGetter and WalletSetter. PRing now to test issue where Owners Wallet does not contain the full data that Publishers Wallet does.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/models/publisher.rb
```diff
@@ -148,6 +137,20 @@ def inspect_brave_publisher_id
...
+      :site
+    elsif self.youtube_channel_id.present?
+      :youtube_channel
+    else
+      :unselected
+    end
+  end
+
+  def owner_identifier
+    "oauth#google:#{auth_user_id}"
```
ayumi: Should this return `nil` if auth_user_id is blank? for clarity that it requires oauth.

app/services/publisher_wallet_getter.rb
```diff
@@ -11,9 +11,16 @@ def initialize(publisher:)
...
-      request.headers["Authorization"] = api_authorization_header
-      request.url("/v2/publishers/#{publisher.brave_publisher_id}/wallet")
+      if publisher.publication_type == :site
+        request.headers["Authorization"] = api_authorization_header
+        request.url("/v2/publishers/#{publisher.brave_publisher_id}/wallet")
+      elsif publisher.publication_type == :youtube_channel
+        request.headers["Authorization"] = api_authorization_header
+        request.url("/v1/owners/#{URI.escape(publisher.owner_identifier)}/wallet")
+      else
+        Rails.logger.warn("PublisherWalletGetter can't get wallet for publication_type #{publisher.publication_type.to_s}")
```
ayumi: This code path should never run, so let's also raise an exception and send it to Sentry (`begin; raise "bla"; rescue => e; Raven.capture_exception(e); end`)

app/services/publisher_wallet_setter.rb
```diff
@@ -18,18 +18,38 @@ def perform
...
+        request.body =
+            <<~BODY
+            {
+              "provider": "uphold", 
+              "parameters": #{JSON.dump(uphold_access_parameters)} 
+            }
+        BODY
+        request.url("/v1/owners/#{URI.escape(publisher.owner_identifier)}/wallet")
+      else
+        Rails.logger.warn("PublisherWalletSetter can't set wallet for publication_type #{publisher.publication_type.to_s}")
```
ayumi: same as the getter; let's also raise an exception and send it to Sentry

Issue #256, opened 2017-11-11
Add locale support to stylesheets


lgebhardt:
We're introducing limited content into the stylesheets for the overlay text on the Publication Type chooser page (email_verified). These strings should be able to be translated, so the stylesheets need to support internationalization. In addition different languages may need different text (fonts, spacing, etc) styles and language specific images.

Based on very limited research I see we could handle this in several ways. Two are linked below, though others probably exist:

https://stackoverflow.com/questions/11577083/access-the-locale-in-ruby-in-css-files
https://stackoverflow.com/questions/9310044/rails-i18n-of-css-file

Let's examine the current and future needs before settling on a strategy.

mrose17:
partial duplicate of https://github.com/brave-intl/publishers/issues/58

PR #257, opened 2017-11-11
Style the Publisher type selector page (internally email_verified)

lgebhardt:
Restyles the publisher type chooser per @jenn-rhim's design.

Still to do: The overlay wording is in the styles. We should add locale support to the styles to enable these to be translated. #256 was added to address this in the future.

screen shot 2017-11-11 at 7 51 54 am

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


config/locales/en.yml
```diff
@@ -233,6 +233,14 @@ en:
     verify_failed_dns_record: "Couldn't verify the DNS record. Please confirm the record in your DNS host, and also note it might take up to 24 hours for DNS to update. You can try again later; or we will check periodically and notify you when we've found it."
     verify_failed_public_file: "Couldn't verify the file. Please confirm successful upload by visiting the link, and also check that it's served over HTTPS. Alternatively you can add a DNS record."
     verify_success: "Verification successful."
+    pub_type_welcome: "Welcome Back!"
+    pub_type_lets_start: "Let's start with choosing who you are based on where you're publishing your content."
+    pub_type_site_operator: "SITE OPERATOR"
```
ayumi: minor: we also could do the capitalization by using CSS `text-transform: uppercase`

PR #258, opened 2017-11-13
Ask publishers for listing permission immediately after verification

### LINKED ISSUE #260: Ask publishers for listing permission immediately after verification
Issue for tracking #258

> Follows specs from @jenn-rhim as closely as possible:

> The dialog can only be dismissed by making a choice.
> The dialog will only appear once. This is guaranteed by setting show_verification_status to null by default, and then setting it to true or false according to the user selection.
> A blur filter is applied to the background when the dialog is displayed.
> The dialog is responsive

dgeb:
Instead of asking publishers for permission to be listed as verified publishers prior to verification, this change introduces a dialog that asks them for this permission _after verification_ on their first visit to the dashboard.

Follows specs from @jenn-rhim as closely as possible:

* The dialog can only be dismissed by making a choice.
* The dialog will only appear once. This is guaranteed by setting `show_verification_status` to null by default, and then setting it to `true` or `false` according to the user selection.
* A blur filter is applied to the background when the dialog is displayed.
* The dialog is responsive (see screenshots below).

In a full width window:

screen shot 2017-11-12 at 9 30 04 pm

In a narrow window:

screen shot 2017-11-12 at 9 29 54 pm

Closes #260

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


mrose17:
@dgeb - looks like travis is having issues...

@ayumi - can you review this PR? thanks!

dgeb:
@mrose17 sorry, I mistakenly included an unrelated testing change which broke travis. Fixed now!

mrose17:
no worries... at least no one posted any private keys (-;

ayumi:
- is there a corresponding github issue for this

- for the noscript case, this patch breaks the dashboard with the invisible overlay. this is unusable.
i think we should start with a display: none overlay and show it via javascript when triggering the modal.

mrose17:
@ayumi - i can't find an issue either. we are running "pretty quickly" on this one...

@dgeb - i share @ayumi's concerns regarding the noscript case. your thoughts/fix?

ayumi:
@mrose17 RE: issue: To maintain project management sanity we must make issues for any work. I've created #260 to track this.

dgeb:
@ayumi Thanks for testing with JS disabled. I've made sure that the overlay does not interfere by marking it `display: none` by default and only showing it with JS, as you suggested.

I've also updated the appearance of the noscript warning to be a little more prominent and consistent with the new look of the site.


screen shot 2017-11-13 at 3 37 21 pm

Issue #255, opened 2017-11-10
Proper branded email template for all notifications in Publishers Brave Payments 


jenn-rhim:
@alexwykoff please add the list of cases with the content :)

![email-template-pub](https://user-images.githubusercontent.com/4369856/32673032-40f3e07a-c602-11e7-913e-88079bd77c2b.png)

jenn-rhim:
@lgebhardt @dgeb

dgeb:
What are the official links for Brave in Medium, Reddit, and Rocketchat?

alexwykoff:
https://basicattentiontoken.org/ the links are all in the footer 

https://medium.com/@attentiontoken 
https://www.reddit.com/r/BATProject/
https://twitter.com/@attentiontoken
https://basicattentiontoken.rocket.chat/

dgeb:
@alexwykoff thanks!

ayumi:
I believe this is resolved as of #261 – if we require content revisions let's create new issues for those.

PR #261, opened 2017-11-14
Branded email

lgebhardt:
Styles the email templates with images, as per #255.
Rewords the login email.

screen shot 2017-11-13 at 9 39 12 pm

Other emails have the new styles, but have not yet had the content reworked. Waiting on further details in #255.

The existing signup email, with new header and footer:
screen shot 2017-11-13 at 9 48 21 pm

The images are embedded in the styles. Ideally I think we should link to them on a CDN to cut the size down for the publishers. I did compress the two largest images using http://compresspng.com.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/views/layouts/mailer.html.slim
```diff
@@ -5,8 +5,17 @@ html
...
   body
     .header
-      = image_tag(attachments["brave_logo_horz_h30.png"]&.url, alt: "Brave Software", class: "logo", height: 30)
-      span.title= content_for(:title)
+      .brand
     .content
       = yield
     .footer
+      .bat
+        a href="https://basicattentiontoken.org/"
```
ayumi: minor: i kinda feel like these URLs should be configured, but it's probably okay

lgebhardt:
@ayumi Changes were made to address the issues you raised as well as changing the signup email after a conversation with @jenn-rhim.

We'll need to add some configuration settings to staging and production now for the new social media urls and the new from_email.

FROM_EMAIL should be in the form of `"Brave Payments Publishers Dev"<the_support_address@replace.me>` (note the quotes) so it will format correctly in email clients.

lgebhardt:
screen shot 2017-11-14 at 10 02 08 am

app/views/publisher_mailer/login_email.html.slim
```diff
@@ -1,14 +1,19 @@
...
 
-p.hint
-  span.attribute= "#{I18n.t("publisher_mailer.shared.contact_help")}:"
-  = mail_to(Rails.application.secrets[:support_email])
+hr
+
+.hint
+  p
+    span= t("publisher_mailer.shared.button_not_working")
+    span= link_to(@private_reauth_url, @private_reauth_url, class: "fallback-link", onMouseOver: "this.style.color='#48494b'", onMouseOut: "this.style.color='#818589'")
```
ayumi: minor: i have a hunch that `onMouseOver` isn't supported in Gmail, however :hover is i think

app/helpers/publishers_helper.rb
```diff
@@ -256,4 +256,10 @@ def publisher_filter_public_file_content(publisher, file_content)
       ""
     end
   end
+
+  def name_from_email(email)
+    return "Publisher" unless email.is_a?(String)
+
+    email.split("@")[0].capitalize
```
ayumi: IMO this is going to look funny more times than not. e.g. if Alice Blogger at BlogLandLLC had an email ablogger@blogland.world it would show as `Hi, ablogger`

i think we can correct this in a followup #270

jenn-rhim:
the button definitely needs some finessing. but pass for now. will fine-tune for the next release...

PR #307, opened 2017-11-20
Email Notifications API

### LINKED ISSUE #7: Add email notifications endpoint for publisher messages
When a publisher accrues a sufficient pending balance, we'll send them a notification email to {webmaster, hostmaster}@foo.example.com (per [RFC 2142](https://tools.ietf.org/html/rfc2142)) and also to WHOIS contacts*.
- publishers will provide a notifications endpoint
- eyeshade will detect pending balances and trigger welcome emails

Clarification needed:
- Which of the WHOIS contacts?

See also: [Brave Payments verification spec](https://gist.github.com/diracdeltas/7c1792a7484ca98ab016710467d1e729).

@mrose17 @diracdeltas

kpfefferle:
Resolves #7 with implementation of an initial use case for verified channel publishers who have not yet set up their Uphold wallet.

A POST to `/api/publishers/[BRAVE_PUBLISHER_ID]/notifications` with param `type=verified_no_wallet` will send the following email (content via @alexwykoff):

Email Preview

_Note that the email template background images are not currently displaying in my local dev environment due to issues with `foreman start`_

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


kpfefferle:
@alexwykoff Do we need a copy of this email sent to the internal address as well?

app/views/publisher_mailer/verified_no_wallet.html.slim
```diff
@@ -0,0 +1,19 @@
...
+== t("publisher_mailer.verified_no_wallet.body_html_2")
+
+p= t("publisher_mailer.shared.signature")
+
+hr
+
+.hint
+  p
+    span= t("publisher_mailer.shared.button_not_working")
+    span= link_to(@publisher_dashboard_url, @publisher_dashboard_url, class: "fallback-link", onMouseOver: "this.style.color='#48494b'", onMouseOut: "this.style.color='#818589'")
```
mixonic: I have a suspicion that the `onMouseOver`, being JavaScript, does not actually run in any email clients.

config/locales/en.yml
```diff
@@ -334,6 +334,17 @@ en:
...
+    verified_no_wallet:
+      subject: "Please create your new wallet to continue receiving funds"
+      body_html_1: |
+        <p>Thanks for verifying your channel! We have a small task for you.</p>
+        <p>We need you to create a wallet with our partner, Uphold.</p>
+        <p>It will only take a couple clicks and you’ll be receiving funds from us as soon as it is completed.</p>
+        <p>Please visit <a href="%{url}">%{formatted_url}</a> to create your wallet.
+      button: "Create My Wallet"
+      body_html_2: |
+        <p>Got questions? Visit <a href="https://basicattentiontoken.org/">basicattentiontoken.org</a> to learn more about how the micropayments work.
```
mixonic: `https` perhaps?

app/services/publisher_notifier.rb
```diff
@@ -3,22 +3,22 @@ class PublisherNotifier < BaseService
...
     @notification_params = (notification_params && notification_params.to_hash.symbolize_keys) || {}
     @publisher = publisher
   end
 
   def perform
-    NotificationMailer.public_send(mailer_method, @publisher, notification_params).deliver_later
-    if NotificationMailer.should_send_internal_emails? && NotificationMailer.respond_to?(mailer_method_internal)
-      NotificationMailer.public_send(mailer_method_internal, @publisher, notification_params).deliver_later
+    PublisherMailer.public_send(mailer_method, @publisher, notification_params).deliver_later
+    if PublisherMailer.should_send_internal_emails? && PublisherMailer.respond_to?(mailer_method_internal)
```
mixonic: @kpfefferle and I talked about this a bit: There currently are not any internal emails sent via this API endpoint, perhaps we could at least add a comment here saying this codepath is for future use. That or simply remove it.

kpfefferle: @mrose17 Do we anticipate that emails sent via the API endpoint will need internal copies sent?

app/views/publisher_mailer/verified_no_wallet.html.slim
kpfefferle: This was copied from here:
https://github.com/brave-intl/publishers/blob/f7385755f1b4b81e850c5bdaeedd8944016f3418/app/views/publisher_mailer/login_email.html.slim#L19

I tested in both macOS Mail and Gmail, and neither seem to execute the JS hover effect. I believe that `a:hover` can possibly be achieved in email via a CSS `<style>` block. Maybe we could try that?

cc: @lgebhardt

config/locales/en.yml
kpfefferle: @mixonic The above `href` does contain HTTPS - ?

app/services/publisher_notifier.rb
ayumi: given we're going to stop logging emails, i think it makes sense now to send the internal redacted copy with any publisher emails.

ayumi:
@mrose17 @alexwykoff do we know what app will utilize this endpoint, and if it's manual or automated

app/views/publisher_mailer/verified_no_wallet.html.slim
lgebhardt: I agree we should probably remove the JS. It was intended for webmail clients, but it just adds complexity for very little gain. CSS might be a good way to go for the hover. It should work in gmail now. But there are still enough clients that strip style tags that we don't want to use it for anything critical.

kpfefferle:
@alexwykoff says that he will have final specs for emails later today, so we likely want to wait to merge this until he has signed off on the contents

app/services/publisher_notifier.rb
kpfefferle: I just added an internal copy method for this initial use case, so this code is now indeed being used.

alexwykoff:
@ayumi presumably eyeshade will use this as events are triggered (verified domain, x number of days where verification started but stalled, etc)

alexwykoff:
@kpfefferle the API shouldn't be blocked by specs for individual emails. We should be tracking that in a separate issue.

kpfefferle:
@alexwykoff @ayumi Issue #7 has meandered a bit over the year it's been open (and caused me some wandering and wasted time due to confusion over what had and hadn't previously been done). I'd recommend closing #7 as the endpoint itself is working and opening a brand new issue for tracking the addition of various emails and their content.

ayumi:
I've created #320 to track further refinements to this email. If there are additional endpoints, let's  create separate issues for each. :green_apple:

PR #319, opened 2017-11-21
Support passing a publisher through to the original account when trying to reclaim a channel

### LINKED ISSUE #305: When the user verifying the channel attempts to verify again, pass them through
As long as the user auth is good, and legit channel, and it is the same user who verified it before, I'd strongly recommend us being forgiving and send the user to the dashboard. --> if fixing it this way is relatively cheap. 

screen shot 2017-11-17 at 11 16 59 am

lgebhardt:
Handles the special case described in #305.

In the special case where a publisher already has a channel claimed, and tries to claim the same channel again using the same email address, they will be taken to the dashboard for their existing account. The new publisher record is removed as it is unneeded.

Adds protection from `destroy` for verified publishers. This is used to assure that the step of destroying the new unneeded publisher is safe.

Fixes #305

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/publishers/omniauth_callbacks_controller.rb
+++ b/app/controllers/publishers/omniauth_callbacks_controller.rb
@@ -17,6 +17,8 @@ module Publishers
       if existing_publisher && current_publisher && current_publisher.youtube_channel_id.nil? &&
           existing_publisher.email == current_publisher.email &&
           existing_publisher.id != current_publisher.id
+        require "sentry-raven"
+        Raven.capture_message("Logging user in instead as existing user with same email and oauth credentials.")
         current_publisher.destroy
         sign_out(current_publisher)
       end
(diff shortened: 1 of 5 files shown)


app/controllers/publishers/omniauth_callbacks_controller.rb
```diff
@@ -1,11 +1,28 @@
...
+      # The presence of a current_publisher may indicate we're attempting to attach a channel to an email verified
+      # publisher OR it's an attempt to log in using oauth.
+      #
+      # If there's an existing publisher with the same oath credentials AND the same email this is an attempt to
+      # register the same channel to effectively the same email. Instead of showing the already registered message
+      # the current_publisher can be deleted and then log them in as the existing_publisher
+      if existing_publisher && current_publisher && current_publisher.youtube_channel_id.nil? &&
+          existing_publisher.email == current_publisher.email &&
+          existing_publisher.id != current_publisher.id
+        current_publisher.destroy
```
ayumi: This should be a very uncommon case, but I'd like to track it in case we made any oops. Could we send an event to Sentry here?

PR #323, opened 2017-11-21
2FA via U2F

### LINKED ISSUE #64: Add multi-factor auth (MFA) / 2-Factor Auth (2FA)
Right now auth is linked to publishers' emails. If publishers emails are compromised, their Brave Publishers account is also compromised. In this case we could improve security if we offered a second factor for authentication.

@diracdeltas @mrose17

mixonic:
Closes #64

This PR introduces support for 2FA via the U2F protocol and API.

Prototype video: https://youtu.be/DZ0BCOWOUzc

Pre-review TODO (not blocked by design):

* [x] Tests (We're past the blocker of creating a strategy for testing the gnarly bits, calling this done)
* [x] Proscribe an HTTPS-based development flow. This stuff can only be tested/used with HTTPS. https://github.com/brave-intl/publishers/pull/334
* [x] ~Youtube-based logins must pass through the 2FA flow.~ Perhaps this should just be part of Youtube? This would imply we should instead restrict users with Youtube auth from adding a 2FA registration in the first place.
* [x] Fix final redirect with `publisher_next_step_path`, currently an error says that route cannot be found
* [x] ~`u2f-api` file is simply a copy of https://demo.yubico.com/js/u2f-api.js. Is there a better way to add this dependency?~ This seem unimportant for now lets punt
* [x] Firefox testing, support. (edit: asserted that our approach is expected to work in production with Firefox nightly and `security.webauth.u2f` flag set to `true`. Firefox U2F support is shaky, and will *not* work with the appId of `https://localhost:3000` https://bugzilla.mozilla.org/show_bug.cgi?id=1421723 instead it requires a full host ala `https://localhost.ssl:3000`.)
* [x] Provide a WIP UI for showing existing registrations
  * [x] In the authentication case, automatically start the auth attempt
  * [x] Errors should be presented to the user with explanation, and should permit reties (both auth attempts and registration attempts)
  * [x] put the pages inside the general grid/container system of the rest of the site
  * [x] on registration if a browser/env does not support U2F the register button should be replaced by a warning/error that the env is not appropriate.
  * [x] on authentication when going into the 2FA flow with an unsupported browser, the user should be presented a warning/error that the env is not appropriate and we should clear their current session (the pending_2fa_current_publisher)
* [x] pass the sign requests to the registration flow, and consider the use-case of a user registering a device a second time. (added a commit passing registered keys, we raise an error code `4` when a user attempts to re-register a device. We can polish this UI with the other error flows)

Blocked by design:

* [ ] UX/UI for rejected browsers.
  * The registration page should not permit you to proceed if the browser is not supported.
  * If authenticating a user in a browser which is not supported, user should be told they cannot proceed.
* [ ] The UX for when a user can add a 2FA device, revoke a device, needs to be designed.
* [ ] The actual device registration process needs design, including error states.
* [ ] The authentication flow needs design including error states.
* [ ] Rebase, commit cleanup

Reviewer Checklist:

Tests

Security:


app/assets/javascripts/u2f_registrations.js
```diff
@@ -0,0 +1,39 @@
...
+
+  /*
+   * Register a u2f device
+   */
+  function registerU2fDevice(formElement, responseInput) {
+    let appId = formElement.querySelector('[name=u2f_app_id]').value;
+    let registrationRequests = JSON.parse(formElement.querySelector('[name=u2f_registration_requests]').value);
+    window.u2f.register(appId, registrationRequests, [], registerResponse => {
+      if (registerResponse.errorCode) {
+        debugger;
```
mixonic: remove

kpfefferle:
@jenn-rhim We could use your direction on design and UX. There are a few primary interactions:

1. Register a new FIDO U2F device
2. Authenticate with a registered FIDO U2F device
3. Manage FIDO U2F devices

## 1. Register a new FIDO U2F device

### Current Prototype

Right now we have a (mostly) unstyled form that allows a logged-in publisher to register a U2F device. The button pictured triggers a request for information from the device, and once we receive the response we store the information we need about the device.
current-reg-1

### Things to Consider

- How does a Publisher get to the U2F registration process?
  - link to dedicated page? included on dashboard?
- How does a Publisher identify a registered device?
  - most sites include a "Name" text field that gets stored to identify the device for future identification/removal
- How do we communicate the various states of the registration process?
  - awaiting response from U2F device
  - error states/messaging (retry button/link?)
  - success states/messaging
- How do we handle browsers that don't support U2F?
  - disable link? hide link? redirect?

### Sites for Comparison

#### [GitHub](https://help.github.com/articles/configuring-two-factor-authentication-via-fido-u2f/)

GitHub Register Button
GitHub Name
GitHub Waiting
GitHub Done

#### [Facebook](https://www.facebook.com/help/401566786855239)

Facebook Add Key Link
Facebook Add Key Modal
Facebook Add Key Modal 2
Facebook Add Key Name
Facebook Add Key Confirmation

## 2. Authenticate with a registered FIDO U2F device

### Current Prototype

We check after a Publisher clicks their emailed login link to see if the Publisher has any registered U2F devices. If they do, we redirect to a page that prompts them to authenticate with their U2F key.
Prototype Authentication

### Things to Consider

- How should we present the authentication?
  - many sites prompt for U2F immediately upon page load
- How do we communicate the various states of the registration process?
  - awaiting response from U2F device
  - error states/messaging (retry button/link?)
  - success states/messaging
- How do we handle browsers that don't support U2F?
  - instruct to use a U2F-compatible browser? messaging?

### Sites for Comparison

#### GitHub

GitHub Authentication

#### Facebook

Facebook Authentication

## 3. Manage FIDO U2F Devices

The Publisher should be able to see which FIDO U2F devices are currently registered to their account. These should be recognizable by the name that the Publisher assigned to each device with the ability for the Publisher to remove any unrecognized/unused devices.

### Things to Consider

- How do we show the Publisher a list of currently registered U2F devices?
- How does a Publisher remove a device?

### Sites for Comparison

#### GitHub

GitHub U2F List

_GitHub removes the U2F key from the list immediately upon clicking the Delete icon (no confirmation step)_

#### Facebook

Facebook U2F List
Facebook Delete Confirm

db/migrate/20171121165127_create_u2f_registrations.rb
```diff
@@ -0,0 +1,13 @@
+class CreateU2fRegistrations < ActiveRecord::Migration[5.0]
+  def change
+    create_table "u2f_registrations" do |t|
```
lgebhardt: Should this use uuids for the id?

```ruby
create_table "u2f_registrations", id: :uuid do |t|
```

mixonic: @lgebhardt What is the motivation for using a UUID here over an integer?

General Brave best practice, or a technical issue you see... etc? We could use a UUID, I know the publisher's table does, but generally I default to an integer.

kpfefferle: This is a good catch @lgebhardt - we had just noticed yesterday that this app has been using UUID IDs for many models. I would agree that we may want to use UUIDs here as well.

lgebhardt: @mixonic I was thinking consistency. I'm not sure there's a compelling technical reason in this case.

kpfefferle: @mixonic I would assume that actions like Delete may reference a URL containing the ID, and using a UUID here would avoid revealing incremented ID's in those URLs. I would see that as an advantage.

mixonic: We went ahead and made this change. Apparently you can set the default primary key type (per https://github.com/rails/rails/issues/23422#issuecomment-326596492) however there is no way to set the default foreign key type which is something that tripped us up when we created the relationship to the publishers model. Seems tricky to navigate in the future, but there it is.

Thanks for the 👀  @lgebhardt

app/views/u2f_registrations/index.html.slim
```diff
@@ -0,0 +1,34 @@
...
+      thead
+        tr
+          th scope="col" Name
+          th scope="col" Registered
+          th scope="col" Last Used
+          th
+      tbody
+        - @u2f_registrations.each do |u2f|
+          tr
+            th scope="row" = u2f.name || "Anonymous Key"
```
mixonic: Your JavaScript is showing ;-) empty strings are truthy in Ruby, this should probably be

```
(u2f.name.blank? ? u2f.name : "Anonymous Key")
```

app/controllers/u2f_authentications_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    begin
+      u2f.authenticate!(
+        session[:challenge],
+        u2f_response,
+        Base64.decode64(registration.public_key),
+        registration.counter
+      )
+    rescue U2F::Error => e
+      redirect_to new_u2f_authentication_path
```
mixonic: Add a logging message with the error

app/assets/javascripts/u2f_registrations.js
```diff
@@ -0,0 +1,39 @@
+;(function() {
+  'use strict';
+
+  /*
+   * Register a u2f device
+   */
+  function registerU2fDevice(formElement, responseInput) {
+    let appId = formElement.querySelector('[name=u2f_app_id]').value;
+    let registrationRequests = JSON.parse(formElement.querySelector('[name=u2f_registration_requests]').value);
+    let registeredKeys = JSON.parse(formElement.querySelector('[name=u2f_registered_keys]').value);
```
mixonic: Call these `sign_requests` this is what they are for authentication (they are the same thing)

test/fixtures/publishers.yml
```diff
@@ -16,6 +16,22 @@ verified:
   verification_token: "e3d801860b2185931ae4dbec06f7119712f928eaf39f0fe0c5a335e42b7eb1c6"
   verified: true
 
+completed:
```
mixonic: Add a comment describing that this is a publisher that has has an email sent with an authentication token

app/views/u2f_registrations/index.html.slim
kpfefferle: Good catch. I had coded (and tested) the default before I added the `name` field to the registration form - so I had tested it only against `nil` values 🤦‍♂️

mixonic:
This ready for serious review. @kpfefferle and I suggest it be reviewed with an eye to landing it into staging (and with the next promotion into production). We believe the work will get better click testing in trunk, we will have fewer integration challenges with other auth features under development, and we will be better able to track/review detailed change once we're working on the main codebase.

For a user who does not have U2F set up this patch should have *zero* impact on their UX. To start using U2F a user must visit `/publishers/u2f_registrations` which we think is obscure enough that no users will be confused. In this patch, we do not add any links to that path.

A technical summary:

* Add the u2f gem.
* Add the vanilla JS UJS gem  for CRUD delete links.
* Add mocha for stubbing some crypto in tests.
* Provide a U2F registrations dashboard at /publishers/u2f_registrations.
* Send users with 2FA to /publishers/u2f_authentications/new as a second login step.
* Provide a warning message to users with 2FA attempting to log in with a browser that doesn't support U2F.
* Tested in:
  * Chrome stable ✅ 
  * Firefix nightly with a TLD (`localhost.ssl`) and the `about:config` flag `security.webauth.u2f` set to `true` ✅ 
  * Safari shows the "U2F not support" page as expected. ✅ 

Further UI/UX details follow:

**Video walkthrough**

I walked through the UX/flows in this screencast (3:41): https://www.youtube.com/watch?v=y9s8CtmgjGE

**/publishers/u2f_registrations**:

screen shot 2017-11-30 at 6 13 58 pm

and with an error:

screen shot 2017-11-30 at 6 14 09 pm

**/publishers/u2f_authentications/new**, the screen you see immediately after clicking a sign in link:

screen shot 2017-11-30 at 6 14 56 pm

And if this fails, the user is prompted to retry:

screen shot 2017-11-30 at 6 15 10 pm

However if a user has 2FA but uses an unsupported browser to sign in (like Safari, Edge, Firefox stable) they will see this page:

screen shot 2017-11-30 at 6 13 45 pm

app/controllers/u2f_authentications_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+        registration.counter
+      )
+    rescue U2F::Error => e
+      Rails.logger.debug("U2F::Error! #{e}")
+      redirect_to new_u2f_authentication_path
+    ensure
+      session.delete(:challenge)
+    end
+
+    registration.update(counter: u2f_response.counter)
```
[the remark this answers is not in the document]
mixonic: Regarding the first item: I do not believe we should increment the counter until after `authenticate!` has completed successfully. The idea of the counter is that we only expect to see increasing counter numbers from a client, never a lower number than what we have already seen. Thus we should not update our internal counter number until we are sure we have a valid authentication.

Regarding the second item, 👍

app/views/layouts/application.html.slim
```diff
@@ -11,6 +11,11 @@ html
     link rel="icon" type="image/png" href=image_path("favicon.ico")
 
   body data-action=params[:action] data-controller=params[:controller]
+    javascript:
```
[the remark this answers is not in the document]
mixonic: My concern is this: If we put it into the normal JavaScript file it would execute before the DOM has been created. So we will use a listener on `DOMContentLoaded` (like https://github.com/brave-intl/publishers/pull/323/files#diff-b7e5bf73cff0ecca1f352a6abb55b95dR50). I'm concerned the HTML may get layout and paint before that event handler fires, and that users would see an unexpected flash of bad content before this check ran.

In most cases I agree with you that inline scripts are a bad style, however in this case I think an inline JS block is the correct approach. I can add a comment here to make clear why this exception is permitted.

app/views/u2f_registrations/_new.html.slim
```diff
@@ -0,0 +1,23 @@
+h3 = t ".heading"
```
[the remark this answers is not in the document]
mixonic: This file is:

```
app/views/u2f_registrations/_new.html.slim
```

I think the one you're saying is similar would be:

```
app/views/u2f_authentications/new.html.slim
```

However these forms are different use-cases (registration vs. authentication) and different UX. There is some similarity, but believe being DRY here would only cause pain as the details about these interfaces further diverge over time. They are not actually the same thing.

db/migrate/20171121165127_create_u2f_registrations.rb
```diff
@@ -0,0 +1,13 @@
+class CreateU2fRegistrations < ActiveRecord::Migration[5.0]
+  def change
+    create_table "u2f_registrations", id: :uuid do |t|
+      t.text "certificate"
+      t.string "key_handle", index: true
+      t.string "public_key"
+      t.integer "counter"
```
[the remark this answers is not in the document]
mixonic: 1. Sure, seems ok
2. I think if we did this we would duplicate a check already being done by the U2F implementation (which checks the counter is increasing already). I don't think we should duplicate logic from the U2F algorithm in our app code.

mixonic:
@ayumi Thanks, I'm landing updates per your feedback and my comments!

Who should we have review this to keep it moving? I'm not sure who you have in mind.

db/migrate/20171121165127_create_u2f_registrations.rb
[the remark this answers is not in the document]
ayumi: where does that check exist currently? i thought maybe since we persist the counter, we should also add safeguards around it.

ayumi:
@mixonic i'm looking for internal brave reviewers, if we can't get that i'll read the u2f spec more deeply and think about it harder

db/migrate/20171121165127_create_u2f_registrations.rb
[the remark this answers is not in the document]
mixonic: As part of `u2f.authenticate!` which we call at:

* https://github.com/brave-intl/publishers/pull/323/files#diff-24cb0fce0ac1bb6a31f16a9fe3732a1fR24

The counter's increase is confirmed. In ruby-u2f:

* https://github.com/castle/ruby-u2f/blob/82bac927030c1bf73b18c3922bb47a4f0fa167b9/lib/u2f/u2f.rb#L60

mixonic:
Changes have been made here and a fresh commit force pushed. @kpfefferle @ayumi I deleted the migration and added a new one to be after the current last migration in the staging branch, you may want to upgrade with the following steps:

```
VERSION=20171113021500 bundle exec rake db:migrate
git fetch mixonic -v
git reset --hard mixonic/u2f
bundle exec rake db:migrate
```

vendor/assets/javascripts/u2f-api.js
```diff
@@ -0,0 +1,756 @@
+// Maintained at https://github.com/google/u2f-ref-code/blob/master/u2f-gae-demo/war/js/u2f-api.js
```
diracdeltas: is it possible to version/update this file so that we inherit changes?

mixonic: I don't believe this file has any formal management or distribution. I wish I did, and hunted a bit.

The [official docs](https://developers.yubico.com/U2F/Libraries/Using_a_library.html) under "Client-side" say "the easiest way is to use the https://demo.yubico.com/js/u2f-api.js library" and simply link to that file dynamically. That seems less than ideal, and unnecessary. I managed to find the Google ref code which seems like the canonical repo where that file is maintained, so I copied from there and provided this attribution link.

PR #330, opened 2017-11-22
Provide link to sign up within flash message if publisher attempts to…

### LINKED ISSUE #325: When telling the user that the login failed...
... give them a link in the notification or somewhere.

in the image below, would have expected "start a new verification" to be a link to the homepage.  the brave logo in the upper-lefthand corner is such a link, but that wasn't obvious to me!

nvonpentz:
… log in with unregistered domain

* Separate en.publishers.new_auth_token_publisher_not_found into two parts, the second of which will be the link

* Modify `find_publisher` to pass both parts as the error message to the publisher controller instead of the removed translation

* Modify publisher controller to flash both messages, adding a link to the homepage for the second

**Security Warning**
This code does contain `.html_safe` - the link would not activate otherwise.  I tried to use `link_to` and ERB instead, but since we are passing plain text via a controller I don't think this is possible.

Resolves #325

Submitter Checklist:

- [x] Submitted a [ticket](https://github.com/brave-intl/publishers/issues) for my issue if one did no already exist.
- [ ] Tagged reviewers.

Test Plan:
Since this is small and straightforward, I tested through the browser.

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -164,7 +164,8 @@ class PublishersController < ApplicationController
     if emailer.perform
       # Success shown in view #create_auth_token
     else
-      flash.now[:alert] = %Q[#{emailer.error[0]}<a href="/">#{emailer.error[1]}</a>].html_safe
+      # Failed to find publisher
+      flash.now[:login_link] = "" # Uses login_link partial instead of explicit message
       render(:new_auth_token)
     end
   end
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -164,7 +164,7 @@ class PublishersController < ApplicationController
     if emailer.perform
       # Success shown in view #create_auth_token
     else
-      flash.now[:alert] = emailer.error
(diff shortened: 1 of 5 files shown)


app/controllers/publishers_controller.rb
```diff
@@ -164,7 +164,7 @@ def create_auth_token
     if emailer.perform
       # Success shown in view #create_auth_token
     else
-      flash.now[:alert] = emailer.error
+      flash.now[:alert] = %Q[#{emailer.error[0]}<a href="/">#{emailer.error[1]}</a>].html_safe
```
ayumi: this looks scary to me, i do not think we should commit this.

Instead I suggest following this guide on flash messages ("Identifiers for more complex messages"); add a new flash[:login_link] and new partial, and check and render that in `application.html.slim`.
https://thepugautomatic.com/2012/08/the-rails-flash/

nvonpentz:
Thanks for your comments @ayumi.  I went in the direction you suggested, and believe I have a better solution now - without `html_safe`.

Issue #273, opened 2017-11-14
Allow a single Owner to manage many Properties


lgebhardt:
Currently "publishers" (people who publish) and the "published property" (a web site, a video channel, etc.) are combined in one table in the database. This creates usability and management  issues for publishers that publish multiple properties. For example statements can not be aggregated between properties, Uphold wallets must be linked to each property, management tasks such as changing the owner's name or email must be performed for each property.

To mitigate these issues the goal is to separate the publishers from their properties in the database and UI of the site. Functionally we will have a model where one Owner owns many properties (sites, video channels, etc).

To accomplish this will be relying on the fact that all verified entries in the Publishers table have a verified email address.  A unique Owner will be created for each unique email address in the table. We can expect there will be differences in the name and phone number of these owner accounts. Flattening this data is still TBD. Simplest case is we take the first entry's values and discard the rest.

The existing Publishers table entries will then be associated with the new owner records. At this point the Publisher's table will represent Properties. Further refinement could be done to normalize different property types. This is not being proposed in this issue.

The login process after this step will involve entering an email or domain, receiving an email containing a login link and using that link to authenticate. A second phase will include the addition of OAuth to allow authentication against google (and potentially other OAuth providers). See #274.

Once owners are associated with multiple properties they will need a way to manage them. #229 address adding and removing properties. @mrose17 noted that some Owners will have 100s of Properties.

@mrose17 indicated that we will start with one wallet per owner. If a need to associate different wallets with different Properties is need we will address in a follow-on release.

Statements on the dashboard will support both per Owner and per Property statements.

mrose17:
@lgebhardt - if you take a look at https://github.com/brave-intl/bat-ledger/blob/master/eyeshade/controllers/owners.js#L393 you'll see how the eyeshade server "taxonimizes" an owner. publishers have an owner field, that if present, refesr to the associated owner. you may not need all the fields (e.g., `provider*` is used to hold the components of the owner-identity.  publishers also have a field to allow them to have uphold information distinct from their owner. for the publishre site, we can consider that a future.

ayumi:
This has essentially been implemented (in `staging2`) by #452 -- lets track blockers for deployment in separate issues.

PR #331, opened 2017-11-27
Feature/expired auth token page

### LINKED ISSUE #304: Expired key screen should be better handled
1. the message is tucked away that I doubt user sees it
2. the message is not user friendly

Likely they'd be puzzled as to what happened as they just returned to the same screen before clicking verify email link.

screen shot 2017-11-16 at 10 38 53 pm

nvonpentz:
The goal is to have every instance of an expired authentication token, both sign up and log in, redirect to a new dedicated view.

Here are images of the new view at the different breakpoints:

image

image

image

The view, `expired_auth_token.html.slim` was roughly modeled after `new_auth_token.html.slim` but relies on a new css class `expired-token-panel`.  Upon clicking 'continue' the form is posted to `publishers_path` like the 'Get Started' button on the landing page.

I followed the [devise documentation](https://github.com/plataformatec/devise/wiki/How-To:-Redirect-to-a-specific-page-when-the-user-can-not-be-authenticated) on how to override the devise method `authenticate_user!` to redirect to a specific page instead of the root path.

Make sure to restart foreman when testing to in development locally.

Resolves #304 

Submitter Checklist:
- [x] Tagged reviewers.

Test Plan:
Tests focus on ensuring that both sign up and log in attempts with an expired token fail.  If the token is valid, flow should continue as normal.

I created two new tests in `publishers_controller_tests.rb` that ensure a invalid and valid token fail and succeed respectively when signing up with a new email address.  I also added more explicit checks to pre-existing tests to ensure expired/invalid tokens fail when logging in, and succeed otherwise.

I updated one test that previously checked that a publisher attempting with an invalid token be redirected to the `root_path` to instead check if redirected to `expired_auth_token`. 

Reviewer Checklist:

Tests

Security:


ayumi:
after discussion, @lgebhardt noted:
> Until the accounts change goes in there won't be a way to login with only an email address. So it's blocked until we either finish the splitting the accounts out, or add the other login options/redirect to the login page instead.

nvonpentz:
The issue this PR attempts to solve (#304) will be handled after #273 has been resolved.

PR #332, opened 2017-11-27
Add client side validations for publishers phone numbers

### LINKED ISSUE #324: Add validations for all publisher input fields
Once logged in, publishers can enter ~100 characters of any kind in to the phone number field.  They should only be able to enter valid phone numbers.

image

nvonpentz:
* Add validation that forces phone numbers to only include numbers, dashes, parentheses, periods, and plus (+).  There is no restrictions on the order or length of characters in accordance with the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle).

* Add title field within html tag to give user more information as to why they can't submit the form if their phone number is invalid

* Add the title message to en.yml for internationalization

Previously, phone numbers were being normalized and saved under a property `phone_normalized`.  If a publisher added an invalid number, only the `phone` property was updated, and `phone_normalized` was untouched.  Only valid numbers will update `phone_normalized`.  So this change does not affect the database, only nudges publishers to provide a valid number client side.

Resolves #324 

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/views/publishers/contact_info.html.slim
+++ b/app/views/publishers/contact_info.html.slim
@@ -20,7 +20,7 @@
             = f.text_field(:name, class: "form-control", placeholder: "Alice Bloglette", required: true)
           .form-group
             = f.label(:phone, t("publishers.verified_phone_html"), class: "control-label")
-            = f.phone_field(:phone, class: "form-control", placeholder: "+1 888 555 9001", required: false)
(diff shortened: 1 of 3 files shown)


nvonpentz:
@mrose17 I originally found it [here](https://martinwolf.org/blog/2015/04/html5-telephone-input-validation), but also added support for periods.

mrose17:
@nvonpentz - thanks going forward, let's give credit as appropriate, e.g.,

// taken from ... with edits to allow for periods.

no need to fix this now, but let's add it in the future, thanks!

PR #334, opened 2017-11-27
Run the development server as SSL.

mixonic:
As we add support for U2F-based 2FA and adopt other cutting-edge web features, we need to be working on an HTTPS domain. This is because some APIs (like U2F) are only available to an SSL host.

## Adding a domain for SSL

Please create an entry in your `/etc/hosts` file for `localhost.ssl` pointing at `127.0.0.1`. The following command should be sufficient:

```
echo "echo '127.0.0.1 localhost.ssl' >> /etc/hosts" | sudo -s
```

I've preferred `localhost.ssl` as a domain over `localhost`. There are two motivations for this:

* Chrome and other browsers have some special `localhost`-only behaviors intended to make development simpler, however the rules are vague and poorly documented. It seems simpler to use a non-standard TLD and avoid any confusion.
* Second we may want to use HSTS at some point. HSTS sets an SSL-only rule for an entire domain, and if we set HSTS for `localhost` we might make it difficult to develop non-SSL projects on that domain. 

## Installing SSL certificates

The built-in development Puma server expects two files to exist in the `ssl/` directory:

```
ssl/server.key
ssl/server.crt
```

These should be a key and certificate for the domain `localhost.ssl`. You may already have certificates for this domain, and re-using those is encouraged.

However if you need to generate a key and certificate, you can use the following commands on OSX:

```
bundle exec rake ssl:generate
```

Will generate a key, signing request, and certificate.

```
bundle exec rake ssl:install
```

Will add the certificate to your OSX keychain.

Once you've run these commands you can simply run the Foreman server like normal and visit https://localhost.ssl:3001/. You may still need to approve an exception to allow the certificate the first time you use it in a browser.

References: https://github.com/brave-intl/publishers/pull/323

### DIFF
--- a/lib/tasks/ssl.rake
+++ b/lib/tasks/ssl.rake
@@ -1,0 +1,22 @@
+namespace :ssl do
+  task :generate => [ :environment ] do
+    sh 'which openssl' do |ok, res|
+      if ! ok
+        puts "The `openssl` executable is not available. Please manually generate a key `ssl/server.key` and certificate `ssl/server.crt`."
+      else
+        sh 'openssl genrsa -out ssl/server.key 4096'
+        sh "echo \"openssl req -new -key ssl/server.key -x509 -nodes -new -out ssl/server.crt -subj /CN=localhost.ssl -reqexts SAN -extensions SAN -config <(cat /System/Library/OpenSSL/openssl.cnf <(printf '[SAN]\\nsubjectAltName=DNS:localhost.ssl')) -sha256 -days 3650\" | bash"
+      end
+    end
+  end
+
+  task :install => [ :environment ] do
+    sh 'which security' do |ok, res|
+      if ! ok
+        puts "The `securty` executable is not available. Please manually trust the certificates in `ssl/`."
+      else
+        sh 'security add-trusted-cert -d -r trustRoot -k ~/Library/Keychains/login.keychain ssl/server.crt'
+      end
+    end
+  end
+end
--- a/lib/tasks/ssl.rake
+++ b/lib/tasks/ssl.rake
@@ -4,18 +4,10 @@ namespace :ssl do
       if ! ok
         puts "The `openssl` executable is not available. Please manually generate a key `ssl/server.key` and certificate `ssl/server.crt`."
       else
-        sh 'openssl genrsa -out ssl/server.key 4096'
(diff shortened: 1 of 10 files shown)


mixonic:
@ayumi I have two concerns:

* Only *some* features will reliably work on HTTP. Providing devs with a flaky and inconsistent development environment doesn't seem preferable to making them do a small amount of setup.
* The emails (and this is a pretty heavily email-driven app) contain links. Either:
  * a) We pick one way to generate links, HTTP or HTTPS. It seems like we *must* pick HTTPS then or some things can't be tested without carefully mangling the URLs.
  * b) We dynamically detect if there is an SSL certificate or if the visitor was using HTTPS and only generate HTTPS links in those cases. This sounds like a fair amount of complexity, enough to be a maintenance burden and confusing to future developers.

These thoughts push me to think we should simply go HTTPS all the time. I agree we want to onboarding/setup to be as simple as possible though, which is why I attempted to automate as much as I could. I could try to go a bit further.

Happy to chat more on Slack tomorrow to nail down a direction.

lib/tasks/ssl.rake
```diff
@@ -0,0 +1,22 @@
+namespace :ssl do
+  task :generate => [ :environment ] do
+    sh 'which openssl' do |ok, res|
+      if ! ok
+        puts "The `openssl` executable is not available. Please manually generate a key `ssl/server.key` and certificate `ssl/server.crt`."
+      else
+        sh 'openssl genrsa -out ssl/server.key 4096'
+        sh "echo \"openssl req -new -key ssl/server.key -x509 -nodes -new -out ssl/server.crt -subj /CN=localhost.ssl -reqexts SAN -extensions SAN -config <(cat /System/Library/OpenSSL/openssl.cnf <(printf '[SAN]\\nsubjectAltName=DNS:localhost.ssl')) -sha256 -days 3650\" | bash"
```
ayumi: on debian 9 this command fails for me

mixonic:
@ayumi Two or three items there I can address:

* Linux support 👍  sounds good, I don't have a box readily available to test, thanks for taking a stab at it and iterating.
* Missing from the simplified command you used is a `SAN` in the certificate. As of Chrome 58, `SAN` is required instead of the `CN` (common name) field. Official announcement: https://www.chromestatus.com/feature/4981025180483584 https://groups.google.com/a/chromium.org/forum/#!topic/security-dev/IGT2fLJrAeo. I'm sure a Linux box would fail since my script references a file on the OSX filesystem. I'll try to remove that requirement (some ideas are in https://alexanderzeitler.com/articles/Fixing-Chrome-missing_subjectAltName-selfsigned-cert-openssl/).
* Re: changing hosts, the motivations for this are in the PR description. I'll swap to using `localhost` as a domain though.
* The self-signed certificates are added to the user's login keychain. This really isn't bad unless you think someone will get these certificates and use them to spoof `localhost` or `localhost.ssl` on you. I'll see if I can just skip that step regardless. Mostly we end up trusting these certs in the browsers anyway 👍 

I'll make some tweaks!

PR #342, opened 2017-11-30
Prefer async polling for host verification

### LINKED ISSUE #297: Host verification can take long enough on production that Heroku's proxy times out
I've experienced this on production while entering the domain `orbitjs.com` as a site on the `Contact info` page. This domain (which I own) does not currently have a `www` host entry, so I believe the long background fetch which times out results in an extra long request.

It would be ideal for host verification to be entirely a background process that's combined with frontend polling.

dgeb:
When a domain is entered, and JS is enabled, the contact info form will be submitted via an xhr. The process of normalizing the domain and inspecting the host will be done in an async background job, and results will be periodically polled while a spinner is shown. Once the domain has been set, the progress train will be advanced.

When JS is disabled, the contact info form will behave as it does currently. It will still be submitted via a standard PATCH and the domain normalization / host inspection will be performed synchronously.

These changes required that a number of JS utils be extracted from the dashboard and made available for use throughout the site. JS modules have been added for fetch utils, spinners, and flash messages.

I think we're getting to a level of dynamism on the frontend where we should consider introducing a framework such as Glimmer or Ember to replace the vanilla JS. I'm getting uncomfortable with the lack of module management for JS and the use of globals. We also need to improve the testing of the growing JS codebase.

Fixes #297 

/cc @ayumi 

Submitter Checklist:

- [X] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/assets/javascripts/flash.js
```diff
@@ -0,0 +1,28 @@
+(function() {
+
+  this.flash = {
+    clear: function() {
+      var containerElement = document.getElementById('flash');
```
ayumi: minor: should we cache `var containerElement = document.getElementById('flash')`?

app/models/publisher.rb
```diff
@@ -167,37 +153,34 @@ def owner_identifier
...
-      :brave_publisher_id,
-      I18n.t("activerecord.errors.models.publisher.attributes.brave_publisher_id.api_error_cant_normalize")
-    )
-  rescue URI::InvalidURIError
-    errors.add(
-      :brave_publisher_id,
-      I18n.t("activerecord.errors.models.publisher.attributes.brave_publisher_id.invalid_uri")
-    )
+  def validate_domain_normalization_status
+    error_status = self.domain_normalization_error_status
```
ayumi: i feel like we should be affirmatively checking normalization status. in this case in `domain_normalization_error_status` we're looking for errors and the default case is to validate.
so if we accidentally added an unknown error status, this validation could pass.

app/controllers/publishers_controller.rb
```diff
@@ -140,6 +141,26 @@ def update_unverified
     end
   end
 
+  def domain_status
```
ayumi: we should consider moving this action to its own Publishers API controller to prevent cluttering of the PublishersController.

i would also be satisfied with a TODO: comment

app/controllers/publishers_controller.rb
```diff
@@ -133,11 +133,37 @@ def update
 
   def update_unverified
     @publisher = current_publisher
-    success = @publisher.update(publisher_update_unverified_params)
-    if success
-      redirect_to(publisher_next_step_path(@publisher))
-    else
-      render(:contact_info)
+    domain = params[:publisher][:brave_publisher_id]
```
ayumi: minor: inconsistency using raw params here and strong params below with `publisher_update_unverified_params`

app/assets/javascripts/spinner.js
```diff
@@ -0,0 +1,69 @@
...
+          elementId = element.id = 'temp-' + tempId++;
+        }
+      } else {
+        elementId = 'cssload-container';
+      }
+
+      if (!element) {
+        element = document.getElementById(elementId);
+        if (!element) {
+          element = document.createElement('div');
```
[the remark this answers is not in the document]
dgeb: This allows the spinner to be used with no markup on the page. 

It allows us to remove the `#cssload-container` element from the dashboard - see https://github.com/brave-intl/publishers/pull/342/commits/588c1bf327644b23e43981e1b89657620a52c6a6#diff-3f493ac8654a07a9dbbeb5563811e0d8L353

It is also used (in a later commit) to spawn the spinner on the contact info page.

app/assets/javascripts/flash.js
dgeb: Sounds legit since we are relying on the flash being a single constant element. I will update 👍

dgeb:
Thanks for the thorough review @ayumi! 

> I still think adding a new field like brave_publisher_id_unnormalized could be a good idea to persist the intermediate state in a safe place (currently it looks like we're just keeping it in an async job / redis).

I'm glad to add `brave_publisher_id_unnormalized` to track this state in the db. This will become more important as we add multi-channel support and separate owner identities. 

I will also replace `domain_normalization_status` with `brave_publisher_id_error` to provide more symmetry among brave_publisher_id-related fields and to make it clear that this field is to be used for errors only (not positive statuses).

app/models/publisher.rb
dgeb: I'm going to tighten this up by replacing `domain_normalization_status` with `brave_publisher_id_error`. We can check for the presence of the error code and throw if the description can't be found. This should prevent anything from slipping through the cracks (such as an unknown error status).

dgeb:
@ayumi This is ready for another review now. I've refactored as you requested (except in spinner.js, as mentioned above) and I think the code is clearer now. Thanks again for the suggestions!

ayumi:
@dgeb thanks! ⛽️ 

Is there an edge case when brave_publisher_id normalization fails due to api error; and the user's unable to try to reattempt normalization because they have to update the Publisher but can't because 
```rb
before_validation :register_brave_publisher_id_error, if: -> { brave_publisher_id_error_code.present? }
# ...
def register_brave_publisher_id_error
self.errors.add(#...)
```
 and we can't

dgeb:
@ayumi Good call. I opened #346 to track this edge case and will address it in a separate PR.

PR #344, opened 2017-12-01
Allow developers to force a Captcha using query string param

### LINKED ISSUE #339: Add reliable way to trigger captcha
### Description
Currently there is no reliable way to generate captchas on production and staging.  This makes it difficult to debug errors with captchas.

We should add a query string param that will generate a captcha.

nvonpentz:
A hidden field captures the value of the param 'captcha' when the form is loaded, and passes it through to the controller when it is submitted.  When the controller checks to see if a captcha should be loaded, it will now also check if `params[captcha]` is true, and load the captcha if so.

**How To Use**
To test this on development, add the rack-attack and recaptcha gems to development.
1. Go to any form that is protected by captchas (homepage, login)
2. Tack on `?captcha=true` to the URL and reload
3. Submit the form

The captcha will open.

Resolves #339.

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -59,7 +59,7 @@ class PublishersController < ApplicationController
 
     @publisher = Publisher.new(pending_email: params[:email])
 
-    @should_throttle = should_throttle_create? || params[:captcha] == "true"
+    @should_throttle = should_throttle_create? || params[:captcha]
     throttle_legit =
       @should_throttle ?
         verify_recaptcha(model: @publisher)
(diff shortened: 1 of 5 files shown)


app/controllers/publishers_controller.rb
```diff
@@ -59,7 +59,7 @@ def create
 
     @publisher = Publisher.new(pending_email: params[:email])
 
-    @should_throttle = should_throttle_create?
+    @should_throttle = should_throttle_create? || params[:captcha] == "true"
```
ayumi: i think we don't need to check for "true", truthiness is sufficient (e.g. /?captcha=1)

app/controllers/publishers_controller.rb
```diff
@@ -76,8 +76,7 @@ def create
         redirect_to(root_path, notice: I18n.t("publishers.invalid_email_value") )
       end
     else
-      Rails.logger.error(I18n.t("recaptcha.errors.verification_failed"))
-      redirect_to(root_path, notice: I18n.t("recaptcha.errors.verification_failed"))
+      redirect_to root_path(:captcha => params[:captcha])
```
ayumi: could you update this to the newer colon hash syntax: `captcha: params[:captcha]`

PR #345, opened 2017-12-01
Clarify @ symbol for DNS site verifcation

### LINKED ISSUE #135: Clarify the meaning of "@" for DNS site verification
In DNS land `@` means "apex" or the root domain, so even though it shows "@" you should not write that in the DNS record.
This isn't obvious, so we should do either:
- Clarify what "@" means
- Remove the "@" and say "record on the root domain, no subdomain"

pic
screen shot 2017-09-28 at 17 35 22

nvonpentz:
### DNS verification page after changes
image

### Changes
* Remove 'Name' form with @ symbol that may confuse users

* Replace with instructions on adding to root domain, clarify this is often represented by @

* Downcase the 'P' in 'Copy and Paste'

* Add translation for 'Copy to Clipboard' and downcase the 'C'

Resolves #135.

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/views/publishers/verification_dns_record.html.slim
+++ b/app/views/publishers/verification_dns_record.html.slim
@@ -16,24 +16,24 @@
           strong= I18n.t("publishers.verification_dns_record_help_2")
 
         p
+          span= I18n.t("publishers.verification_dns_record_instruction3")
+
+        p
           span= I18n.t("publishers.verification_dns_record_instruction1")
           strong= current_publisher.brave_publisher_id
           span= I18n.t("publishers.verification_dns_record_instruction2")
 
         .form-group
-          label.control-label= I18n.t("publishers.verification_option_dns_record_name")
-          span.form-control
-            span.color-orange#name= "@"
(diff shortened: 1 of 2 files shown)


app/views/publishers/verification_dns_record.html.slim
```diff
@@ -15,25 +15,25 @@
...
-        .form-group
-          label.control-label= I18n.t("publishers.verification_option_dns_record_name")
-          span.form-control
-            span.color-orange#name= "@"
-          button.btn.btn-xs.btn-default.copy-button data-clipboard-target="#name" Copy to Clipboard
         .form-group
           label.control-label= I18n.t("publishers.verification_option_dns_record_type")
           span.color-orange.form-control#type TXT
-          button.btn.btn-xs.btn-default.copy-button data-clipboard-target="#type" Copy to Clipboard
+          button.btn.btn-xs.btn-default.copy-button data-clipboard-target="#type"
```
ayumi: minor: i wonder if we need this copy to clipboard

Issue #329, opened 2017-11-22
Staging issues


nvonpentz:
As requested by @mrose17, I have tested https://publishers-staging.basicattentiontoken.org/ and compiled the problems into this single issue.

![image](https://user-images.githubusercontent.com/12549658/33142614-1df5928a-cf85-11e7-9841-b75b484a89b3.png)

1. ~~When editing contact information, publishers can update their phone number with an invalid number.  This is described in full in #324~~ This has been resolved in [PR 332](https://github.com/brave-intl/publishers/pull/332).

2. When clicking "Generate" to produce the statement, the statements appear as "Delayed" after about a minute of wait time.  It should report that I have no contributions, and no statements - or that it could not generate the reports.  There is no option to download a report.

3. Repeatedly clicking "Generate" repeatedly produces the same statement.  We only should be able to generate a single statement for a given time period.  These "statements" persist after logging out and in again. @jenn-rhim do we want the statements to persist?

4. The statement lists on the publishers page are not organized in any way.  We can generate statements of any time period in any order (with duplicates), and they are displayed in the order they are generated. 

The UI improvements for statements will be addressed in #265.

5. I **do not** receive the statements in my email when authenticated as a youtube publisher.  I suspect this is because the call the eyeshade is timing out and that all this activity is just client side javascript.

jenn-rhim:
@mrose17 

1. Can we reset the statement return line so it only displays the status of the latest request? i.e. always one line, and subsequent request will replace the previous one.

2. As a temporary solution for those who just signed up -- Can we have the server to return settlement = yes or no, so that if no, we display 'no statement available' and disable the action? we will come up with a better solution in the meantime.

dgeb:
I'm starting this work by cleaning up the statement generation in a couple ways:

* Statement date is no longer shown as a column. Since statements generated > 24 hours ago will be purged, the date is rather meaningless.

* As statements are generated for a period, the corresponding option is eliminated from the select element. Once all options have been exhausted, the statement generation controls are hidden. Thus, only one statement per period can be generated. 

Here's the WIP with a few statements generated:

![screen shot 2017-11-30 at 5 04 35 pm](https://user-images.githubusercontent.com/29122/33458487-6d8b1002-d5f4-11e7-834a-47c140d4ce07.png)

And here's how it looks with all the options exhausted:

[...]

PR #347, opened 2017-12-04
Simplify and improve statement generation

dgeb:
This work addresses the dashboard statement generation issues mentioned in #329 as follows:

* Statement date is no longer shown as a column. Since statements generated > 24 hours ago will be purged, the date is rather meaningless.

* As statements are generated for a period, the corresponding option is eliminated from the select element. Once all options have been exhausted, the statement generation controls are hidden. Thus, only one statement per period can be generated. 

Here's how the dashboard looks with a few statements generated:

screen shot 2017-11-30 at 5 04 35 pm

And here's how it looks with all the options exhausted:

screen shot 2017-11-30 at 5 19 59 pm

I believe @jenn-rhim will want to further refine this approach, but this is at least a start.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/views/publishers/home.html.slim
+++ b/app/views/publishers/home.html.slim
@@ -76,7 +76,7 @@ javascript:
 
       var defaultCurrencySelect = document.getElementById('publisher_default_currency');
       if (defaultCurrencySelect) {
-        defaultCurrencySelect.addEventListener('change', function (event) {
+        defaultCurrencySelect.addEventListener('change', function(event) {
           submitForm('update_default_currency_form', 'PATCH', true);
           refreshBalance();
         }, false);
@@ -101,6 +101,7 @@ javascript:
 
       var generateStatement = document.getElementById('generate_statement');
       var generateStatementResult = document.getElementById('generate_statement_result');
+      var statementGenerator = document.getElementById('statement_generator');
       var statementPeriod = document.getElementById('statement_period');
       var generatedStatements = document.getElementById('generated_statements');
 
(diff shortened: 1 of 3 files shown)


app/views/publishers/home.html.slim
```diff
@@ -139,56 +140,58 @@ javascript:
...
               dynamicEllipsis.stop(statementDownloadDiv);
-              statementDownloadDiv.innerText = 'Delayed'
-            }
-            generateStatement.style.display = 'inline-block';
-          });
-      }, false);
+              statementDownloadDiv.innerHTML = '<a href="/publishers/statement?id=' + statementId + '">Download</a>';
+              generateStatement.style.display = 'inline-block';
+            })
+            .catch(function(e) {
```
mixonic: Is this `catch` expecting that the error is a timeout? Could you check to be sure the error is what you expect for this logic, and propagate if it is unexpected? For example:

```js
.catch(function(e) {
  if (e instanceof TimeoutError === false) {
    // Error is unexpected, propagate
    throw e;
  }
  // Logic to handle the timeout case
});
```

This will keep the code from swallowing unexpected errors, something that makes it difficult to debug unexpected faults later on.

ayumi: This seems reasonable; let's feel free to do a little patch after this one. Thanks!

dgeb: Yep, I agree. Tracking this in #352.

PR #351, opened 2017-12-05
Refactor validations related to brave_publisher_id

### LINKED ISSUE #346: Provide means to retry setting of brave_publisher_id
As @ayumi [pointed out](https://github.com/brave-intl/publishers/pull/342#issuecomment-348647242), #342 introduces an edge case in which `brave_publisher_id` normalization / inspection fails and cannot be reattempted. We should ensure that `brave_publisher_id_error_code` can be cleared out so that users don't need to recreate an account to retry entering a domain.

dgeb:
This fixes a couple edge cases that were not handled well in #342:

* When domain normalization/inspection fails, it is now possible to reattempt submitting the domain (or an updated domain) in the contact info form.

* When a domain that matches an already verified domain has been entered, the error is now displayed properly to the user when JS is enabled.

Fixes #346 

/cc @ayumi 

Submitter Checklist:

- [X] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -37,8 +37,15 @@ class Publisher < ApplicationRecord
   validates :brave_publisher_id, absence: true, if: -> { auth_user_id.present? }
   validates :auth_user_id, absence: true, if: -> { brave_publisher_id.present? }
 
-  # TODO: Show user normalized domain before they commit
-  before_validation :register_brave_publisher_id_error, if: -> { brave_publisher_id_error_code.present? }
+  # ensure that site publishers do not mix:
+  # - normalized and unnormalized domains
+  # - normalized domains and domain-related errors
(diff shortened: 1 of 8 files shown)


app/models/publisher.rb
```diff
@@ -172,13 +181,19 @@ def generate_verification_token
     update_attribute(:verification_token, PublisherTokenRequester.new(publisher: self).perform)
   end
 
+  def clear_brave_publisher_id
```
ayumi: I don't think this method is used

dgeb: Ah good catch. This method was from an intermediate iteration and should have been abandoned. 👍

dgeb:
Ok I'll include some more controller tests and try to wrap this up first thing tomorrow.

dgeb:
Feeling better about this now. I've added controller tests for both xhr and standard paths to update the domain, showing a failure case followed by a success.

Please take a look when you're ready @ayumi   👀

PR #357, opened 2017-12-06
Add locale label for brave_publisher_id_unnormalized

### LINKED ISSUE #356: "Brave Publisher ID Unnormalized" label during signup
Today while re-following the signup flow, I noticed that the "Contact Information" form has a field labeled "Brave publisher id unnormalized" instead of "Website Domain":

image

This seems to have happened because the form now uses the `brave_publisher_id_unnormalized` attribute instead of the `brave_publisher_id` attribute. I have a PR incoming to add a locale setting for `brave_publisher_id_unnormalized` that matches the one for `brave_publisher_id`.

kpfefferle:
Resolves #356 

image

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -24,6 +24,7 @@ en:
     attributes:
       publisher:
         brave_publisher_id: "Website Domain"
+        brave_publisher_id_unnormalized: "Website Domain"
         phone_normalized: "Phone Number"
         show_verification_status: "List my site in Brave's verified publisher site once verified."
         name: "Contact Name"


kpfefferle:
I think we probably want to merge this before deploying to production. cc: @ayumi @dgeb

dgeb:
Thanks for fixing this @kpfefferle. I also added this string in https://github.com/brave-intl/publishers/pull/351/files#diff-37c9b835927e3c363588f172de9bb6faR27 but it must have gotten lost in a subsequent commit.

Issue #349, opened 2017-12-04
Tweak mobile views for dashboard and log in pages


nvonpentz:
After tweaking the mobile views for the homepage in #337 , we have yet to adjust the mobile views for the dashboard and login page.

We need to determine when to address this.

mixonic:
I'm going to focus my effort on an improvement to the logic flow (currently not mobile friendly). Here is a screenshot of the current login page on iPhone 7:

<img width="385" alt="screen shot 2017-12-06 at 2 02 08 pm" src="https://user-images.githubusercontent.com/8752/33679937-187635e4-da8e-11e7-83f8-b120237908be.png">

@jenn-rhim is considering the impact to the dashboard which is probably more extensive and requires detailed mocks.

mrose17:
Blocked on #382 and #385

nvonpentz:
Resolved by #365 and others

PR #365, opened 2017-12-07
Tweak styles for the login page on mobile

mixonic:
Refs #349

* Change the viewport scaling to be `1`
* Add padding so the background gradient is visible after the container box on a small screen
* Remove minimum widths from several elements
* Tweak styles at small and extra-small breakpoints to be more reasonable

TODO:

* [x] general approval by @jenn-rhim 
* [x] Need to sanity check logged-in dashboard pages, perhaps limit this change to the login page. **Done, this only impacts the login page.**

**Mobile at 375px (iPhone7)**

before:

screen shot 2017-12-07 at 11 08 18 am

after:

screen shot 2017-12-07 at 11 08 27 am

screen shot 2017-12-07 at 11 08 42 am

**Mobile at 768px (tablet)**

before:

screen shot 2017-12-07 at 11 09 32 am

screen shot 2017-12-07 at 11 09 40 am

after:

screen shot 2017-12-07 at 2 39 46 pm

screen shot 2017-12-07 at 2 39 57 pm

**At all sizes larger than this, there is no visual change**

### DIFF
--- a/app/views/layouts/publishers.html.slim
+++ b/app/views/layouts/publishers.html.slim
@@ -2,4 +2,9 @@
   .main-content
     #content.container
       = yield
-= render(template: "layouts/application")
+
+= render \
+  template: "layouts/application",
+  locals: { \
+    viewport_initial_scale: ((action_name === 'new_auth_token') && '1.0') \
+  }
--- a/app/views/publishers/new_auth_token.html.slim
+++ b/app/views/publishers/new_auth_token.html.slim
@@ -1,13 +1,9 @@
 - content_for(:navbar_content) do
 
-.publisher-domain-panel.publisher-panel.col-center
-  .publisher-domain-name
-    | &nbsp;
-
 .publisher-panel.login-panel.col-center
   .sub-panel.process-panel
     .row
-      .col.col-md-12
+      .col
         h3.text-center= t("publishers.new_auth_token")
     .row
       .col.col-md-5.site-login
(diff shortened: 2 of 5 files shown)


jenn-rhim:
@mixonic 
- the mobile view looks great.  
- for tablet, we need to reduce the content block width to be about 65-70% of the viewport on tablet (both landscape and portrait), and redefine the text sizes the same way as the mobile view. 

a quick mock:

image

jenn-rhim:
looks all good to me!

app/views/publishers/new_auth_token.html.slim
```diff
@@ -1,13 +1,9 @@
 - content_for(:navbar_content) do
 
-.publisher-domain-panel.publisher-panel.col-center
-  .publisher-domain-name
-    | &nbsp;
```
mixonic: @lgebhardt I think this landed in an old commit from you- I've removed it b/c it was adding visual padding but I don't think it was doing anything besides that. Does that sound correct?

app/views/layouts/publishers.html.slim
```diff
@@ -2,4 +2,9 @@
   .main-content
     #content.container
       = yield
-= render(template: "layouts/application")
+
+= render \
+  template: "layouts/application",
+  locals: { \
+    viewport_initial_scale: ((action_name === 'new_auth_token') && '1.0') \
```
mixonic: This limits these changes to the login page, and ensures they do not impact the dashboard itself.

PR #366, opened 2017-12-07
Create background job to clean up stalled uphold codes and access params

### LINKED ISSUE #224: Create background job to clean up stalled Uphold processing
There are a couple scenarios in which a publisher may have a stalled Uphold state:
1) If the Uphold code has not been exchanged for access parameters within the 5 minute window
2) If the access parameters have been acquired but not successfully sent to eyeshade

For scenario 1 we should clear out any codes that are more than 5 minutes old.

Scenario 2 should not happen unless eyeshade has been offline for a while, or there has been a communication problem with eyeshade for a particular publisher. I'd recommend being much less aggressive than in scenario 1 - perhaps by clearing out access parameters that are older than an hour or two.

To accomplish these goals we'll need a new timestamp related to Uphold changes - perhaps `uphold_updated_at`. We'll also need a new background job to filter jobs by Uphold state and this timestamp.

nvonpentz:
Resolves #224

### Changes
* Create migration for publisher `uphold_updated_at`

* Set `uphold_updated_at` on all changes in the Uphold registration flow

* Add two scopes to Publisher model, `:has_stalled_uphold_code` and `:has_stalled_access_params`

* Add CleanStalledUpholdCodesAndAccessParametersJob that filters by these scopes and removes stalled codes/access params, logs

* Tell sidekiq to run this job at 4am every day

* Add tests for the cleaning job

* Append logic to existing tests in publishers controller that verify the job runs correctly

### Discussion
I'm hoping for feedback on these points in addition to other criticism:

- How frequently should we run the background job?  Currently runs once a day at 4am.

- Should we log the ID of each publisher who's codes are cleaned, or just the number of codes cleaned? Currently it does both.

- Should we use a `private` method within the CleanStalledUpholdCodesAndAccessParametersJob class to create the filters instead of the scopes on the model `:has_stalled_uphold_code` and `:has_stalled_uphold_access_parameters`?

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:
I added tests for the job that verify codes/params are removed if five minutes have passed (codes) or two hours have passed (params), and not if otherwise.  The model validations do a lot of the work for us.

I also ran the job and added a check at the end of two existing tests in the publishers controller to better simulate real world conditions.

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/jobs/clean_stale_uphold_data_job.rb
+++ b/app/jobs/clean_stale_uphold_data_job.rb
@@ -1,9 +1,9 @@
-class CleanStalledUpholdCodesAndAccessParametersJob < ApplicationJob
+class CleanStaleUpholdDataJob < ApplicationJob
   queue_as :scheduler
 
   def perform
     # clear uphold codes sitting for over 5 minutes   
-    publishers = Publisher.has_stalled_uphold_code
+    publishers = Publisher.has_stale_uphold_code
     n = 0
     publishers.each do |publisher|
       raise if publisher.uphold_status != :code_acquired
@@ -13,18 +13,19 @@ class CleanStalledUpholdCodesAndAccessParametersJob < ApplicationJob
       n += 1
       Rails.logger.info("Cleaned stalled uphold code for #{publisher.brave_publisher_id}.")
     end
-    Rails.logger.info("CleanStalledUpholdCodesAndAccessParametersJob cleared #{n} stalled uphold codes.")
+    Rails.logger.info("CleanStaleUpholdDataJob cleared #{n} stalled uphold codes.")
 
     # clear uphold access params sitting for over 2 hours
-    publishers = Publisher.has_stalled_uphold_access_parameters
+    publishers = Publisher.has_stale_uphold_access_parameters
     n = 0
     publishers.each do |publisher|
       raise if publisher.uphold_status != :access_parameters_acquired
       publisher.uphold_access_parameters = nil
       publisher.uphold_updated_at = Time.now
       publisher.save!
+      n += 1
       Rails.logger.info("Cleaned stalled uphold access parameters for #{publisher.brave_publisher_id}.")
     end
-    Rails.logger.info("CleanStalledUpholdCodesAndAccessParametersJob cleared #{n} stalled uphold access parameters.")
+    Rails.logger.info("CleanStaleUpholdDataJob cleared #{n} stalled uphold access parameters.")
   end
 end
--- a/app/jobs/clean_stale_uphold_data_job.rb
+++ b/app/jobs/clean_stale_uphold_data_job.rb
@@ -2,6 +2,7 @@ class CleanStaleUpholdDataJob < ApplicationJob
   queue_as :scheduler
 
   def perform
+    require "sentry-raven"
     # clear uphold codes sitting for over 5 minutes   
     publishers = Publisher.has_stale_uphold_code
     n = 0
@@ -11,9 +12,11 @@ class CleanStaleUpholdDataJob < ApplicationJob
       publisher.uphold_updated_at = Time.now
       publisher.save!
       n += 1
-      Rails.logger.info("Cleaned stalled uphold code for #{publisher.brave_publisher_id}.")
+      Rails.logger.info("Cleaned stalled uphold code for #{publisher.id}.")
+      Raven.capture_message("Cleaned stalled uphold code for #{publisher.id}.")
     end
     Rails.logger.info("CleanStaleUpholdDataJob cleared #{n} stalled uphold codes.")
+    Raven.capture_message("CleanStaleUpholdDataJob cleared #{n} stalled uphold codes.")
 
     # clear uphold access params sitting for over 2 hours
     publishers = Publisher.has_stale_uphold_access_parameters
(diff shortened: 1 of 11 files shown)


app/jobs/clean_stalled_uphold_codes_and_access_parameters_job.rb
```diff
@@ -0,0 +1,30 @@
+class CleanStalledUpholdCodesAndAccessParametersJob < ApplicationJob
```
dgeb: minor: this name is a mouthful - perhaps `CleanStalledUpholdDataJob` or even `CleanStaleUpholdDataJob`?

app/models/publisher.rb
```diff
@@ -55,6 +55,20 @@ class Publisher < ApplicationRecord
...
+    where.not(encrypted_uphold_code: nil)
+    .where("uphold_updated_at < ?", Time.now - 5.minutes)
+  }
+
+  # publishers that have access params that havent accepted by eyeshade
+  # can be cleared after 2 hours
+  scope :has_stalled_uphold_access_parameters, -> {
+    where.not(encrypted_uphold_access_parameters: nil)
+    .where("uphold_updated_at < ?", Time.now - 2.hours)
+  }
```
dgeb: These scopes are tested indirectly with the job. Perhaps unit test them as well?

app/jobs/clean_stalled_uphold_codes_and_access_parameters_job.rb
nvonpentz: Haha yes that's better, I will make the change.

app/models/publisher.rb
nvonpentz: That's a good idea, I will add them.  I'll also add some tests that make sure a publisher can  successfully acquire new uphold codes and access params if they have already been cleared.

app/models/publisher.rb
```diff
@@ -55,6 +55,20 @@ class Publisher < ApplicationRecord
 
   scope :created_recently, -> { where("created_at > :start_date", start_date: 1.week.ago) }
 
+  # publishers that have uphold codes that have been sitting for five minutes
+  # can be cleared if publishers do not create wallet within 5 minute window
+  scope :has_stalled_uphold_code, -> {
```
ayumi: Minor: As these scopes are only used in one place we could consider moving them to the one place only.

app/jobs/exchange_uphold_code_for_access_token_job.rb
```diff
@@ -12,6 +12,7 @@ def perform(publisher_id:)
       # The code acquired from https://uphold.com/authorize is only good for one request and times out in 5 minutes
       # it should now be cleared
       publisher.uphold_code = nil
+      publisher.uphold_updated_at = Time.now
```
ayumi: Minor: We've repeated this line -- to update the uphold_updated_at -- several times so I'd consider an Publisher model callback to update it automatically e.g.:
`before_validation :update_uphold_updated_at, if: -> { exists? && uphold_code_changed? }`

app/jobs/clean_stalled_uphold_codes_and_access_parameters_job.rb
```diff
@@ -0,0 +1,30 @@
...
+    # clear uphold codes sitting for over 5 minutes   
+    publishers = Publisher.has_stalled_uphold_code
+    n = 0
+    publishers.each do |publisher|
+      raise if publisher.uphold_status != :code_acquired
+      publisher.uphold_code = nil
+      publisher.uphold_updated_at = Time.now
+      publisher.save!
+      n += 1
+      Rails.logger.info("Cleaned stalled uphold code for #{publisher.brave_publisher_id}.")
```
ayumi: - We should log the publisher.id
- Also I think we should send events to Sentry ([Raven.capture_message](https://docs.sentry.io/clients/ruby/usage/#optional-attributes)) to become aware of the extent of this issue; then we'll review in the medium future.

nvonpentz:
Thank you @dgeb and @ayumi for your great feedback.  I've made all the changes you requested and pushed the updated branch.

@ayumi, the only thing I did not change that you mentioned was moving the scopes.  I found two ways to do this:

  1. [Move this scopes into a concern and include it in publisher.rb](https://stackoverflow.com/questions/25444838/move-scopes-from-model-to-some-module-in-rails-4)
  2. [Make use of the Rails `ScopeRegistry` by moving scopes to `models/publisher/scopes/`](https://rmosolgo.github.io/blog/2015/03/12/move-activerecord-scopes-into-separate-files/)

Both of these methods would reduce the visual clutter in the model, but don't restrict the use of these scopes outside of the job.  Should I use one of these methods, or did you have something in mind?

app/jobs/clean_stale_uphold_data_job.rb
```diff
@@ -0,0 +1,34 @@
...
+    publishers.each do |publisher|
+      raise if publisher.uphold_status != :code_acquired
+      publisher.uphold_code = nil
+      publisher.save!
+      n += 1
+      Rails.logger.info("Cleaned stalled uphold code for #{publisher.id}.")
+      Raven.capture_message("Cleaned stalled uphold code for #{publisher.id}.")
+    end
+    Rails.logger.info("CleanStaleUpholdDataJob cleared #{n} stalled uphold codes.")
+    Raven.capture_message("CleanStaleUpholdDataJob cleared #{n} stalled uphold codes.")
```
ayumi: the correct way to use capture_message is to do
```rb
Raven.capture_message("A nonchanging message", param_one: "apple")
```

having the messages share the same content is necessary to make sentry work nicely

nvonpentz: Noted, thanks.

PR #367, opened 2017-12-07
Add timeout message for Uphold status polling

### LINKED ISSUE #223: handle `PUT /v2/publishers/{publisher}/wallet` gracefully
if the publishers site gets an error, it stays in this state "for ever": 
image

with `encrypted_uphold_code_iv`, `encrypted_uphold_access_parameters`, and  `encrypted_uphold_access_parameters_iv` non-NULL and the other `encrypted_uphold_*` parameters equal to `NULL`

kpfefferle:
Resolves #223 

When the status check polling for the Uphold wallet connection keeps failing (whether Publishers, Eyeshade, or Uphold is having trouble completing wallet verification), this introduces a timeout message that displays after 30 seconds letting the Publisher know that we're having some trouble:

Uphold Timeout

This gets the Publisher out of the visually "stuck" state while the background job keeps attempting to verify the Uphold wallet. Should the Publisher reload the page or logout/login, then they will again see the polling message unless/until it again times out.

The worst case scenario of Uphold verification not working at all ultimately will be handled by @nvonpentz's work on #366 which would then result in this page returning to the original "connect to Uphold" prompt after the stalled verification codes have been cleared out.

cc: @jenn-rhim for approval on timeout message wording

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

If you want to make your local dev environment timeout during this polling, temporarily comment out this line to prevent the Uphold verification from completing:
https://github.com/brave-intl/publishers/blob/4b6575c8fb6b1f104f8a879802dc9c7fb40e2e70/app/jobs/upload_uphold_access_parameters_job.rb#L9

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -396,6 +396,7 @@ class PublishersController < ApplicationController
         render(json: {
           status: publisher_status(publisher).to_s,
           status_description: publisher_status_description(publisher),
+          timeout_message: publisher_status_timeout(publisher),
           uphold_status: publisher.uphold_status.to_s,
           uphold_status_description: uphold_status_description(publisher)
         }, status: 200)
--- a/app/helpers/publishers_helper.rb
+++ b/app/helpers/publishers_helper.rb
@@ -146,6 +146,15 @@ module PublishersHelper
     end
   end
 
+  def publisher_status_timeout(publisher)
+    case publisher_status(publisher)
+    when :uphold_processing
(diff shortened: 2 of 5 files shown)


jenn-rhim:
@kpfefferle 
Catch-all message:

"We are having trouble connecting to your Uphold wallet. Please check back later."

kpfefferle:
@jenn-rhim Thanks for the feedback! Updated accordingly:

Uphold Timeout

Issue #348, opened 2017-12-04
Determine solution for growing js code


nvonpentz:
As @dgeb noted in #342, we are in need of a solution to better manage the javascripts.
> I think we're getting to a level of dynamism on the frontend where we should consider introducing a framework such as Glimmer or Ember to replace the vanilla JS. I'm getting uncomfortable with the lack of module management for JS and the use of globals. We also need to improve the testing of the growing JS codebase.

**Problems**
- Lack of module management
- Shared global variables across unrelated scripts
- Lack of consistent javascript testing

**Proposed solutions**

I will update this post as we expand on the problems and discuss potential solutions.

mixonic:
Currently the codebase uses Rails 5 + Sprockets + Uglify gem + UglifyJS v1 for production asset compilation. Unfortunately UglifyJS v1 is not compatible with modern post-ES5 JavaScript (`let`, `() => {}` functions etc).

We explored three options to resolve this, and found neither to be trivial:

* There is a flag `harmony: true` which can be passed to the Uglify gem. This enables experimental ES6 support, however it does this by leveraging an unsupported and deprecated JavaScript library. This does not sound like a good way to move forward.
* Sprockets can potentially use babel as a compilation step, and once in Babel we could use [babel-minify](https://github.com/babel/minify) to do minification instead of uglify. I'm unsure if any versioning issues exist here (what version of babel will sprockets be happy with, what version does babel-minify require).
* Adopt the "happy path" for Rails, the [webpacker](https://github.com/rails/webpacker) gem. Unfortunately this means migrating away from the Sprockets build pipeline and related features (like importing a file with a comment). This is probably the best direction to head for build pipeline stability if we wanted to continue adding JS to the Rails app directly.

ayumi:
I discussed briefly with @dgeb on how to handle JS here, and I think I'd prefer to go with the minimal Vanilla JS strategy.

So the last suggestion above on "happy path" webpacker gem, this sounds appealing to me. I think Webpack is cool also.

mixonic:
@ayumi Can you say more about "minimal Vanilla JS strategy"? I'm not sure what you mean. "Vanilla JS" would probably mean dropping Sprockets, which we currently use. I think we should keep using a minification solution (like uglify v2) as well.

Yeah webpacker isn't trivial to adopt but it is the Railsy happy path. I think it would an improvement over the current sprockets system.

ayumi:
[...]

PR #369, opened 2017-12-07
Introduce capybara and headless chrome based tests

mixonic:
This PR introduces full-stack feature testing, including JavaScript code. It is roughly related to #348.

Technical summary:

* Introduce minitest-rails-capybara for a feature test harness.
* Make all feature tests (which inherit from `Capybara::Rails::TestCase`) always run the full browser environment.
* Introduce Capybara for driving selenium and browser tests.
* Add chromedriver-selenium which permits us to use headless chrome.
* Write two simple proof-of-concept tests, including one which requires authentication.
* By default have webmock permit network connects (which Capybara uses), however disable them for tests using the `ActionDispatch::IntegrationTest` base class (most of ours, the integration tests).

w/ @kpfefferle.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -118,6 +118,9 @@ group :development, :test do
   # Sweet REPL. To use, drop in "binding.pry" anywhere in code.
   gem "pry"
   gem "mocha"
+  gem "minitest-rails-capybara"
+  gem "capybara-selenium"
+  gem "chromedriver-helper"
 end
 
 group :production, :staging do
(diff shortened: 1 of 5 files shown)


ayumi:
travis is not stoked

mixonic:
@ayumi give it another shot, thanks :-)

ayumi:
I had to install chromium on debian, but now it works :+1: 
I'll add a note to the README.

PR #386, opened 2017-12-11
Have CI precompile assets

mixonic:
Recent staging builds have failed https://dashboard.heroku.com/apps/bat-publishers-staging/activity/builds/a9f62851-f233-4542-8ed0-4a9da485a8e1 The error causing this is a syntax error raised when the Uglify gem attempts to parse modern JavaScript code (in this case `let`).

In this PR, we have two goals:

* Fix the bug for now by not using modern JavaScript (`let`)
* Ensure that CI will fail in the future for other developers.

We can further pursue plans for being able to use modern JS via https://github.com/brave-intl/publishers/issues/348

/cc @kpfefferle 

TODO:

* [x] Confirm Travis failure as expected
* [x] Remove ES6 usage from JS

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -7,6 +7,8 @@ Rails.application.configure do
   # and recreated between test runs. Don"t rely on the data there!
   config.cache_classes = true
 
+  config.assets.js_compressor = :uglifier
+
   # Do not eager load code on boot. This avoids loading your whole application
   # just for the purpose of running a single test. If you are using a tool that
   # preloads Rails for running tests, you may have to set it to true.
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,12 @@ bundler_args: --jobs=3 --retry=3 --without production --without staging
 dist: trusty
 addons:
   postgresql: "9.6"
+  chrome: "stable"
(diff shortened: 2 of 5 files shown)


mixonic:
I confirmed that the build fails when asset precompilation fails:

screen shot 2017-12-11 at 2 10 15 pm

screen shot 2017-12-11 at 2 10 21 pm

screen shot 2017-12-11 at 2 10 03 pm

config/environments/test.rb
```diff
@@ -7,6 +7,8 @@
   # and recreated between test runs. Don"t rely on the data there!
   config.cache_classes = true
 
+  config.assets.js_compressor = :uglifier
```
mixonic: A side-effect of this change is that developers running tests locally will probably have a slower test suite. Before, the JS in test mode was only  concatenated. Now, it is also uglified (which is not fast). The suite locally for me was about 50 seconds.

The work-around here would be to create a new environment just for building these assets, but I fear it would get out of sync over time. This seems like a good iterative step.

ayumi:
this is good for now :open_umbrella: 

i have to say, though, our tests are getting kind of slow

PR #388, opened 2017-12-12
Back off statement sync retries and stop entirely after 3 minutes

### LINKED ISSUE #368: Improve polling of eyeshade for generated statements
A background job (`SyncPublisherStatementJob`) is used to retrieve publisher statements from eyeshade. Currently this job is not very sophisticated: on failure it simply retries itself every 3 seconds.

@mrose17 requested the following backoff strategy be used instead:

> it uses a backoff strategy and tries for no more than 3m. i'm thinking something along the lines of starting at 3s then going up to 10s until 3m is elapsed and then doing a sentry/slack call

dgeb:
As per @mrose17's instructions in #368, if the `SyncPublisherStatementJob` fails, it will continue to retry itself for 3 minutes, gradually backing off its wait time from 3 seconds to 10 seconds.

Also adds more complete tests for this job and its associated services.

Fixes #368 

/cc @ayumi 

Submitter Checklist:

- [X] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/jobs/sync_publisher_statement_job.rb
+++ b/app/jobs/sync_publisher_statement_job.rb
@@ -1,14 +1,40 @@
 class SyncPublisherStatementJob < ApplicationJob
   queue_as :default
 
-  def perform(publisher_statement_id:)
+  def perform(publisher_statement_id:, first_attempt: nil)
+    if first_attempt
+      time_elapsed = Time.now.to_i - first_attempt
+      if time_elapsed > 3.minutes
+        Raven.capture_message('SyncPublisherStatementJob timed out', publisher_statement_id: publisher_statement_id)
+        return
+      end
+    else
+      first_attempt = Time.now.to_i
+      time_elapsed = 0
+    end
+
     publisher_statement = PublisherStatement.find(publisher_statement_id)
 
     PublisherStatementSyncer.new(publisher_statement: publisher_statement).perform
 
     publisher_statement.reload
+
     unless publisher_statement.contents.present?
-      SyncPublisherStatementJob.set(wait: 3.seconds).perform_later(publisher_statement_id: publisher_statement.id)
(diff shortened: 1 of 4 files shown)


app/jobs/sync_publisher_statement_job.rb
```diff
@@ -1,14 +1,32 @@
 class SyncPublisherStatementJob < ApplicationJob
   queue_as :default
 
-  def perform(publisher_statement_id:)
+  def perform(publisher_statement_id:, first_attempt: nil)
+    if first_attempt
+      time_elapsed = Time.now.to_i - first_attempt
+      return if time_elapsed > 3.minutes
```
ayumi: should we send a sentry message here?

```rb
Raven.capture_message('SyncPublisherStatementJob timed out', publisher_statement_id: publisher_statement_id)
```

app/jobs/sync_publisher_statement_job.rb
```diff
@@ -1,14 +1,32 @@
...
+
     publisher_statement = PublisherStatement.find(publisher_statement_id)
 
     PublisherStatementSyncer.new(publisher_statement: publisher_statement).perform
 
     publisher_statement.reload
+
     unless publisher_statement.contents.present?
-      SyncPublisherStatementJob.set(wait: 3.seconds).perform_later(publisher_statement_id: publisher_statement.id)
+      if time_elapsed < 20.seconds
```
ayumi: minor: this might be cleaner if extracted in a small private method

app/jobs/sync_publisher_statement_job.rb
dgeb: Sure, that makes a lot sense 👍

app/jobs/sync_publisher_statement_job.rb
dgeb: ok, will do

dgeb:
@ayumi thanks for the review and suggestions! I think it's ready for another look :eyes:

PR #406, opened 2017-12-15
Ensure that omniauth failures are properly redirected

### LINKED ISSUE #399: new_session_path for OmniAuthCallbacksController is not defined
### Description
When a user is denied authorization by Omni Auth, Omni auth looks for a `new_session_path` which is not defined.

We should define `new_session_path` in the OmniAuthCallbacksController to handle this failure mode.

### Steps To Reproduce
You will have to use a google account you have not previously used with the google api project from console.developers.google.com
1. Go to the landing page
2. Click "login"
3. Click "login with google"
4. Choose an account, and click.  This will bring up:
image
5. Click "cancel"

We can also trigger this error if we follow the same steps (1-4) and instead:
1. Click "allow".  This will bring up the dashboard
2. Go back one page in your browser

#### Expected Result
If they click "cancel" they should be redirected back to login page with flash message saying authentication through google is required 

If they trigger this by going back in the browser, they should be redirected to the landing page.

#### Actual result
White screen

dgeb:
Authenticated publishers should be redirected to the email verified
page, while unauthenticated publishers should be redirected to the
landing page.

Fixes #399 

/cc @nvonpentz @ayumi 

Submitter Checklist:

- [X] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/publishers/omniauth_callbacks_controller.rb
+++ b/app/controllers/publishers/omniauth_callbacks_controller.rb
@@ -119,5 +119,15 @@ module Publishers
       sign_out(current_publisher)
       redirect_to '/', notice: t('youtube.oauth_error')
     end
+
+    def after_omniauth_failure_path_for(scope)
+      publisher = current_publisher
+
+      if publisher
+        email_verified_publishers_path
+      else
+        '/'
+      end
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


nvonpentz:
Looks good to me, covers both failure modes 👍

PR #410, opened 2017-12-18
Explain why we need Google access on choose verification type page

### LINKED ISSUE #409: Add explanation for why we need to access a youtube publisher's Google account
We have had many errors corresponding to when a user clicks "Cancel" on this screen:
image

The original error that was reporting this has been patched in #406, but it raises the question, why have [~200 people](https://sentry.io/brave-software/publishers-production-t2/?query=is%3Aunresolved+new_session&statsPeriod=14d) clicked "Cancel"?

I suspect it's because we do not give any reason to why we need access to their their google account, scaring off privacy concerned publishers.  A user in the [r/BATProject reddit](https://www.reddit.com/r/BATProject/comments/7dfa22/youtube_demonetization_batbrave_awareness/) said this:
> FYI I was going to sign up to Brave Payments, then you asked me to sign into my Google account with clear no explanation (on that page). You may want to clarify to people why you are wanting access to their Google account. I cancel until I could learn more info

We should add more specific instructions in two places, the choose verification page, and the oauth dialog that google brings up (pictured above).

nvonpentz:
Resolves #409.

#409 specified that we add the explanation in both choose verification page, as well as the google oauth consent screen.  This covers only the verification screen, as it is not possible to add a message to the consent screen as we previously thought.

image

### Changes
* Add new translation for the note
* Add the note to email_verified.html.slim
* Add the styles to publishers.sass

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/assets/stylesheets/application/publishers.sass
+++ b/app/assets/stylesheets/application/publishers.sass
@@ -739,6 +739,13 @@ body[data-controller="u2f_registrations"]
       background: url(asset-path("img-indi-creator@1x.png")) center 66px no-repeat
       &::before
         content: 'YOUTUBE CREATOR'
+      .choice-description
+        p
+          margin-bottom: 5px
+      p.text-center.note-text
+        line-height: 17px
+        width: 350px
+        margin: auto
 
     #taken_youtube_channel_modal
       width: 650px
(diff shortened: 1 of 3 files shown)


ayumi:
cc @jenn-rhim

PR #417, opened 2017-12-20
Add setup instructions for local eyeshade

### LINKED ISSUE #416: Add setup instructions for running app with eyeshade locally

nvonpentz:
Between these additions and the setup instructions I added to the bat-ledger README, everything should be covered.

In config/secrets.yml, we may want to change  the default for `api_eyeshade_offline` under `development:` from `true`, to instead use `<%= ENV["API_EYESHADE_OFFLINE"] %>`, so we end up toggling the secrets script instead of the config file.  In this case I would replace instruction (4).

Resolves #416.

Submitter Checklist:
- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,8 @@ These steps based on [directions at the omniauth-google-oauth2 gem](https://gith
 1. Follow the [setup instructions](https://github.com/brave-intl/bat-ledger) for bat-ledger
 2. Add `export API_EYESHADE_BASE_URI="http://127.0.0.1:3002"` to your secrets script
 3. Add `export API_EYESHADE_KEY="00000000-0000-4000-0000-000000000000"` to your secrets script
-4. In [config/secrets.yml](https://github.com/brave/publishers/blob/master/config/secrets.yml), under `development:` set `api_eyeshade_offline` to `false`
+
+To stop using Eyeshade locally, set `API_EYESHADE_BASE_URI=""`.
 
 ### Run
 
--- a/README.md
+++ b/README.md
@@ -62,6 +62,13 @@ You may need to wait up to 10 minutes for the changes to propagate.
 
 These steps based on [directions at the omniauth-google-oauth2 gem](https://github.com/zquestz/omniauth-google-oauth2#google-api-setup).
 
+### Local Eyeshade Setup
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -62,6 +62,13 @@ You may need to wait up to 10 minutes for the changes to propagate.
 
 These steps based on [directions at the omniauth-google-oauth2 gem](https://github.com/zquestz/omniauth-google-oauth2#google-api-setup).
 
+### Local Eyeshade Setup
+
+1. Follow the [setup instructions](https://github.com/brave-intl/bat-ledger) for bat-ledger
+2. Add `export API_EYESHADE_BASE_URI="http://127.0.0.1:3002"` to your secrets script
+3. Add `export API_EYESHADE_KEY="00000000-0000-4000-0000-000000000000"` to your secrets script
+4. In [config/secrets.yml](https://github.com/brave/publishers/blob/master/config/secrets.yml), under `development:` set `api_eyeshade_offline` to `false`
```
ayumi: i think we can remove the default `api_eyeshade_offline`, and instead if api_eyeshade_base_uri is not set then assume we want an offline eyeshade.

nvonpentz:
Since there were ~100 references to `:api_eyeshade_offline` in the code, I thought it would be simpler and safer to adjust the value of `api_eyeshade_offline` within `secrets.yml` to check if `API_EYESHADE_BASE_URI` has been set instead of changing all the references to `:api_eyeshade_offline`.  If it is preferred to remove all references to `:api_eyeshade_offline` instead, I can do this instead.

I tested this by setting and unsetting `API_EYESHADE_BASE_URI` and registering a new youtube channel.  When  set to `"http://127.0.0.1:3002"`, local eyeshade was used, when set to`""`, it was not.

Note: I wasn't able to generate the statements for the a publisher (domain or youtube channel) using local eyeshade.  I can't generate them using the current staging branch (without the above changes), so I don't believe its an issue with the modifications in this PR.  It could be some issue with my local copy of eyeshade since statements are generated and emailed on the live staging/production sites, but this is curious because I was able to generate statements locally a few weeks ago.

Issue #404, opened 2017-12-14
SMTP To address blank for some verification emails 


nvonpentz:
### Description
There is some case in which a verification email is scheduled, but fails to send because the `To:` field is blank.  The failing method is `PublisherMailer#verify_email`.

We should check to make sure there are working validations in all places that trigger this method. There should be no instance in which we allow an email to be scheduled if there is no email address associated with the publisher.

### Additional information
We first see this error on December 7th, which suggests recent modifications to the code might have enabled this.

~The full error report is in [sentry](https://sentry.io/brave-software/publishers-production-t2/issues/414805093/)~

This now corresponds to either of this sentry report:
https://sentry.io/brave-software/publishers-production-t2/issues/577126805

ayumi:
Good catch. This is happening pretty frequently and it affects a core email, so I think we should get to the bottom of this with a Sentry event closer to the source.

Could you add a check for blank To: headers around PublisherMailer #verify_email -- either within that mailer method or in the places which call that mailer method? If that check fails then we should send a message to Sentry.

yachtcaptain23:
@nvonpentz Would you see if this is still happening.

nvonpentz:
@yachtcaptain23 It looks like it still is.  I updated the original comment with the new sentry report.

yachtcaptain23:
Updated sentry report. Downgrading since it is very low severity https://sentry.io/organizations/brave-software/issues/890334091/?project=225116&query=publishermailer&statsPeriod=90d&utc=false

yachtcaptain23:
Duplicate of https://github.com/brave-intl/publishers/issues/2680. Will handle the fix there

PR #421, opened 2017-12-20
Add sentry log when sending verification email to blank address

nvonpentz:
When we attempt to send a verification email with a blank recipient, we now will raise an error, and log the exception on sentry.  The publisher information, and publication type is also logged.

This should help us pin down why we are attempting to send blank verification emails in the first place.

Refs #404 

### Changes
* Raise if "To" address is blank in PublisherMailer#verify_email, send to sentry
* Add corresponding test to PublisherMailerTest

Submitter Checklist:
- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/mailers/publisher_mailer.rb
+++ b/app/mailers/publisher_mailer.rb
@@ -66,6 +66,17 @@ class PublisherMailer < ApplicationMailer
   def verify_email(publisher)
     @publisher = publisher
     @private_reauth_url = generate_publisher_private_reauth_url(@publisher)
+
+    if @publisher.pending_email.blank?
+      begin
+        raise "SMTP To address must not be blank for PublisherMailer#verify_email"
+      rescue => e
+        require 'sentry-raven'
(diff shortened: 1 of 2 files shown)


app/mailers/publisher_mailer.rb
```diff
@@ -66,6 +66,17 @@ def welcome_internal(publisher)
...
     @publisher = publisher
     @private_reauth_url = generate_publisher_private_reauth_url(@publisher)
+
+    if @publisher.pending_email.blank?
+      begin
+        raise "SMTP To address must not be blank for PublisherMailer#verify_email"
+      rescue => e
+        require 'sentry-raven'
+        Raven.capture_exception(e,
+                                publisher: @publisher,
```
ayumi: minor: this will probably work, but it does rely on downstream serialization magic. once it lands, we can check sentry to confirm it's ok.

Issue #414, opened 2017-12-19
2FA Warn users after setting up U2F


mixonic:
After a user registers a U2F device they should be warned that registering a TOTP device as a fallback is strongly recommended.

jenn-rhim:
i've covered this in https://github.com/brave-intl/publishers/issues/419, which captures the breath of the 2FA home states and content. 

I'd like to use this ticket to zoom in on the implementation of the warning. 
- it would be really nice if we can apply some dynamic css element to the message box appearing under TOTP section when the user is returning from the dialog. some sort of reveal transition. suggestions are welcome!  :)

related screens:
1. when the user set up the security key first and only:
![2fa-home-a0k1](https://user-images.githubusercontent.com/4369856/34187196-067e09aa-e4e5-11e7-83c6-29017f0405ae.png)


2. when the user removes TOTP, which leaves the U2F as the only 2FA.
![2fa-disableauthapp-1](https://user-images.githubusercontent.com/4369856/34187231-33871702-e4e5-11e7-94bd-7e6b3d771f32.png)

mixonic:
This was addressed:

* The prompt to add TOTP if you only have U2F: https://github.com/brave-intl/publishers/pull/428
* The modal for confirming removal of any device (TOTP or U2F)L https://github.com/brave-intl/publishers/pull/424

This issue can be closed IMO. Please re-open if I am mistaken!

Issue #394, opened 2017-12-14
2FA U2F set up


jenn-rhim:
![4-securitykeysetup1](https://user-images.githubusercontent.com/4369856/33984166-825f1df8-e06b-11e7-81a9-4cde18c2686f.png)

![5-securitykeysetup2](https://user-images.githubusercontent.com/4369856/33984175-89121e52-e06b-11e7-91fd-c9a945bacfef.png)

![10-securitykey-setup-error1](https://user-images.githubusercontent.com/4369856/33984189-931ae85c-e06b-11e7-9e27-614150f51780.png)

![11-securitykey-setup-error2](https://user-images.githubusercontent.com/4369856/33984191-9712e23e-e06b-11e7-82d8-5e40a8f5cb25.png)

**Remove Security Key: when TOTP is still in effect** 
![18-2fa-removekey-1](https://user-images.githubusercontent.com/4369856/34025851-182cabc0-e107-11e7-8ce5-39f905d706c9.png)

**Remove Security Key: no TOTP**
![19-2fa-removekey-2](https://user-images.githubusercontent.com/4369856/34025876-34481e48-e107-11e7-86cc-b2815abaf5a5.png)

mixonic:
The removal warnings are completed. The only remaining work here is the icons and style of notices/warnings during registration.

mixonic:
I'd like to reconcile how errors are presented here with https://github.com/brave-intl/publishers/issues/411

PR #424, opened 2017-12-21
Add modal confirmation for 2FA changes

mixonic:
Per mocks in https://github.com/brave-intl/publishers/issues/414 and https://github.com/brave-intl/publishers/issues/394 prompt the user to confirm their action when removing a 2FA registration. This flow is triggered by clicking "Disable" when a TOTP registration exists or "Remove" for any one U2F device. In each case the copy the users sees may vary, however they are always provided a list of what 2FA methods will exist after confirmation.

A technical summary:

* I've added an abstraction which allows developers to create contents for possible modals in tags `<script id="my-modal-content" type="text/html">`. To open these modals add the attribute`data-js-confirm-with-modal` on any link to specify what `id` to look for. For example to open the example script tag above: `<a data-js-confirm-with-modal="my-modal-content">`. Only when the user confirms the action will the `a` tag be permitted to perform other event handlers or its default action. Further documentation is in the source.
* Scope a large amount of publishers CSS which was being applied to the 2FA pages back to only the publishers controller.
* Add `.warning` and `strong.error` that sets `color` to a newly added `$braveRed`.
* There are feature test for the popup. I've changed the test runner to use truncation for fixtures instead of transaction during feature tests. This is requisite for DB mutations in feature tests not to leak out of the test.

Screenshots of the UI follow.

**/publishers/two_factor_registrations page with TOTP and a U2F key**

screen shot 2017-12-21 at 1 47 29 pm

**Confirmation to delete TOTP when a U2F key exists**

screen shot 2017-12-21 at 1 47 33 pm

**Confirmation to delete a U2F key when TOTP key exists**

screen shot 2017-12-21 at 1 47 39 pm

**Confirmation to delete TOTP when it is the last 2FA**

screen shot 2017-12-21 at 1 47 46 pm

**Confirmation to delete a U2F key when it is the last 2FA**

screen shot 2017-12-21 at 1 48 05 pm

**A confirmation screen on iPhone X**

screen shot 2017-12-21 at 1 48 47 pm

TODO:

* [x] Brush up  some of the modal documentation in source
* [x] Add strings to the translations file
* [x] Update this PR with screenshots

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/assets/javascripts/modal_shared.js
+++ b/app/assets/javascripts/modal_shared.js
@@ -1,0 +1,142 @@
+;(function() {
+  'use strict';
+
+  /*
+   * Implement a simple confirm/deny modal system on top of the md- modal
+   * classes. This relies on the .js-shared-modal div in the application
+   * layout.
+   *
+   * To use this system you'll need a template and a trigger. For example:
+   *
+   * ```html
+   * <script type="text/html" id="my-template">
+   *   <h3>This headline appears in a modal!</h3>
+   * </script>
+   * <a href="/some-destination" data-js-confirm-with-modal="my-template">
+   *   Visit /some-destination, but first confirm
+   * </a>
+   * ```
+   *
+   * When the trigger is clicked the value of `data-js-confirm-with-modal` is
+   * used to find a template. That template is rendered in the modal popup.
+   * Any default behavior or other click event handlers on the trigger do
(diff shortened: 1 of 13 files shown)


mixonic:
This is ready to go, I believe.

Issue #419, opened 2017-12-20
2FA Home: content & various states


jenn-rhim:
I had to make minor updates on the content in Hardware security keys section to make the 'browser limitation' blurb more noticeable.  -- this is to prevent user frustration if they discover it after the fact.  

- reference mocks below indicating different states of 2FA main page + the contextual help

--------------------------

## Two-factor Authentication

Two-factor authentication (2FA) is a method of confirming your identity by using two different forms of verification when you access Brave Payments in order to increase security (Recommended for protecting your account).

**App on your phone**
Use an app on your phone to get an authentication code. You will be asked to type in this code when logging in. 

**Hardware security keys**
Security key is a small device that connects to your computer via a USB port and works with FIDO Universal 2nd Factor (U2F). You will be asked to insert and press the key instead of typing in a code. Currently, security key is supported by Google Chrome and Opera. 
*A list of security keys that have been tested

--
**Limited Browser Support** contextual help flyout
This means you will need to update to the latest Google Chrome or Opera in order to set up and log in using security keys. More browsers will start supporting the security key in the near future. We will inform you on the changes.

--
**Tested Security Keys** contextual help flyout
The following products have been tested and proven to work with our system. 
- YubiKey
- Ledger Blue
- Product 3
- Product 4
- Product 5
--------------------------

**1. 2FA Disabled**
![2fa-home-a0k0](https://user-images.githubusercontent.com/4369856/34271774-a0321cae-e642-11e7-8900-82e700ae4663.png)

2. TOTP Enabled; U2F disabled
![2fa-home-a1k0](https://user-images.githubusercontent.com/4369856/34186816-d334cc3e-e4e2-11e7-88a8-a5a8d2d95258.png)

3. TOTP disabled; U2F enabled
![2fa-home-a0k1](https://user-images.githubusercontent.com/4369856/34186810-cdc3cc28-e4e2-11e7-9bfd-ce662be2bbd7.png)

4. Both enabled
![2fa-home-a1k1](https://user-images.githubusercontent.com/4369856/34186821-d7d3a076-e4e2-11e7-8af9-ff7e5c0caa41.png)

5. Contextual help for the supported browser
![2fa-overlay-browser](https://user-images.githubusercontent.com/4369856/34186827-dcf399bc-e4e2-11e7-988f-479dafe1f94b.png)

6. Contextual help for the supported key list
[...]

PR #428, opened 2017-12-22
Add notable prompt for TOTP fallback

mixonic:
Per https://github.com/brave-intl/publishers/issues/419 show a recommendation that the user should enable TOTP if they have only a single U2F device registered. If the user has 0 or more than 1 devices registered they see the a tamer message.

I've diverged from the mock and simply used the `well` class for bootstrap.

**Prompt when there is only one U2F device**

screen shot 2017-12-21 at 3 59 03 pm

**When 0 devices ("Authenticator app has not been set up" is the same for 2+)**

screen shot 2017-12-21 at 4 13 24 pm

Reviewer Checklist:

Tests

Security:


test/controllers/u2f_registrations_controller_test.rb
```diff
@@ -56,17 +56,38 @@ def canned_u2f_response
...
+    publisher = publishers(:completed)
+    u2f_registration = u2f_registrations(:default)
+    publisher.u2f_registrations << u2f_registration
+
+    sign_in publisher
+    get two_factor_registrations_path
+
+    assert_response :success
+    assert_match u2f_registration.name, response.body
+    assert_match /Set up an authenticator as\sthe secondary 2FA/, response.body
```
ayumi: minor: it may be better to check for tags with [data-test-id] in case strings change

test/controllers/u2f_registrations_controller_test.rb
```diff
@@ -56,17 +56,38 @@ def canned_u2f_response
...
+    additional_u2f_registration = u2f_registrations(:additional)
+    publisher.u2f_registrations << additional_u2f_registration
+
     sign_in publisher
     get two_factor_registrations_path
 
     assert_response :success
     assert_match u2f_registration.name, response.body
+    assert_match additional_u2f_registration.name, response.body
+    assert_match "Authenticator app has not been set up", response.body
```
ayumi: minor: same comment as above

PR #429, opened 2017-12-22
Add endpoint to notify unverified pubs they have money waiting

### LINKED ISSUE #375: [WIP] Email - Threshold met for site operator
**Conditions :** 
A domain exists which has acquired $10 worth of BAT and has thus far not started verification.

**Frequency :** 
Monthly until verification is complete.

**Subject :** 
 Money is waiting for <_domain_>, verify your site with Brave to claim your funds!

**Body :** 

Salutation : 
Hi!

Paragraph : 
Your site has accumulated quite a bit of fundage and we’d love to hand it over.    Please visit https://publishers.basicattentiontoken.org/ to start the verification process.  There are no fees, no gimmicks, we just want you to get paid and keep making awesome content.

Call-to-Action Button Text : 
Verify My Site

Post-script : 
Got questions? Visit https://basicattentiontoken.org/ to learn more about how the micropayments work.

We look forward to having you join the thousands of domains and YouTube channels already receiving funds from their audience.

nvonpentz:
A `POST` to `/api/publishers/notify_unverified` with the params:
````
publisher_type: "domain",
params: {
    domain: "<domain>"
}
````
Will scrape the WHOIS information for `<domain>` and send a notification email to each distinct valid email address found.  Currently, `publisher_type` can only be `"domain"` until we determine a solution for #372.  I chose this format for the params above because it matches those in the existing `PublisherNotifier` service class for verified publisher notifications.

I chose not to use the existing `PublisherNotifier` service class for this endpoint because that endpoint required an existing publisher record.  We would have to create a publisher record for each valid email address found for the unverified publisher, or refactor the `PublisherNotifier` class to not take a publisher record at all.  The requirements of these two seem different enough that they require their own endpoint and service class.

If we can't scrape any valid emails, no notification emails are sent and we return a 500.  If the params in the POST are malformed or incomplete, no emails are sent and we return a 400.

This is the notification email:
image

Resolves #375.
References #372

Submitter Checklist:
- [x] Tagged reviewers.

Test Plan:
Because the 'whois' gem does not support mocking, I separated the whois lookup into a service, `GetWhoisEmailsForDomain`.  The tests in `Api::PublishersControllerTest` mock the response from this service and perform some validations to make sure we do not send duplicate emails, or attempt to send to malformed email addresses.

There are also test for both new emails, `unverified_reached_threshold` and `unverified_reached_threshold_internal`.

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -81,9 +81,9 @@ gem "clipboard-rails"
 gem "rotp", "~> 3.3"
 
 # WHOIS lookup for unverified publishers
-gem "whois", "~> 4.0"
+gem "whois", "~> 4.0", require: false
 
-gem 'whois-parser', '~> 1.0'
+gem "whois-parser", "~> 1.0", require: false
 
 group :development, :test do
   # Call "byebug" anywhere in the code to stop execution and get a debugger console
--- a/Gemfile
+++ b/Gemfile
@@ -80,6 +80,11 @@ gem "clipboard-rails"
 # One-time passwords for 2fa
 gem "rotp", "~> 3.3"
 
+# WHOIS lookup for unverified publishers
+gem "whois", "~> 4.0"
+
+gem 'whois-parser', '~> 1.0'
+
 group :development, :test do
   # Call "byebug" anywhere in the code to stop execution and get a debugger console
   gem "byebug", platform: :mri
--- a/app/services/publisher_notifier_unverified.rb
+++ b/app/services/publisher_notifier_unverified.rb
@@ -1,9 +1,7 @@
 class PublisherNotifierUnverified < BaseService
-  attr_reader :notification_params, :notification_type
-
   PUBLISHER_TYPES = %w(domain) # TODO: youtube
 
-  def initialize(publisher_type:, publisher_id: )
+  def initialize(publisher_type:, publisher_id:)
     @publisher_id = publisher_id
     @publisher_type = publisher_type
 
@@ -15,16 +13,16 @@ class PublisherNotifierUnverified < BaseService
   end
 
   def perform
-    if @publisher_type == "domain"
-      return perform_domain
-    else
-      return perform_youtube
+    case @publisher_type
+    when "domain"
+      perform_domain
     end
+    # TO DO: youtube
   end
 
   def perform_domain
     domain = @publisher_id
-    contacts = GetWhoisEmailsForDomain.new(@domain).perform
+    contacts = GetWhoisEmailsForDomain.new(domain).perform
 
     if contacts.empty?
       raise NoEmailsFoundError.new("No contacts listed on whois info for '#{domain}'")
@@ -48,10 +46,6 @@ class PublisherNotifierUnverified < BaseService
       PublisherMailer.unverified_domain_reached_threshold(domain, email).deliver_later
       PublisherMailer.unverified_domain_reached_threshold_internal(domain, email).deliver_later
     end
-  end
-
-  def perform_youtube
-    # TO DO
   end
 
   class BlankParamsError < RuntimeError; end
--- a/app/services/publisher_notifier_unverified.rb
+++ b/app/services/publisher_notifier_unverified.rb
@@ -1,11 +1,13 @@
 class PublisherNotifierUnverified < BaseService
   attr_reader :notification_params, :notification_type
 
-  PUBLISHER_TYPES = %w(youtube, domain)
+  PUBLISHER_TYPES = %w(domain) # TODO: youtube
 
-  def initialize(publisher_type:, notification_params: {})
-    @notification_params = (notification_params) || {}
+  def initialize(publisher_type:, publisher_id: )
+    @publisher_id = publisher_id
     @publisher_type = publisher_type
+
+    ensure_params_exist
 
     if !PUBLISHER_TYPES.include?(@publisher_type)
       raise InvalidPublisherTypeError.new("#{@publisher_type} is an invalid publisher type")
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: Gemfile [hunk 1]
```diff
@@ -80,6 +80,11 @@ gem "clipboard-rails"
 # One-time passwords for 2fa
 gem "rotp", "~> 3.3"
 
+# WHOIS lookup for unverified publishers
+gem "whois", "~> 4.0"
+
+gem 'whois-parser', '~> 1.0'
```
ayumi: minor: prefer `"` for this project

Gemfile
```diff
@@ -80,6 +80,11 @@ gem "clipboard-rails"
 # One-time passwords for 2fa
 gem "rotp", "~> 3.3"
 
+# WHOIS lookup for unverified publishers
+gem "whois", "~> 4.0"
```
ayumi: could you update `Gemfile.lock` as a result of the new gems?

ayumi: also, you could probably do `require: false` for this (then require it manually)

app/services/publisher_notifier_unverified.rb
```diff
@@ -0,0 +1,67 @@
+class PublisherNotifierUnverified < BaseService
+  attr_reader :notification_params, :notification_type
+
+  PUBLISHER_TYPES = %w(youtube, domain)
```
ayumi: let's remove the youtube references and require that the type be `domain` for now.  we can add a comment up top noting TODO: youtube

config/locales/en.yml
```diff
@@ -216,14 +216,26 @@ en:
...
       erroneous_email_help: "If you did not make this request, you can ignore this email or contact us at "
       terms_of_service: "Terms of Service"
       medium: "Medium"
       twitter: "Twitter"
       reddit: "Reddit"
       rocketchat: "Rocket.Chat"
-
+    
+    unverified_domain_reached_threshold:
+      subject: "Money is waiting for [%{publication_title}], verify your site with Brave to claim your funds!"
```
ayumi: maybe we don't need the `[]`.. they look a little funny like it's a template email

ayumi:
needs merge conflict resolution :)

nvonpentz:
I settled on a POST to `/api/publishers/[brave_published_id]/notifications_unverified` with the params:
```
publisher_type: ["domain" | "youtube"],
publisher_id: ["example.com" | "UCtsfHRe2WQnkNH5WYJWL-Yw"]
```

Youtube is currently not implemented and will throw an error.  This is ready for review.
cc @mrose17

app/services/publisher_notifier_unverified.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    ensure_params_exist
+
+    if !PUBLISHER_TYPES.include?(@publisher_type)
+      raise InvalidPublisherTypeError.new("#{@publisher_type} is an invalid publisher type")
+    end
+  end
+
+  def perform
+    if @publisher_type == "domain"
```
ayumi: minor: we plan on adding more platforms in the future, so a `case` statement might be more sense. however, we can do that later.

app/services/publisher_notifier_unverified.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    # send notification to valid emails
+    emails.each do |email|
+      PublisherMailer.unverified_domain_reached_threshold(domain, email).deliver_later
+      PublisherMailer.unverified_domain_reached_threshold_internal(domain, email).deliver_later
+    end
+  end
+
+  def perform_youtube
+    # TO DO
```
ayumi: minor: might be better to leave out this out, or to have the method raise an error when called.

test/controllers/publishers_controller_test.rb
```diff
@@ -357,7 +357,7 @@ def request_login_email(publisher:)
     # verify brave gets an internal email copy of confirmation email
     email = ActionMailer::Base.deliveries.find do |message|
       message.to == Rails.application.secrets[:internal_email]
-      message.subject == "<Internal> #{I18n.t('publisher_mailer.confirm_email_change.subject', publication_title: '')}"
```
ayumi: minor: unrelated change? tests seem ok though

app/services/publisher_notifier_unverified.rb
```diff
@@ -0,0 +1,76 @@
+class PublisherNotifierUnverified < BaseService
+  attr_reader :notification_params, :notification_type
+
+  PUBLISHER_TYPES = %w(domain) # TODO: youtube
+
+  def initialize(publisher_type:, publisher_id: )
```
ayumi: minor: unnecessary space after `publisher_id:`

app/services/publisher_notifier_unverified.rb
```diff
@@ -0,0 +1,76 @@
+class PublisherNotifierUnverified < BaseService
+  attr_reader :notification_params, :notification_type
```
ayumi: These are unused

app/services/publisher_notifier_unverified.rb
```diff
@@ -0,0 +1,76 @@
...
+    ensure_params_exist
+
+    if !PUBLISHER_TYPES.include?(@publisher_type)
+      raise InvalidPublisherTypeError.new("#{@publisher_type} is an invalid publisher type")
+    end
+  end
+
+  def perform
+    if @publisher_type == "domain"
+      return perform_domain
```
ayumi: minor: `return` here is redundant

app/services/publisher_notifier_unverified.rb
```diff
@@ -0,0 +1,76 @@
...
+    else
+      return perform_youtube
+    end
+  end
+
+  def perform_domain
+    domain = @publisher_id
+    contacts = GetWhoisEmailsForDomain.new(@domain).perform
```
ayumi: `@domain` is undefined, did you mean `@publisher_id`?

test/controllers/publishers_controller_test.rb
nvonpentz: This test actually started to fail. I suspect it was because we removed the brackets from the email templates.  This seemed to fix it though.

app/services/publisher_notifier_unverified.rb
nvonpentz: It should never be called since 'youtube' is not an accepted `PUBlSHER_TYPE`, but I will remove it until we implement it.

test/controllers/publishers_controller_test.rb
mixonic: hm, I'm having to change this back to having a space in a current branch 🤔

PR #435, opened 2018-01-03
Add a new "incomplete implemenation" error for U2F

### LINKED ISSUE #389: U2F Browser support warning
The U2F "no browser support" message does not appear for the Brave browser. Improve the detection.

mixonic:
This error will throw for Brave, it shouldn't appear for any other known browser. Currently the timeout is three seconds, which seems generous enough for a browser like Chrome or Opera to open the iframe to the U2F extension.

Fixes #389

Examples of the error:

**Failing to register a U2F device in Brave**

screen shot 2018-01-03 at 2 44 08 pm

**Failing to authenticate with a U2F device in Brave**

screen shot 2018-01-03 at 2 42 24 pm

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/assets/javascripts/u2f_authentications.js
+++ b/app/assets/javascripts/u2f_authentications.js
@@ -34,6 +34,9 @@
         case 5: // TIMEOUT
           window.U2FShared.showError('u2f-error-timeout');
           break;
+        case 99900: // IMPLEMENTATION_INCOMPLETE
+          window.U2FShared.showError('u2f-error-implementation-incomplete');
+          break;
       }
 
       formElement.classList.remove('js-u2f-working');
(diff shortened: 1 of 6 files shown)


mixonic:
@ayumi ugh, I'll confirm that but I presume you are correct. That is pretty lame. Good catch.

I could change the `IMPLEMENTATION_INCOMPLETE` timeout to 30s and change the text to:

> Please insert your security key and re-attempt authentication. Some browsers (like Brave) have an incomplete implementation of this protocol. Please be sure to use a current release of Google Chrome or Opera.

And leave the `TIMEOUT` timeout at 30s and the message as:

> There was an unexpected timeout waiting your device to respond to the authentication request. Please re-attempt authentication and activate your security key when it is blinking.

ayumi:
@mixonic sgtm

PR #440, opened 2018-01-04
Allow publishers to send themselves a new login link

### LINKED ISSUE #381: Inform the user if the token is expired and allow them to send a new key
when the access token is expired, allow them to resend the fresh key to the email they came from in one click.

login error

nvonpentz:
### Changes
* Create custom failure app according to [this](https://github.com/plataformatec/devise/wiki/How-To:-Do-not-redirect-to-login-page-after-session-timeout) and [this](https://github.com/plataformatec/devise/wiki/How-To:-Redirect-to-a-specific-page-when-the-user-can-not-be-authenticated)
* Allow verified publishers to send themselves a new login link if they use an expired one
* Redirect session timeouts to login page
* Create new view for "send me login link page"

image

Resolves #381 

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

I modified exisiting tests that assumed attempts to claim expired auth tokens would take them to the dashboard.

I created three new tests that verify that:
1. unverified publishers attempting to use expired token are still redirected to the dashboard with a flash message
2. verified publishers are are shown the "send me a new login link" page when they attempt to use an expired token
3. publishers whose session has expired will be sent to the login page with a flash message

1 & 2 are easy to verify through the browser:
1. create a new publisher, expire their token before clicking the verification link, then click the link.  This should bring you to the homepage.
2. create a new publisher, and click the link in the verification email.  Now click that link again, you should be brought to the "send me a new login link" page and send a valid link from there.

Reviewer Checklist:

Tests

Security:


test/controllers/publishers_controller_test.rb
```diff
@@ -46,15 +46,75 @@ class PublishersControllerTest < ActionDispatch::IntegrationTest
...
  
     get url
     assert_redirected_to home_publishers_url, "precond - publisher is logged in"
 
     get url
-    assert_redirected_to root_url, "re-used URL is rejected, publisher not logged in"
+    assert_redirected_to expired_auth_token_publishers_path(publisher_id: publisher.id), "re-used URL is rejected, publisher not logged in"
+  end
+
+  test "expired login link takes unverified publishers to dashboard" do
```
ayumi: minor: s/dashboard/root, as dashboard could be confused with the publisher dashboard

test/controllers/publishers_controller_test.rb
```diff
@@ -46,15 +46,75 @@ class PublishersControllerTest < ActionDispatch::IntegrationTest
...
+    perform_enqueued_jobs do
+      post(publishers_path, params: SIGNUP_PARAMS)
+    end
+    publisher = Publisher.order(created_at: :asc).last
+
+    # expire token and attempt to cliam
+    publisher.authentication_token_expires_at = Time.now
+    publisher.save
```
ayumi: minor: `save!` generally better to fail harder

app/views/publishers/expired_auth_token.html.slim
```diff
@@ -0,0 +1,19 @@
...
+  .sub-panel.col-center
+    .row.col-center
+      .col.col-md-12
+        h3.text-center= t("publishers.login_link_expired_title")
+    .row
+      .col.col-md-8.col-md-offset-2.text-center.col-center
+        p= t("publishers.login_link_expired_message")
+        = form_for @publisher, {method: :post, url: create_auth_token_publishers_path} do |f|
+          = f.hidden_field(:brave_publisher_id, :value => @publisher.brave_publisher_id)
+          = f.hidden_field(:email, :value => @publisher.email)
```
ayumi: minor: no captcha tags, so in a very edge case users would have to fill out the captcha form. that's okay i think

lib/devise/custom_failure_app.rb
```diff
@@ -0,0 +1,32 @@
+# Overrides the Devise defaults for unauthenticated requests
+class CustomFailureApp < Devise::FailureApp
+  def redirect_url
+    warden_message = warden_options[:message]
+    attempted_path = warden_options[:attempted_path]
+    unauthenticated = warden_options[:action] == "unauthenticated"
+    has_token_claim_params = params[:token].present? && params[:id].present?
```
ayumi: Nice, I didn't realize that login links all had the publisher id inside!

PR #441, opened 2018-01-04
2FA Settings: Content, icons, and tooltips

### LINKED ISSUE #412: 2FA Contextual Help
*note: we may potentially need to create a Pub specific help section on the site in the future. till then, we'll handle extra sideline content in the form of contextual help flyout.
**any suggestions on the wording are welcome.

**current browser support limitation**
2fa-overlay-browserw

**supported/tested u2f keys**
2fa-overlay-supportedkeys

### LINKED ISSUE #419: 2FA Home: content & various states

kpfefferle:
Resolves #412 
Resolves #419 

**Content and Icons:**
Content and Icons

**Browser Tooltip:**
Browser Tooltip

**Device Tooltip:**
Device Tooltip

Submitter Checklist:

- [ ] Tagged reviewers.

Reviewer Checklist:

Tests

Security:


app/views/application/_usb.html
```diff
@@ -1,14 +1,12 @@
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="" height="75" viewBox="0 0 73 75">
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="58" height="50" viewBox="0 0 58 50">
```
mixonic: This size change will also impact the U2F authentication flow. We should confirm that looks correct before landing.

kpfefferle: I couldn't find anywhere else where this partial was being used, but if we need to we can resize via CSS in one place or another.

kpfefferle: Ah, I see it here:
https://github.com/brave-intl/publishers/blob/aa6eab1fcc7af350499f71d0878977e1f6693e06/app/views/two_factor_authentications/_u2f.html.slim#L25

kpfefferle: @mixonic I looked into this, and it seems my "new" version is actually the exact same rendered size for the icon itself, but without added padding and rotation information included in the SVG file. This seems to be an improvement in the export itself, and it still looks good when shown during auth:
USB icon during 2FA auth

ayumi:
looks good, thanks! :parasol_on_ground:

Issue #422, opened 2017-12-20
Use webpacker gem


ayumi:
Capturing discussion from #348.

As noted by @mixonic:

> Currently the codebase uses Rails 5 + Sprockets + Uglify gem + UglifyJS v1 for production asset compilation. Unfortunately UglifyJS v1 is not compatible with modern post-ES5 JavaScript (let, () => {} functions etc).

> Adopt the "happy path" for Rails, the webpacker gem. Unfortunately this means migrating away from the Sprockets build pipeline and related features (like importing a file with a comment). This is probably the best direction to head for build pipeline stability if we wanted to continue adding JS to the Rails app directly.

PR #443, opened 2018-01-05
Migrate to webpacker

mixonic:
This PR migrates all the application's JS from sprockets over to webpacker and modern ES. It creates several "packs" and uses them appropriately to only load whatever JS is required for a page. For example the `static` pack doesn't include a bunch of stuff used in the `application`.

Closes https://github.com/brave-intl/publishers/issues/422

TODO:

* [x] Some failures in the Capybara tests
* [x] Complete a smoke test of the registration process.


ayumi:
a little Gemfile.lock conflict

mixonic:
@ayumi I just rebased. This passed local smoke testing for me. I would like to continue testing it out on staging.

I did some work this afternoon to ensure we touch the existing publishers/home javascript as little as possible. This should make it easier to integrate this work with that of @dgeb. I still think the inline JS should be refactored into packs eventually.

PR #448, opened 2018-01-08
Allow users to connect with uphold when the state has stalled

### LINKED ISSUE #439: Allow cancel and retry Uphold connecting
Sometimes Uphold connection gets stuck. When that happens, we retry in the background.
A better UX might be to let the user themself cancel and retry the connection, so they feel more control over the process.

nvonpentz:
As outlined in #224, there are two cases we have to deal with:
> 1. If the Uphold code has not been exchanged for access parameters within the 5 minute window
> 2. If the access parameters have been acquired but not successfully sent to eyeshade

This PR proposes changes in the UI so it is more obvious to the user how/when to reset the connection for (1) and (2)

Last week we encountered an instance of (1).

(1) can be reproduced by commenting out `ExchangeUpholdCodeForAccessTokenJob#perform`, signing up a new publisher, and clicking "Create Uphold Wallet".  When returned to the publisher dashboard, the user was stuck on this (old) screen:

image

However, they still had the option to "Create Uphold Wallet".  If they were to click this button, it would reset the Uphold connection without problems - we do not need to do any more to clear the uphold codes.

In this PR we simply change the "Create Uphold Wallet" button to "Reconnect to Uphold", and also change the message.  The (new) screen:
image

(2) can be reproduced by commenting out the call to [UploadUpholdAccessParametersJob](https://github.com/brave-intl/publishers/blob/9aae7b40f8a4637c015d71d1b9412ea7739067d3/app/jobs/exchange_uphold_code_for_access_token_job.rb#L17) in `ExchangeUpholdCodeForAccessTokenJob#perform`.  When returned to the dashboard from uphold, the JavaScript will continuously check if the uphold connection is complete for ~30 seconds.  After failure it will display (old):
image

The proposed change would include a "Reconnect to Uphold" in this (new) view:
image

Resolves #439 

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/test/controllers/publishers_controller_test.rb
+++ b/test/controllers/publishers_controller_test.rb
@@ -616,6 +616,43 @@ class PublishersControllerTest < ActionDispatch::IntegrationTest
     assert_match(I18n.t('publishers.verification_uphold_state_token_does_not_match'), response.body)
   end
 
+  test "when uphold fails to return uphold_access_parameters, publisher has option to reconnect with uphold" do
+    publisher = publishers(:completed)
+
+    # sign in publisher
+    request_login_email(publisher: publisher)
+    url = publisher_url(publisher, token: publisher.reload.authentication_token)
+    get(url)
+
+    # give pub uphold state token
+    uphold_state_token = SecureRandom.hex(64)
+    publisher.uphold_state_token = uphold_state_token
+    expected_uphold_code = 'ebb18043eb2e106fccb9d13d82bec119d8cd016c'
+    publisher.save
+
+    # simulate return to homepage after creating wallet on uphold.com
+    # simulate failed response from uphold.com to get access params
+    stub_request(:post, "#{Rails.application.secrets[:uphold_api_uri]}/oauth2/token")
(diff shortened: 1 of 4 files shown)


test/controllers/publishers_controller_test.rb
```diff
@@ -616,6 +616,43 @@ def request_login_email(publisher:)
...
+    # verify uphold :code_acquired but not :access params
+    assert_equal publisher.reload.uphold_status, :code_acquired
+
+    # verify message tells publisher they need to reconnect
+    assert_select("div#publisher_status.uphold_processing") do |element|
+      assert_equal element.text, I18n.t("publishers.status_uphold_processing")
+    end
+
+    # verify button says 'reconnect to uphold' not 'create uphold wallet'
+    assert_select("div#uphold_connect div.panel-section a.btn.btn-primary") do |element|
```
mixonic: Might use a `data-test-` selector on this element instead of this long and likely-to-change selector.

nvonpentz: thanks for tip!

PR #456, opened 2018-01-10
Regress minitest to 5.10.3

### LINKED ISSUE #455: When tests fail minitest stops with an undefined method (for test name) error
Details are described in this issue https://github.com/seattlerb/minitest/issues/730

Propose reverting back to 5.10.3

lgebhardt:
Works around issue https://github.com/seattlerb/minitest/issues/730 by setting minitest to 5.10.3 in the Gemfile.lock. I don't like to edit Gemfile.lock, but this seems to be running correctly (and it's only for testing).

closes #455

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -110,6 +110,9 @@ group :test do
   # Clean state in-between tests which modify the DB
   gem "database_cleaner"
 
+  # Locking to 5.10.3 to workaround issue in 5.11.1 (https://github.com/seattlerb/minitest/issues/730)
+  gem 'minitest', '5.10.3'
+
   gem "webmock", "~> 3.0"
 end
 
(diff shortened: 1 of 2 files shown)


mixonic:
@lgebhardt You can pin a specific version in the `Gemfile` http://bundler.io/v1.5/gemfile.html#gemfiles and add a comment about the reason for a future traveller to untangle. I think that might be better than tweaking the lock alone.

mixonic:
Looks great, thanks :-)

zenspider:
This has been fixed in rails.

PR #457, opened 2018-01-10
Add 2FA setup prompt to end of signup

### LINKED ISSUE #411: 2FA trap/promo at sign up
In order to make 2FA discovery and enabling easy, insert this step just before the user lands on the dashboard (right after the opt-out screen).

- it takes the user directly to the 2FA home page
- allows user to skip enabling

2fa-trap

kpfefferle:
Resolves #411 

Redirects from dashboard only once to 2FA prompt, and not again once `two_factor_prompted_at` date is set. It does not prompt for Publishers who have already set up 2FA.

2FA Prompt

Submitter Checklist:

- [ ] Tagged reviewers.

Reviewer Checklist:

Tests

Security:


app/views/two_factor_registrations/prompt.html.slim
```diff
@@ -0,0 +1,14 @@
+.main-content
+  .container
+    .row
+      .sub-panel.login-panel.col-center.col-xs-12.col-sm-10.col-md-8.col-lg-6
+        .col.col-xs-12.col-sm-10.col-md-8.col-lg-10.col-center.text-center
```
mixonic: Seems odd to need both responsive `col` sizes here and on the parent `div`?

test/controllers/two_factor_registrations_controller_test.rb
```diff
@@ -0,0 +1,64 @@
+require "test_helper"
+
+class TwoFactorRegistrationsControllerTest < ActionDispatch::IntegrationTest
```
mixonic: Nice refactor to bring these together!

app/views/two_factor_registrations/prompt.html.slim
kpfefferle: IIRC this allows the inner content to be narrower than the `.sub-panel` on some bigger screens while filling the `.sub-panel` on smaller screens. I ripped this off from your classes in the `two_factor_authentications` templates here:

https://github.com/brave-intl/publishers/blob/a7ac955d75da1c167fbe362e7f19af5fe195d497/app/views/two_factor_authentications/index.html.slim#L3
https://github.com/brave-intl/publishers/blob/a7ac955d75da1c167fbe362e7f19af5fe195d497/app/views/two_factor_authentications/_totp.html.slim#L5

kpfefferle:
As discussed on 1/10, I will rebase this on top of #452 to avoid conflicts as well as avoid prompting Publishers who currently log in with Google OAuth

PR #458, opened 2018-01-10
Create rake task to check how many publishers have abandoned verification

### LINKED ISSUE #454: Check for Publishers who have paused during verification
Publisher verification for websites is a multi-step process, and there's a chance to fall off during the process. We should make an easy way to measure this. This data will help us decide if and when to reach out again to nudge them to resume (#431).

Consider publishers who might have initiated verification on an account A, then resumed it on a new account B (so A is still pending).

A rake task might be appropriate so we can run it in the future.

nvonpentz:
I added a new model method `Publisher#get_winback_publishers` that returns the publishers that have abandoned the verification process, but do not have another verified publisher account.  It looks at all the unverified publishers that have a non nil `brave_published_id` (that is only publishers who at least filled out the contact information page), and checks if there is a verified publisher with the same `email`, `brave_published_id`, or `phone_normalized`.  If there is such a verified publisher, the unverified publisher is considered a winback publisher and not included.

The rake task uses this the method and prints out the number of verified, unverified publishers, and unverified publishers that qualify for the winback email.

Resolves #454 

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/models/publisher.rb
```diff
@@ -239,6 +240,36 @@ def dont_destroy_verified_publishers
     throw :abort if verified?
   end
 
+  # Returns publishers who have abandoned verification process and do not have another publisher account
+  def self.get_winback_publishers
+    verified_publishers = Publisher.where(verified: true)
```
mixonic: At some point in the future loading all publishers into memory at once will not work. In the loop over `unverified_publishers` you could instead make a query to find a verified publisher by the unverified email.

app/models/publisher.rb
```diff
@@ -239,6 +240,36 @@ def dont_destroy_verified_publishers
...
+
+    # Only include publishers who have filled contact information
+    unverified_publishers = Publisher.where(verified: false).where.not(brave_publisher_id: [nil]).where("created_at < ?", WINBACK_THRESHOLD.ago)
+
+    winback_publishers = []
+    unverified_publishers.each do |publisher|
```
mixonic: Similarly `.each` here forces all records matching the query to be loaded into memory at once. Using `find_each` is probably preferable, as it will only load 1000 records into memory at once: http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each

nvonpentz:
Changes:
* Change method name from `get_winback_publishers` to `win_back_publishers`
* Use `find_each` instead of `each` for the loop

@mixonic Each of the checks within the loop execute a single query with a limit of one publisher, so we should be ok with overloading memory after this change.

app/models/publisher.rb
```diff
@@ -239,6 +240,36 @@ def dont_destroy_verified_publishers
     throw :abort if verified?
   end
 
+  # Returns publishers who have abandoned verification process and do not have another publisher account
+  def self.win_back_publishers
```
ayumi: Could we move this logic into a service like `PublishersUnverifiedCalculator`?

app/models/publisher.rb
```diff
@@ -239,6 +240,36 @@ def dont_destroy_verified_publishers
...
+    verified_publishers = Publisher.where(verified: true)
+
+    # Only include publishers who have filled contact information
+    unverified_publishers = Publisher.where(verified: false).where.not(brave_publisher_id: [nil]).where("created_at < ?", WIN_BACK_THRESHOLD.ago)
+
+    win_back_publihsers = []
```
ayumi: typo here and throughout

test/services/publisher_unverified_calculator_test.rb
```diff
@@ -0,0 +1,63 @@
+require 'test_helper'
+
+class PublisherUnverifiedCalculatorTest < ActiveJob::TestCase
+  test "only returns unverified publishers who don't share an email, brave_publisher_id, or phone number with a verified publisher" do
+    # clear database
+    Publisher.delete_all
```
ayumi: i think this is breaking the tests, because the fixtures are deleted and not recreated, impacting subsequent tests

can we comment this out for now?

nvonpentz: good catch.  i'll adjusts the tests so fixtures do not interfere.

nvonpentz:
I think it was the time traveling that was upsetting the other tests, I put them in their own block to address this.  For good measure I removed the call to delete the fixtures, and changed the names on the publishers so there would be not conflicts within this test.

ayumi:
thanks!

PR #473, opened 2018-01-11
Allow channel verification to be resumed

### LINKED ISSUE #413: Multi-channel support: allow user to resume verification process
Once the verification key has been handed out to the user at the 3rd step, we keep the channel record along with the key and expose the incomplete channel on the dashboard for the user to re-enter the process.

this is because:
1. the session can time out while the user is completing the out-of-bound steps
2. we need a mechanism of revoking the key (though not guaranteed), so that they can clean restart. 

*channel record should be discarded if the process is abandoned prior to the key distribution: i.e. first 2 steps as they are very simple to restore.
**warning screen when the user removes incomplete channel (ask them to delete the key file)

dashboard-incompletech

dgeb:
When a channel has not yet been verified, it is shown on the dashboard with a subtly transparent background, as well as the message “One more step to complete…”. A button labeled “Let’s Finish” allows users to resume verification.

Also consolidates channel-specific elements + styles for the dashboard.

Closes #413 

Note: the marketing opt-in toggles will be moved from the channel to the publisher level in a subsequent PR.

screen shot 2018-01-11 at 2 11 50 pm

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


lgebhardt:
@jenn-rhim wanted abandoned site channels to not show up. Abandoned is defined as we haven't shown the user the verification token.

ayumi:
@lgebhardt okay, that makes sense. i'll open an issue to GC abandoned channels

Issue #465, opened 2018-01-10
Bootstrap 4


mixonic:
We will migrate the site off Bootstrap 3 and onto Bootstrap 4 (beta.3, currently). Bootstrap 4 has better customization options for its typographic and layout grids.

Our workflow will be to first migrate the main dashboard page. This will be PR'd and can be landed in staging. Then we will incrementally convert the rest of the dashboard pages. The theory is that having completed the most complex page first, other subsequent pages will be simple.

This does mean that staging, and potentially production, will have a period where pages may be slightly inconsistent. For example the navigation in the top bar will be implemented in v4 on some pages and in v3 on others making it slightly different. This would be temporary however.

This effort will additionally:

* Change bootstrap 4 to be consumed via npm instead of via a Ruby gem.
* Introduce new breakpoints at more reasonable and modern sizes.
* Add mobile responsiveness to pages as we migrate them.
* Introduce a new typographic and layout grid.

This effort will *not*:

* Change the dashboard background image. This would be too jarring going between 3 and 4 during the transition.
* Change button colors.

ayumi:
sgtm!

just a note to consider the eventual merge resolution between this and `staging2` / #452 Multi-chan

ayumi:
closing as the foundations of this work have been implemented with #490 

further steps to transition off of bootstrap 3 will be tracked with https://github.com/brave-intl/publishers/labels/epic%20%2F%20Style%20refactor

Issue #464, opened 2018-01-10
Update font selection


mixonic:
According to recent mocks, the dashboard should be using [Poppins](https://fonts.google.com/specimen/Poppins) for headlines and [Muli](https://fonts.google.com/specimen/Muli) for some body text.

Currently the site uses Open Sans. We use 4 variants, and load each font individually from our own servers (not a CDN).

<img width="990" alt="screen shot 2018-01-10 at 1 11 56 pm" src="https://user-images.githubusercontent.com/8752/34795657-e004508e-f607-11e7-99de-bece662d8bbf.png">

We would like to adjust this to improve a) the font for the dashboard and b) the performance of the website.

With those goals in mind, we suggest:

* We use Poppins with weights of 400. This will permit us to use it for headlines.
* We make the "Brave Payments" logo in the header (currently "Open Sans") into an SVG rendered with Poppins 300. This avoids us loading a whole variant (Poppins 300) just for the logo.
* We use Muli 400 and 600 for regular and bold text across the site, as the newer mocks suggest. There are a few places in the mocks where Poppins 300 is used for a label instead of a headline. In those places we will use Muli instead.
* We will load these fonts via Google Fonts instead of as assets from the publisher dashboard domain.
* For the first pass we will *not* change any fonts on the marketing page besides the logo "Brave Payments" font.

These plans are 100% open to feedback.

ayumi:
Can we load fonts from our own servers? It would make me feel marginally better privacy wise.

everything else sounds good to me (cc @jenn-rhim )

mixonic:
@ayumi We should try to shrink the payload regardless of where we host. It would be really, really nice to get these assets behind HTTP2 and/or on a CDN. We can avoid using Google Fonts for the CDN, it simply means a few more hurdles to jump.

ayumi:
@mixonic good thought. Fastly could work as we use it on other Brave properties, as could Cloudfront.

https://publishers.basicattentiontoken.org currently loads 4 fonts, and uncached on a good connection it takes 600 to 1000 ms

using a cdn for fonts might be more of a short/medium term thing (in +1--2 weeks)

kpfefferle:
@jenn-rhim We noticed today that the sketch file we have been using for reference has two different variations of the header nav "Brave Payments" word mark.

One variation uses Open Sans with color `#696B6D`:
[...]

Issue #123, opened 2017-09-25
Button styles should be only in the shared button CSS file


ayumi:
https://github.com/brave-intl/publishers/pull/121#pullrequestreview-65052986

ayumi:
superseded by style refactor project https://github.com/brave-intl/publishers/labels/epic%20%2F%20Style%20refactor

we can check back after

PR #485, opened 2018-01-16
Split landing page CSS into separate manifest

kpfefferle:
This is a preliminary step to attacking #464 and #465.

- Isolates the landing page "static" CSS so that we can begin migration of the rest of the application to new fonts and Bootstrap 4 without interference
- Reduces the CSS payload size for the initial landing page
- Only imports the portions of Bootstrap needed to render the landing page

Before (528k `application.css`):
Landing Page Before

After (140k `static.css`):
Landing Page After


mixonic:
Note that this PR is against `staging2`.

I've confirmed tests pass for this branch locally.

screen shot 2018-01-16 at 3 16 27 pm

app/assets/stylesheets/static/static.sass
```diff
@@ -0,0 +1,226 @@
...
+
+@media (min-width: $screen-lg-min)
+  .col-left
+    padding-left: 80px
+    padding-right: 20px
+  .col-right
+    padding-left: 20px
+    padding-right: 80px
+
+.form-control
```
ayumi: minor: in the future we may want to extract buttons so we can reuse them #123

PR #489, opened 2018-01-17
Referral promo

nvonpentz:
### Additions
**Launch Promo Rake Task**
* Generates a `promo_token_2018q1` for each publisher via the `PublisherPromoTokenGenerator`
* Sends `PromoMailer#activate_promo_2018q1` upon successful registration

**PublisherPromoTokenGenerator**
* Generates a promo token to be used to access promo pages / activate the promo for the publisher (only for existing publishers at the time of promo launch)

**PromoRegistrationsController**
* `create` sets `Publisher#promo_enabled_2018q1` to `true`, engages the `PromoRegistrar`, which sets creates a `promo_registration` for the channel and sets `promo_registration.referral_code` to the code returned by the promo server.
* `index` displays various screens depending on whether the promo is running, the publisher has enabled the promo, and whether the publisher has any verified channels
* Publishers can access their promo page without authorization with a `promo_token_2018q1` in the params if no active session is found

**PromoRegistrar**
* Takes a publisher and registers every verified with the promo server

**PromoRegistrationGetter**
* Used to by the `PromoRegistrar` to get an existing in case the promo api tells us that there is a duplicate error

**PromoMailer**
* `activate_promo_2018q1` - sent once by the launch_promo rake task.  Owner accounts created after the launch of the promo will not get this email
* `promo_enabled_2018q1_verified` - sent once a publisher has activated the promo and has verified channels
* `promo_enabled_2018q1_unverified` - sent once a publisher with no verified channels activates the promo
* `new_channel_registered_2018q1` - sent when a publisher verifies a new channel while the promo is enabled

**Channel Model additions**
* `after_save :register_channel_for_promo, if: :should_register_channel_for_promo` - whenever a new site or youtube channel is verified, it launches `register_channel_for_promo` which uses the `RegisterChannelForPromoJob`

**RegisterChannelForPromoJob**
* engages the `PromoRegistrar`, and upon successful registration, emails `PromoMailer#new_channel_registered_2018q1` to the publisher

### TO DO
* Integrate statistics into the dashboard

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/views/publishers/promo_activate.html.slim
```diff
@@ -0,0 +1,14 @@
+- content_for(:navbar_content) do
```
mixonic: These `content_for(:navbar_content)` blocks in all these files have no content. Can they simply be removed?

mixonic:
Also there is a `.DS_Store` file here that should be removed

app/views/publishers/promo_activate.html.slim
nvonpentz: Yes they can; I've removed them.  Thanks for the comment.

app/controllers/referral_promos_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      else
+        render(:activate)
+      end
+    else
+      render(:over)
+    end
+  end
+
+  def create
+    promo_running = active_promo_id.present?
```
ayumi: minor: i think a more railsy way to do this is to have a `before_action` filter which checks for stuff and redirects in case the condition fails.

config/routes.rb
```diff
@@ -34,6 +34,11 @@
       resources :u2f_authentications, only: %i(create)
       resources :totp_registrations, only: %i(new create destroy)
       resources :totp_authentications, only: %i(create)
+      resources :referral_promos, only: %i(create) do
```
ayumi: i think adding the `index` default action might do what `get :show` on collection is doing

`only: %i(create index)`

app/controllers/referral_promos_controller.rb
```diff
@@ -0,0 +1,43 @@
+class ReferralPromosController < ApplicationController
+  before_action :authenticate_publisher!
+
+  def show
```
ayumi: should it be `index`? since we plan to view all active promo codes on one page.

app/controllers/referral_promos_controller.rb
```diff
@@ -23,8 +23,17 @@ def create
       @publisher.promo_enabled_2018q1 = true
       @publisher.save!
 
-      # TO DO: register channel with promo service
-      # TO DO: create new referral promo from response
+      channels = @publisher.channels.where(verified: true)
```
ayumi: TODO: when number of channels is >5 generate codes asynchronously

app/controllers/referral_promos_controller.rb
```diff
@@ -23,8 +23,17 @@ def create
...
+
+      channels.each do |channel|
+        # TO DO: Consider a clever query instead of checking .blank?
+        if channel.referral_promo.blank?
```
ayumi: i'd check also for presence of referral_promo.referral_code

ayumi: consider moving this whole block here (the check if a referral promo needs to be created; and creation of a referral promo) into ChannelPromoRegistrar since all the logic happens together always.

ChannelPromoRegistrar could raise an error if a Channel already has a ReferralPromo or it's is not verified.

app/services/channel_promo_registrar.rb
```diff
@@ -0,0 +1,42 @@
+# Register channel for promo
+class ChannelPromoRegistrar < BaseApiClient
+  def initialize(channel:)
```
ayumi: minor: add a TODO note that promo ID is hardcoded, and we should make it a service param later if needed.

app/controllers/referral_promos_controller.rb
nvonpentz: Yes `index` feels better suited, I'll change the actions/routes

app/controllers/referral_promos_controller.rb
nvonpentz: Thanks, I considered this but wasn't sure where was most appropriate

app/controllers/referral_promos_controller.rb
nvonpentz: I thought about this too.

So apply a `before_action` to `create` that redirects to `index` if the promo is not running, and have `index` render `over` as it does already?

ayumi: i think we should render the `index` view always, since eventually that page will probably show their participation in past promos as well as opt into future promos.

for now, if the promo is over we can show a section which is blank except for the message.

app/models/referral_promo.rb
```diff
@@ -0,0 +1,12 @@
+class ReferralPromo < ApplicationRecord
```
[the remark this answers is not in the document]
nvonpentz: Maybe `ReferralPromoRegistration` and `ReferralPromoRegistrationsController`? We could drop `Referral` and/or prefix with `Channel` to be verbose.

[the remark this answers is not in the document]
ayumi: I like `PromoRegistration` !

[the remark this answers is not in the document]
nvonpentz: sounds good!

app/controllers/promo_registrations_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    Rails.application.secrets[:active_promo_id].present?
+  end
+
+  def publisher_activated_promo
+    current_publisher.promo_enabled_2018q1
+  end
+
+  def require_active_promo
+    return if promo_running
+    redirect_to index
```
ayumi: related to above `render :index` bits -- should this be `redirect_to promo_registrations_path` ?

app/services/publisher_promo_token2018q1_generator.rb
```diff
@@ -0,0 +1,15 @@
+# Generates a token for a publisher for 2018q1 promotion
+class PublisherPromoToken2018q1Generator < BaseService
```
ayumi: you can just call this PublisherPromoTokenGenerator

then have it take a param in `#initialize` for `promo_id:` and blow up if it's not the hardcoded code (free-bats-2018q1)

app/mailers/promo_mailer.rb
```diff
@@ -0,0 +1,19 @@
...
+    @private_promo_2018q1_auth_url = generate_private_promo_2018q1_auth_url(publisher)
+
+    mail(
+      to: @publisher.email,
+      subject: default_i18n_subject
+    )
+  end
+
+  # Maybe move this to publisher helper?
+  def generate_private_promo_2018q1_auth_url(publisher)
```
ayumi: agree, generally mailers should be pretty devoid of state changes
I think we should generate the promo activate tokens in a rake task.

to get the url, you could create a PromoHelper and make a `publisher_promo_auth_url` which takes a publisher and promo id

config/locales/en.yml
```diff
@@ -261,6 +261,12 @@ en:
       title: "Sorry! The promotion has ended."
       description: "We're sorry that you missed on this program.  But you can still influence your fans to support you by using Brave!"
     back_to_dashboard: "Back to Dashboard"
+  promo_mailer:
+    activate_promo_2018q1:
+      title: "Limited Time Offer: Free tokens for your referral!"
+      subject: "Limited Time Offer: Free tokens for your referral!"
+      body: "Getting more fans on Brave means more income for oyu. And we'd like to chip in for that effort. We are giving away bonus tokens for refrring your fans to Brave. Find out more and activate this promo!"
```
ayumi: minor: oyu

config/locales/en.yml
```diff
@@ -261,6 +261,12 @@ en:
       title: "Sorry! The promotion has ended."
       description: "We're sorry that you missed on this program.  But you can still influence your fans to support you by using Brave!"
     back_to_dashboard: "Back to Dashboard"
+  promo_mailer:
+    activate_promo_2018q1:
+      title: "Limited Time Offer: Free tokens for your referral!"
```
ayumi: minor: copy changes maybe? can track that separately

app/jobs/launch_promo2018q1_job.rb
```diff
@@ -0,0 +1,23 @@
+# Generates the promo tokens and sends emails
+class LaunchPromo2018q1Job < ApplicationJob
```
ayumi: rake task would be good i think.
it'd generate the tokens and queue up all the emails

app/services/publisher_promo_token2018q1_generator.rb
```diff
@@ -0,0 +1,15 @@
...
+  attr_reader :publisher
+  
+  def initialize(publisher:)
+    @publisher = publisher
+  end
+
+  def perform
+    return if Rails.application.secrets[:active_promo_id].blank?
```
ayumi: make sure we don't accidentally overwrite existing promo tokens

consider making this service omnipotent by adding a check for publisher.promo_token_2018q1
if it's present, return nil early

you could add a `force:` param to allow manual overwriting

app/services/promo_registrar.rb
```diff
@@ -12,44 +15,47 @@ def perform
     channels.each do |channel|
       if should_register_channel?(channel)
         referral_code = register_channel(channel)
-        promo_registration = PromoRegistration.new(channel_id: channel.id, promo_id: active_promo_id, referral_code: referral_code)
+        promo_registration = PromoRegistration.new(channel_id: channel.id, promo_id: @promo_id, referral_code: referral_code)
```
ayumi: minor: to increase code readability, consider checking for referral code and short circuiting:
```ruby
  next if referral_code.nil?
```

this also protects from the `#save!` exception case below

app/services/promo_registrar.rb
```diff
@@ -12,44 +15,47 @@ def perform
...
         promo_registration.save!
```
ayumi: when referral_code is nil i believe this will raise and stop code generation for subsequent channels

lib/tasks/launch_promo.rake
```diff
@@ -0,0 +1,28 @@
+require "#{Rails.root}/app/helpers/promos_helper"
+include PromosHelper
```
ayumi: I have a feeling that `include` in the top level namespace might be a bad idea

i think moving the include to the task body might be okay?

app/services/publisher_promo_token_generator.rb
```diff
@@ -0,0 +1,36 @@
...
+      else
+        raise InvalidPromoIdError.new("#{@promo_id} is an invalid promo id") # Rescued and reported in launch promo rake task
+    end
+  end
+
+  def perform_2018q1
+    already_has_promo_token = publisher.promo_token_2018q1.present?
+
+    if already_has_promo_token && !@force
+      Rails.logger.info("Publisher #{@publisher} already has a promo token, use force=true to overwrite.")
```
ayumi: maybe return `nil` here, as Rails.logger.info() returns `true`. this would help in checking if the service generated a code or not.

test/tasks/launch_promo_test.rb
```diff
@@ -0,0 +1,30 @@
...
+    # end
+  end
+
+  test "generates a promo token for each publisher" do
+    assert_difference("Publisher.where.not(promo_token_2018q1: nil).count", Publisher.count) do
+      assert_difference("ActionMailer::Base.deliveries.count" , Publisher.count) do
+        Rake::Task["promo:launch_promo"].invoke
+      end
+    end
+  end
```
ayumi: I would add a test to verify idempotence; meaning

1. add publisher A
2. run rake task once
3. add publisher B
4. run rake task again
5. verify A only got 1 email

lib/tasks/launch_promo.rake
```diff
@@ -0,0 +1,28 @@
...
+  # Generates the promo tokens and sends emails
+  task :launch_promo => :environment do
+    begin
+      unless promo_running?
+        puts "Promo is not running, check the active_promo_id config var."
+      end
+
+      publishers = Publisher.all
+
+      publishers.find_each do |publisher|
```
ayumi: consider grouping this with the below block within `#find_each`:
```ruby
publishers.find_each do |publisher|
  token = PublisherPromoTokenGenerator.new(publisher: publisher)
  next unless token
  PromoMailer.activate_promo_2018q1(publisher).deliver_later
end
```

lib/tasks/launch_promo.rake
```diff
@@ -0,0 +1,28 @@
...
+        PublisherPromoTokenGenerator.new(publisher: publisher)      
+      end
+
+      publishers.find_each do |publisher|
+        PromoMailer.activate_promo_2018q1(publisher).deliver
```
ayumi: this will send an email to all publishers. if we encounter an error in the middle of the task, it would difficult to resume without spamming the publishers. see above

app/helpers/promos_helper.rb
```diff
@@ -0,0 +1,23 @@
+module PromosHelper
+  def generate_private_promo_auth_url(publisher)
+    promo_token = PublisherPromoTokenGenerator.new(publisher: publisher).perform
```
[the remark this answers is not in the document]
nvonpentz: Where should this method go?  It looks like we have a similar method for normal auth tokens in the PublishersHelper https://github.com/brave-intl/publishers/blob/37a0cb108f1d4d2f8089004859d2a893f050251b/app/helpers/publishers_helper.rb#L214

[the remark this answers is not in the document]
ayumi: indeed, that's also suboptimal

i think we're only generating this promo auth token infrequently, in the rake task.
so maybe you can move the PublisherPromoTokenGenerator invocation there.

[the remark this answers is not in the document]
ayumi: currently we're using this helper in the Mailer.
maybe we'd change it to just read the `promo_token` from Publisher.

app/mailers/application_mailer.rb
```diff
@@ -13,9 +13,14 @@ def self.should_send_internal_emails?
 
   private
 
-  def add_image(image_file_name)
-    path = Rails.root.join("app/assets/images/mailer/#{image_file_name}")
-    attachments.inline[image_file_name] = File.read(path)
+  def add_image(image_file_name, source = "mailer")
```
ayumi: i would actually prefer we remove the `source` param and be more explicit about `image_file_name`, maybe renaming that param to `image_path`.

then in this method body:
`    path = Rails.root.join("app/assets/images/#{image_path}")`

and change the previous usages of `add_image` such as `add_image("logo-reddit.png") to `add_image("mailer/logo-reddit.png")`

the goal is to minimize logic and help readers of the code.

app/views/promo_mailer/new_channel_registered_2018q1.html.slim
```diff
@@ -0,0 +1,18 @@
...
+  h3= t("promo_mailer.new_channel_registered_2018q1.subject")
+
+p.salutation= t("publisher_mailer.shared.salutation", name: @publisher.name)
+
+p = t("promo_mailer.promo_activated_2018q1_verified.body_one")
+
+.promo--item-label = "#{@channel.publication_title}"
+.promo--referral-link-container
+  span.promo--referral-link = referral_link(@channel.promo_registration.referral_code)
+  = link_to("", tweet_url(@channel.promo_registration.referral_code), style: "background-image: url(#{attachments["tweet.png"]&.url})", class: 'promo--share-button')
```
ayumi: - does it work in Gmail?

- `&.url` means these images could fail silently.
if we're to use that, it's important to create a test to confirm the image shows up.

app/views/publishers/home.html.slim
```diff
@@ -393,8 +434,57 @@ noscript
...
           = form_for(channel, html: {id: "remove_channel_#{channel.id}"}) do |f|
         .clearfix
-
 .row.tos-row
   .col.col-details.col-md-6.col-md-push-6.col-xs-10.col-xs-center.col-sm-center
     .terms-of-service
       = link_to(t("shared.terms_of_service"), terms_of_service_url)
+
+- if current_publisher.promo_status(promo_running?) == :active
+  javascript:
```
ayumi: @mixonic where's the best place to put referral promo specific javascript?

app/models/channel.rb
```diff
@@ -50,4 +53,37 @@ def details_not_changed?
...
+    when "YoutubeChannelDetails"
+      return self.details.youtube_channel_id
+    when "SiteChannelDetails"
+      return self.details.brave_publisher_id
+    else
+      nil
+    end
+  end
+
+  def promo_enabled?
```
ayumi: I think we should create a model Concern for model ref promo methods
https://stackoverflow.com/questions/14541823/how-to-use-concerns-in-rails-4

app/models/publisher.rb
```diff
@@ -159,6 +164,36 @@ def owner_identifier
     "publishers#uuid:#{id}"
   end
 
+  def promo_status(promo_running)
```
ayumi: i think we should put model promo methods in a model Concern (this one would be a Concern just for Publisher). for cleanliness

https://stackoverflow.com/questions/14541823/how-to-use-concerns-in-rails-4

app/models/promo_registration.rb
```diff
@@ -0,0 +1,12 @@
+class PromoRegistration < ApplicationRecord
+  has_paper_trail
+  
+  belongs_to :channel, validate: true, autosave: true
+  # belongs_to :publisher, polymorphic: true, validate: true, autosave: true
```
ayumi: minor: a commented line

app/models/publisher.rb
```diff
@@ -40,9 +41,13 @@ class Publisher < ApplicationRecord
   # (see `verify_uphold` method below)
   validates :uphold_access_parameters, absence: true, if: -> { uphold_verified? }
 
+  validates :promo_token_2018q1, uniqueness: true,  if: -> { self.promo_token_2018q1 != nil }
```
ayumi: you can add a :allow_nil condition here:

```ruby
validates :promo_token_2018q1, uniqueness: true,  allow_nil: true
```

app/models/publisher.rb
```diff
@@ -159,6 +164,36 @@ def owner_identifier
...
+      :over
+    elsif self.promo_enabled_2018q1
+      :active
+    else
+      :inactive
+    end
+  end
+
+  def promo_stats_status
+    promo_disabled = !self.promo_enabled_2018q1
```
ayumi: minor: `self` is redundant

app/mailers/application_mailer.rb
```diff
@@ -13,9 +13,14 @@ def self.should_send_internal_emails?
...
+    if source == "mailer"
+      path = Rails.root.join("app/assets/images/mailer/#{image_file_name}")
+      attachments.inline[image_file_name] = File.read(path)
+    elsif source == "assets"
+      path = Rails.root.join("app/assets/images/#{image_file_name}")
+      attachments.inline[image_file_name] = File.read(path)
+    end
```
ayumi: (minor: if we decided on using the method as is, probably there would be an `else` case which raised an error)

app/services/promo_registrar.rb
```diff
@@ -0,0 +1,113 @@
...
+    {
+      "owner_id": @publisher.id,
+      "promo": @promo_id,
+      "channel": channel.channel_id, 
+      "title": channel.publication_title,
+      "channel_type": "website",
+    }.to_json
+  end
+
+  def make_empty_string_nil(param)
```
ayumi: this is implemented in rails by:
```ruby
"".presence
# => nil
```

app/views/layouts/mailer.html.slim
```diff
@@ -5,9 +5,14 @@ html
     = stylesheet_link_tag('mailer', media: 'all')
   body
     .main
-      .header style="background-image: url(#{attachments["header-pattern.png"]&.url})"
-        .brand
-          = image_tag(attachments["brave-payments.png"]&.url, alt: t("shared.app_title"), height: 31, width: 268)
+      - if @_controller.is_a?(PromoMailer)
```
ayumi: this is kind of hacky

i think we should have a separate PromoMailer layout

or (i'm not sure how to do this): a layout which inherits from `mailer.html.slim`

app/views/publishers/home.html.slim
```diff
@@ -393,8 +434,57 @@ noscript
...
+        setTimeout(function(){
+          if (channelPromoInfoContainer.classList.contains('channel--promo-info-container-slideout')) {
+            channelPromoInfoContainer.classList.remove('channel--promo-info-container-slideout');
+          }
+        }, 100);
+      }
+
+      function openReferralLinkAnimation() {
+        var channelPromoInfoContainer = this.parentNode;
+        console.log("close referral animation");
```
ayumi: pls remove comments before shipping

app/helpers/promos_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+    referral_code
+  end
+
+  def offline_promo_stats
+    {"times"=>[Time.now.to_s], "series"=>{"name"=>"downloads", "values"=>[rand(0..1000)]}}
+  end
+
+  def total_possible_referrals(publisher)
+    stats = publisher.promo_stats_2018q1
+    if publisher.promo_stats_2018q1 == "{}"
```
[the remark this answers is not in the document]
nvonpentz: promo_stats_2018q1 is stored as a JSONB datatype and it defaults to `'{}'` https://github.com/nvonpentz/publishers/blob/8b4c8bd2c3e521db0a62e1539be762652872f275/db/migrate/20180131155047_add_columns_to_publisher.rb#L3.

Would you rather it default to `nil` or something else and instead check for that?

app/views/layouts/mailer.html.slim
nvonpentz: i agree, and will look into this but at a lower priority for now

app/views/promo_mailer/new_channel_registered_2018q1.html.slim
nvonpentz: This does not consistently work in gmail.  Another thing I noticed is that gmail displays these images only as attachments. ex:

image

I do not have a plan to resolve this yet, any input would be much appreciated as this is hard to debug since there is no problem locally.

app/views/layouts/mailer.html.slim
nvonpentz: Now using a PromoMailer layout

app/helpers/promos_helper.rb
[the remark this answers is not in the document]
ayumi: i believe you can default to `{}` (empty Hash) which is preferable
https://stackoverflow.com/a/38006327

app/views/promo_mailer/new_channel_registered_2018q1.html.slim
ayumi: a majority (source: my previous job) of people use gmail so it's necessary if we want to ship the button

maybe try css `background` shorthand? some hints here (could be hearsay)
https://litmus.com/community/discussions/5645-background-images-breaking-in-gmail#comment-8973

PR #490, opened 2018-01-17
Bootstrap 4

mixonic:
Here we begin the migration to Bootstrap 4 and some design tweaks (like fonts), but we're trying to do it as iteratively as possible.

We've forked the [bootstrap-sass](https://github.com/twbs/bootstrap-sass) gem, which currently provides Bootstrap 3, to a new gem called [bootstrap-3-sass](https://github.com/kpfefferle/bootstrap-3-sass). The only real difference is that all SCSS files in the new gem are namespaces behind the `bootstrap-3` prefix instead of `bootstrap`. Doing this frees the `bootstrap` namespace for the new Bootstrap 4 code.

Additionally we've added a dependency on the [bootstrap](https://github.com/twbs/bootstrap-rubygem) gem. This gem itself provides Bootstrap 4 as `bootstrap` to SCSS.

Thus both Bootstrap 3 and Bootstrap 4 are available in the same codebase. Of course having a single CSS file with both Bootstrap versions would be nonsense, thus the current `application.scss` has been moved to `application-bootstrap-3.scss` and new `application.scss` has been introduced.

The migration process will basically be to port each page over to the new styles. We have two development  tools to help this process:

* Though by default the `application-bootstrap-3.scss` file is loaded, you can opt-in to the new `application.scss` by adding the query param `bootstrap4` to a URL. This is only semi-useful, of course the HTML is unchanged.
* As we port a page to the new styles, we can opt that page into the new file by providing a `content_for(:head_style_tags)` block in the updated template that points at the `application.scss`. Eventually, we would expect all pages to use a `content_for` block opting in and we can simply change the default.

TODO:

* [x] Should all legacy CSS go into a single folder? I think so, then new stuff can be at the top level or in another directory.
* [x] ~Let's port a real page like `publishers/home.html.slim` over to the new styles and see how well this system works before merging this PR.~ We migrate the login and login-email-sent pages. This seems like the smallest single unit of user experience we could migrate.

TODO after the conversion is complete, not blocking merge:

* Drop `pull-left` `pull-right` in favor of better layout tooling (flexbox) or `float-left` `float-right` at the least.


mixonic:
FYI I've rebased and force pushed this branch

mixonic:
FYI I've rebased and force pushed this branch

mixonic:
**Login, before**

screen shot 2018-01-22 at 1 42 36 pm

**Login, after**

screen shot 2018-01-22 at 1 42 43 pm

image 2018-01-22 13-46-02

**Login when active input, after**

screen shot 2018-01-22 at 1 48 54 pm

**Login at small, before**

screen shot 2018-01-22 at 1 43 05 pm

**Login at small, after**

screen shot 2018-01-22 at 1 43 10 pm

**Login at large, before**

screen shot 2018-01-22 at 1 43 18 pm

**Login at large, after**

screen shot 2018-01-22 at 1 43 25 pm

mixonic:
@kpfefferle and I think this is ready for merge in its current state. We can continue to iterate on other pages once this lands.

jenn-rhim:
@mixonic @kpfefferle  looks good to me!

app/views/publishers/new_auth_token.html.slim
```diff
@@ -1,25 +1,26 @@
...
-        .col.col-xs-12.col-sm-8.col-center
-          = f.email_field(:email, autofocus: true, class: "form-control", placeholder: t("publishers.enter_your_email"))
-          - if params[:captcha]
-            = hidden_field_tag(:captcha)
-          - if @should_throttle
-            .form-group
-              = recaptcha_tags
-      .row.form-row
-        .col.col-xs-12.col-sm-8.col-center
+    .row
```
ayumi: comment: how exciting that we can clean up all the CSS classes!

mixonic:
@ayumi identified that the wordmark SVG is rendering incorrectly on a Linux VM with Chromium 63.0.3239.84 (Developer Build) built on (64-bit). Also reports the same thing in Brave.

mixonic:
The above rendering inconsistency was due to the wordmark being a text element instead of an SVG path. This  has been corrected.

PR #491, opened 2018-01-17
[Multichannel] Move `visible` flag from channels to publishers

### LINKED ISSUE #403: Remove Opt-in step from verification: applies to all pubs
**Background:**  the current opt-in has a problem of trying to get combined agreements from the users on 1) being listed on our Brave Verified Publishers list and on 2) being ok to be used in our marketing effort (effectively co-marketing) in one button.  3) We also need to get TOS agreement, which has been missing.

The result of the discussion came down to the following decisions:
1.  By signing up Brave Payments, the publisher is automatically in our Brave Verified Publishers list.
2. Channels they verify will automatically get the 'Verified' badge that will be shown in the browser Payments table.
--> '# 1 and # 2' mean that we don't need the user's opt-in.

3. We do need the user's consent on their logos/names potentially being used on our properties (websites, social media, and any promos/campaigns). --> this applies to the publisher level, not individual channel level. 

## This means **the changes in Pub sign-up & verification flow**.
**1. During the sign-up**: when the user made the round trip from the email link, we present them one more screen to complete registration.  The screen will ask their name, phone # (optional), and the opt-in checkboxes. This will be the full screen mode, not pop-up.

**2. We no longer need the opt-in step** at the end of channel verification (for any channel), which simplifies the flow.

(screens to come soon: legal input/review required)

dgeb:
This flag controls whether channels can be listed as Brave Verified Publishers. Instead of setting this flag on individual channels, it should now be set once per publisher.

A checkbox has been added to both the complete registration step and the dashboard.

Closes #403

/cc @ayumi @jenn-rhim 

The updated "complete registration" panel:

screen shot 2018-01-17 at 12 25 16 pm

The updated dashboard (note the new "Verified" label on the channel - this is just a first attempt):

screen shot 2018-01-17 at 12 24 25 pm

Submitter Checklist:

- [X] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/assets/stylesheets/application/publishers.sass
+++ b/app/assets/stylesheets/application/publishers.sass
@@ -282,12 +282,24 @@ body[data-controller="publishers"]
   &[data-action="email_verified"],
   &[data-action="complete_signup"]
     .process-panel
-      padding: 80px 200px 80px 100px
+      padding: 80px 200px 80px 80px
       background: #fff url(asset-path("img-brands.png")) top right no-repeat
+      .content-panel form
+        width: 350px
+        .form-check-input
+          vertical-align: top
+          margin-top: 5px
+        .form-check-label
+          padding-left: 15px
+          width: 330px
+          font-size: 14px
+          font-weight: 300
+        .panel-controls
+          margin-top: 45px
       h1
         color: #ff3f3f
       .tos
-        margin-bottom: 20px
+        margin-top: 20px
         font-size: 14px
         color: $gray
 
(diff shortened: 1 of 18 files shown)


jenn-rhim:
@ayumi i know that you're on top of it but just wanted to be sure, given the wording used to describe this flag above... this is not the 'Verified' flag, but only sets the marketing opt-in as a global flag per account.  Verified checkmark should be handled per channel.

PR #494, opened 2018-01-18
Introduce publishers:transform_legacy_data rake task

### LINKED ISSUE #477: Migrate Publishers and YoutubeChannels to Owners, Channels and ChannelDetails
Currently we have Publishers and YoutubeChannels, and we'd like to migrate them to the new model: Owners with multiple Channels.

Publishers -> Owner + (Channel / Site + details)
YoutubeChannels -> Channel / Youtube + details

In multi-chan, email is now a unique key on Owner.
Previously a logical person who owned many websites might many Publisher records, each with the same email.
In that case we should merge all Publisher records sharing the same email, creating a single Owner record and assigning itall the corresponding Site and Youtube channels.

This would probably be a rake task. It would be good to be tested, and a bonus if we had Dry Run mode. We should also take a backup of the database before migrating.

Related to #273 and #452

lgebhardt:
This rake task transforms the legacy publishers data to owners and channels. Legacy publishers which share an email address are consolidated to a single Publisher (Owner) record.

Each legacy publisher is also migrated to a channel associated with the common Owner created above. Channel type is either site or youtube.

* Sessions are being dropped.

* U2F registrations from all channels owned by the same owner are recreated with the owner.

* Existing Totp registrations are moved over only if there is a single one across the owner space.

An Owner's `visible` status is set to true if all the rolled up legacy publishers `show_verification_status` flags were true.

The rake task runs in a dry run mode unless the `commit` flag is provided:
```bash
bin/rake publishers:transform_legacy_data

Performing a trial run. Changes will be rolled back
Going to transform 46 email verified out of 155 legacy publishers
------------------------------------
Importing publisher 0 Id: 01b64ade-b5d1-4f75-8916-686da52c8a00 Name: 
...
------------------------------------
16 Owners created
10 verified Site Channels created
5 unverified Site Channels created
13 Youtube Channels created
------------------------------------
Trial Run: Rolling Back

```

```bash
bin/rake publishers:transform_legacy_data[commit]

Going to transform 46 email verified out of 155 legacy publishers
------------------------------------
Importing publisher 0 Id: 01b64ade-b5d1-4f75-8916-686da52c8a00 Name: 
...
------------------------------------
16 Owners created
10 verified Site Channels created
5 unverified Site Channels created
13 Youtube Channels created
------------------------------------
Publishers data has been migrated to owners and channels

```

JSON representations of the objects being created is output during the run to aid in troubleshooting.

Fixes #477

Ideally this will be run against a copy of the production data and results should be carefully inspected.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/lib/tasks/temp/transform_legacy_data.rake
+++ b/lib/tasks/temp/transform_legacy_data.rake
@@ -1,0 +1,223 @@
+require "legacy_data"
+
+namespace :publishers do
+  desc "Import legacy publishers"
+  task :transform_legacy_data, [:commit] => [:environment] do |t, args|
+
+    extend LegacyData
+
+    unless args[:commit]
+      puts "Performing a trial run. Changes will be rolled back"
+    end
+
+    def separator
+      puts "------------------------------------\n"
+    end
+
+    # `totp_registrations` are a one-to-one relationship
+    # A Publisher/Owner should only have one TOTP active at a time
+    # If more than one per owner we won't import them automatically. Owners will be contacted and registrations will be imported
+    def add_totp_registration(owner, legacy_publisher)
+      totp_registration = legacy_publisher.legacy_totp_registration
+      return unless totp_registration
+
+      count_totp = LegacyData::LegacyTotpRegistration.joins(:legacy_publisher).where("legacy_publishers.email": legacy_publisher.email).count
+      if count_totp > 1
+        puts "Skipping totp import for #{legacy_publisher.email}. #{count_totp} records found."
+        return
+      end
+
+      params = {
+          secret: totp_registration.secret,
+          publisher_id: owner.id,
+          last_logged_in_at: totp_registration.last_logged_in_at,
+          created_at: totp_registration.created_at
+      }
+
+      new_totp_registration = TotpRegistration.new(params)
+      new_totp_registration.save!
+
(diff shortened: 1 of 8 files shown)


lib/tasks/temp/transform_legacy_data.rake
```diff
@@ -0,0 +1,223 @@
...
+      end
+
+      params = {
+          secret: totp_registration.secret,
+          publisher_id: owner.id,
+          last_logged_in_at: totp_registration.last_logged_in_at,
+          created_at: totp_registration.created_at
+      }
+
+      new_totp_registration = TotpRegistration.new(params)
```
ayumi: You might need to manually assign `#secret` here.

lgebhardt:
Fixed two issues: 
* a bug in `attribute_encrypted` was preventing access to the decrypted values, which was fixed by loading the Legacy models inside the rake task
* more importantly a wrong assumption on my part that we could decrypt a value from the rename legacy table

Now the values are copied as simple text attributes, bypassing encryption issues.

Drops support for moving over the uphold intermediate values. There should be very few (likely none given it's got a five minute window at most) that will be affected and in case they are the user will just see they need to connect.

PR #501, opened 2018-01-19
Enable channels to be removed

### LINKED ISSUE #385: Multi-channel support: dashboard channel list
For now, the channel info is limited to
1. channel icon: favicon for the website (if we can get it easily, otherwise generic icon shown in the first mockup below), YT channel avatar
2. channel name: the same, domain or YT channel name
3. date that the channel is added
4. action, 'Remove Channel' --> will get user's confirmation on a dialog
5. 'Brave Verified Pub' opt-in toggle
6. Status: if not verified -- "Verification in progress" or "Your website is not approved"
7. Your website is not approved accompanies an action, 'Request Review' link.

*Uphold wallet card: connection status -- Connected (green circle) or Not Connected (gray circle) with the ability to disconnect.

*the opt-in toggle is removed from statement/contact card; becomes channel-specific.

8-publisher dashboard

9-pubdash2

### LINKED ISSUE #413: Multi-channel support: allow user to resume verification process

dgeb:
A `remove` link per channel on the dashboard deletes the channel with an xhr and, if successful, animates the channel row off the page. A confirm dialog is used to ensure that the publisher really wants to delete the channel.

The ChannelsController is updated to deal with xhr json requests and no longer deals with redirects or flash updates.

closes #385, closes #413

cc @ayumi 

Submitter Checklist:

- [X] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


dgeb:
> Do you know if we need to run PublisherChannelSetter again after a removal?

Just chatting with @lgebhardt about this. We definitely need to run a service to sync with Eyeshade, but it's unclear right now whether the `/v2/owners` endpoint (used by `PublisherChannelSetter`) will automatically handle deleting channels.

@mrose17 If we POST to `/v2/owners` with a new set of channels, will eyeshade diff the channels and delete missing ones?

dgeb:
According to @mrose17: 

> if you want to remove a channel from a publisher, please use ` DELETE /v1/owners/{owner}/{publisher}`

I'll introduce a new service to perform channel removal and then update this PR.

dgeb:
@ayumi Ok, thanks for testing! 🎯 

> youtube channel removal works

Nice 😎 

> should I be able to remove a site channel verification in progress? like when you didn't get past the Verify by adding your DNS record

Yes, according to @jenn-rhim in #413 (see the in-progress "amazingblog.com" example in the mock). The `Remove` link just appears to the right. This PR includes that link for unverified channels here: https://github.com/brave-intl/publishers/pull/501/commits/e2d7243ddd21790fd382dc706b4921231003ed7a#diff-3f493ac8654a07a9dbbeb5563811e0d8R372 

> also merge conflict

Just resolved ✅

Issue #383, opened 2017-12-08
Multi-channel support: verification process


jenn-rhim:
- no fundamental changes to this process for now
- except this module is spawn off of the dashboard and returns the user to the dashboard

**Changes:**
1. remove global header for Brave Payments
2. process name, step indicator, and cancel in the process header
    -- process name: Add Channel
    -- step indicator: no half circle needed (every step is a whole circle)
    -- cancel returns to the dashboard 

- **no header** on the channel type screen:
![7-1-addch-type](https://user-images.githubusercontent.com/4369856/33748700-d78b0432-db7e-11e7-8f6b-0843cb081ef1.png)

- remove all other fields except 'domain' field on the domain info screen:
![5-step1 domain info](https://user-images.githubusercontent.com/4369856/33748740-21f43052-db7f-11e7-81be-1d6db5887d5b.png)

- verification method screen:
![6-step2 verif method](https://user-images.githubusercontent.com/4369856/33748786-5584db38-db7f-11e7-9eec-4f38616bb164.png)

dgeb:
@ayumi I believe this can closed via #501

PR #503, opened 2018-01-19
Custom navbar for adding site channels

dgeb:
Reworks the navbar in accordance with @jenn-rhim’s requests in #383:

* Show `Add Channel` on the left in place of `Brave Payments`. This is now text and not a link.
* Removes the login panel on the right in place of a `Cancel` button that returns you to the dashboard.

Does not attempt to rework the progress train settings because more clarification is needed. There are unspecified states in #383 during the verification process.

screen shot 2018-01-19 at 4 56 03 pm

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/views/site_channels/_nav.html.slim
+++ b/app/views/site_channels/_nav.html.slim
@@ -1,0 +1,12 @@
+nav.navbar.navbar-default.navbar-static-top.top-nav-collapse
+  .container-fluid
+    .menu-container
+      .nav.pull-right
+        = link_to(I18n.t("shared.cancel"), home_publishers_path, class: 'btn btn-secondary btn-highlight btn-cancel pull-right')
+
+      .nav.pull-left
+        / = link_to(root_path) do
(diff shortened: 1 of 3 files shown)


app/views/site_channels/_nav.html.slim
```diff
@@ -0,0 +1,12 @@
+nav.navbar.navbar-default.navbar-static-top.top-nav-collapse
+  .container-fluid
+    .menu-container
+      .nav.pull-right
+        = link_to(I18n.t("shared.cancel"), home_publishers_path, class: 'btn btn-secondary btn-highlight btn-cancel pull-right')
+
+      .nav.pull-left
+        / = link_to(root_path) do
+        .navbar-brand.page-title= t "publishers.add_channel"
```
ayumi: minor: if we ever add non-"Add channel" related actions to SiteChannelsController this would show as well.

it is all right for now

PR #483, opened 2018-01-16
Introduce new styles for errors in the header of U2F

mixonic:
This PR closes https://github.com/brave-intl/publishers/issues/393 and closes https://github.com/brave-intl/publishers/issues/394. It introduces a new error style in the header of sub panels.

**Login page**

localhost_3000_publishers_log_in

**Login confirmation**

localhost_3000_publishers_log_in 1

**Pending authentication via U2F**

localhost_3000_publishers_two_factor_authentications 2

**Errored authentication via U2F**

localhost_3000_publishers_two_factor_authentications 1

**Warning on TOTP  re-registration**

localhost_3000_publishers_totp_registrations_new 6

**Error on U2F registration**

localhost_3000_publishers_u2f_registrations_new 1

**Error on U2F registration, mobile**. The other pages also adapt in this way: Less padding and no icon.

localhost_3000_publishers_u2f_registrations_new

**GIF to illustrate flow, responsiveness**

screencast 2018-01-16 11-23-22

TODO:

* [x] Final sanity checks of altered pages, and pages maybe altered by these changes
* [x] Some screenshots
* [x] Add further inline documentation in CSS for sub-panel
* [x] Add further inline documentation in JS for error manager

PR #510, opened 2018-01-23
Add Bootstrap 4 styling for login alert

kpfefferle:
This is a follow-up to #490. I noticed during my testing this morning that we neglected to style the alert that appears if the email address can't be found during login.

Log In Alert

I also performed some follow-up cleanup work that I think we should continue doing along the way during this transition to Bootstrap 4:
- Arrange the login flow translation strings so that they can use template-relative shorthand when possible.
- Sort our `bootstrap-config` variables to match the order that they appear in the original `bootstrap/variables` that they are modifying (with section header comments).

cc: @mixonic

### DIFF
--- a/app/views/layouts/application.html.slim
+++ b/app/views/layouts/application.html.slim
@@ -36,7 +36,11 @@ html
 
       .notifications.navbar-static-top#flash
         - if flash[:alert]
-          .alert.alert-warning.flash== flash[:alert]
+          .alert.alert-warning.flash= flash[:alert]
+        - if flash[:alert_html_safe]
+          / This should only be used in cases where we know the flash content is NOT user-provided
+          / (such as translation strings that include minor formatting or links)
+          .alert.alert-warning.flash== flash[:alert_html_safe]
         - if flash[:notice]
           .alert.alert-success.flash= flash[:notice]
         - if content_for(:content_form_errors)
--- a/app/views/layouts/application.html.slim
+++ b/app/views/layouts/application.html.slim
@@ -36,9 +36,7 @@ html
 
       .notifications.navbar-static-top#flash
         - if flash[:alert]
-          .alert.alert-warning.flash= flash[:alert]
-        - if flash[:login_link]
-          = render partial: "publishers/login_link"
+          .alert.alert-warning.flash== flash[:alert]
         - if flash[:notice]
           .alert.alert-success.flash= flash[:notice]
         - if content_for(:content_form_errors)
--- a/app/views/publishers/create_auth_token.html.slim
+++ b/app/views/publishers/create_auth_token.html.slim
@@ -4,8 +4,8 @@
 .single-panel--wrapper
   .single-panel--content
     .row
-      h3.single-panel--headline = t("publishers.create_auth_token")
+      h3.single-panel--headline= t ".heading"
 
     .row
       .col-small-centered
-        p.text-center= t("publishers.create_auth_token_body")
+        p.text-center= t ".body"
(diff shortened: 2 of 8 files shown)


app/views/publishers/create_auth_token.html.slim
```diff
@@ -4,8 +4,8 @@
 .single-panel--wrapper
   .single-panel--content
     .row
-      h3.single-panel--headline = t("publishers.create_auth_token")
+      h3.single-panel--headline= t ".heading"
```
mixonic: Why no space after the class?

app/views/layouts/application.html.slim
```diff
@@ -36,9 +36,7 @@ html
 
       .notifications.navbar-static-top#flash
         - if flash[:alert]
-          .alert.alert-warning.flash= flash[:alert]
-        - if flash[:login_link]
-          = render partial: "publishers/login_link"
+          .alert.alert-warning.flash== flash[:alert]
```
mixonic: Seems risky- have you audited to make sure nothing is putting user-submitted data into the flash string?

app/views/publishers/create_auth_token.html.slim
kpfefferle: It works either way; that's just how I've always written my Slim tags with computed contents I guess?

mixonic: I would like to pick a consistent style. I don't care much, but I've been using spaces after the class.

I just don't want to think about it reviewing diffs. @ayumi any input?

mixonic:
@jenn-rhim We discussed the styles for flash notices or errors possibly being changed to look like they do in PR #483. Should we change to that style for these errors as well?

app/views/layouts/application.html.slim
kpfefferle: Yes, I checked and only translation strings ever get put into `flash[:alert]`. As with other cases where we've used the html-safe `==`, this lets us lean on translation strings for simple formatting and links rather than using a one-off partials as was previously being done in this case.

app/views/publishers/create_auth_token.html.slim
kpfefferle: The existing templates were also using no-space as the default before we entered the project.

jenn-rhim:
@mixonic
That’d be a good thing. We might as well. With that said, we need to fix the workflow here to be more forgiving: i.e. regardless of the user’s account being in our database, and of log in or sign up, I want the user to be able to move forward and receive email to come back. It should be us figuring out whether it’s a sign up or actually log in to present the correct next screen, as long as there’s no significant work involved for this.

app/views/layouts/application.html.slim
ayumi: so originally I shared this concern of html_yolo, so that's why we created a separate login_link bit -- to isolate the area in which we're using html_safe

can we maybe make a flash[:alert_html_safe] in cases where we really want it?

app/views/publishers/create_auth_token.html.slim
mixonic: I'm happy to go this direction! Wish we could lint it 😭

mixonic:
@kpfefferle and I discussed this and we advocate to merge this work as it. It fixes what is basically a bug about how error states are presented on login and signup on staging2 right now.

We will take on the improved alert/error UI as our next effort.

kpfefferle:
@ayumi Your suggestion for an opt-in `html_safe` alert variant is a good one, so it's now implemented that way 👍 

We consider this good to merge to get the basic style and cleanup work in, and we're going to work on migrating the alert style over to the new look per @jenn-rhim's feedback as a separate PR.

kpfefferle:
@ayumi Whoops, I think we expected this one to get merged before #511, but I've rebased and done some cleanup here so we should be good now.

PR #514, opened 2018-01-23
Bootstrap 4: Sign up alerts

kpfefferle:
- Refactors i18n alert strings handled by the `Publishers#create` action.
- Changes the redirect for these alerts to `sign_up_publishers_path` since the sign up form is no longer on the main landing page.
- Re-enables the Bootstrap 3 alerts CSS on the static landing page as I noticed those were unstyled when an alert redirected there (which I'm sure still affects some auth-related alerts).


kpfefferle:
Note that we will want to follow this up with styling similar to what @mixonic is currently working on for the log in alerts.

PR #515, opened 2018-01-23
Style alert, warning and notice messages in panels

mixonic:
This introduces unified support for `:alert`, `:notice` and `:warning` as flash messages for anywhere in the app and all have `_html_safe` variants as well.

This also introduces styles for different flash messages within a panel on the Bootstrap 4 pages. A panel template should use this render partial:

```slim
.single-panel--wrapper
  = render "panel_flash_messages" // <------
  .single-panel--content
    h3.single-panel--headline Howdy user!
```

With that partial flash messages will no longer appear below the navbar, instead they will appear as part of that panel. For example:

**`flash[:alert] = message`**

screen shot 2018-01-23 at 1 59 55 pm

**`flash[:notice] = message`**

edit: Colors here have changed, see https://github.com/brave-intl/publishers/pull/515#issuecomment-359998380

screen shot 2018-01-23 at 1 59 43 pm

**`flash[:warning] = message`**

screen shot 2018-01-23 at 1 59 29 pm

TODO:

* [x] Rebase over https://github.com/brave-intl/publishers/pull/514 when that is merged.
* [x] Investigate porting this same solution to errors from creating an account (if appropriate).
* [ ] ~Confirm the `notice` colors here with @jenn-rhim. I think the green is suspect. I think we want something signifying "success" and hence green. Currently the background is `#9bf0e1` and the foreground text `#009683` both taken from the master styles palette. Also, should we use an icon here?~ I've tweaked the colors here to be Good Enough, I think. They can definitely be changed down the line, and an icon could be added. See https://github.com/brave-intl/publishers/pull/515#issuecomment-359998380
* [ ] ~The `alert` colors are background `#eefbfd` and foreground text `#00bcd6`, these are based on the current site. The text color is the exact same as our link color, so links are basically invisible. Seems bad. /cc @jenn-rhim~ I've added an underline to links in this area. See https://github.com/brave-intl/publishers/pull/515#issuecomment-359999616
* [x] Document the usage of `= render "panel_flash_messages"`.
* [x] Centralize some documentation for flash options in general.

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -45,7 +45,8 @@ class PublishersController < ApplicationController
     email = params[:email]
 
     if email.blank?
-      return redirect_to sign_up_publishers_path, alert: t(".missing_email")
+      flash[:warning] = t(".missing_email")
+      return redirect_to sign_up_publishers_path
     end
 
     @publisher = Publisher.new(pending_email: email)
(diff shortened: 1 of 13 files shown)


mixonic:
I've updated this PR to show sign up errors in the panel as well. Some of these previously were just swallowed silently, so this audit was a useful one.

**User submits with no email**

This is validated on the client side, so shouldn't be common.

screen shot 2018-01-23 at 5 03 22 pm

**There is an unexpected error saving the new user**

screen shot 2018-01-23 at 5 07 21 pm

**User tries to sign up with an email that already has an account**

Oddly this previously emailed the user a login link without telling them what happened. They were sent to the "you signed up" page without their email showing up 😬.

screen shot 2018-01-23 at 5 06 18 pm

**User is throttled**

screen shot 2018-01-23 at 5 03 44 pm

**User successfully signs up!**

screen shot 2018-01-23 at 5 06 44 pm

mixonic:
I've changed the green here to be:

```
background-color: #e2f8f3; /* theme-color-level('brave-success' aka #6cdac1, -10) */
color: #387164; /* theme-color-level('brave-success' aka #6cdac1, 6) */
```

The use of `theme-color-level` is based on how [alerts in bootstrap are created](https://github.com/twbs/bootstrap/blob/ee83a963f3f1ec318b62eafa5c123cb2ea8eb37f/scss/_alert.scss#L49).

I think this is an improvement, here is a screenshot:

screen shot 2018-01-23 at 6 13 10 pm

mixonic:
To address links being invisible in the `alert` sections I've added an underline. This should be Good Enough to move forward. An example:

screen shot 2018-01-23 at 6 23 42 pm

app/controllers/publishers_controller.rb
```diff
@@ -87,8 +89,8 @@ def resend_email_verify_email
 
     PublisherMailer.verify_email(@publisher).deliver_later
     PublisherMailer.verify_email_internal(@publisher).deliver_later if PublisherMailer.should_send_internal_emails?
+
     session[:created_publisher_id] = @publisher.id
-    session[:created_publisher_email] = @publisher.pending_email
```
mixonic: Unused.

jenn-rhim:
looking good, @mixonic !

PR #518, opened 2018-01-24
Bootstrap 4: Panel flash followup

kpfefferle:
This is a follow up to review of #515. The only change I make here is that I recommend that we stick to the convention within Slim templates of using static classes when possible in place of the `div` selector (or attached directly to other elements) and only use `class=""` for dynamically generated class names. IMO, this helps make it readily apparent which classes are statically applied vs. dynamically generated.

cc: @mixonic

### DIFF
--- a/app/views/application/_panel_flash_messages.html.slim
+++ b/app/views/application/_panel_flash_messages.html.slim
@@ -17,7 +17,7 @@
 - [:alert, :notice, :warning].each do |name|
   - {alert: 'icon_circled_i', warning: 'icon_circled_x'}[name].tap do |icon_partial|
     - if flash[name]
-      div class="single-panel--content single-panel--content--#{name} icon-and-text--wrapper"
+      .single-panel--content.icon-and-text--wrapper class="single-panel--content--#{name}"
         - if icon_partial
           .icon-and-text--icon
             = render icon_partial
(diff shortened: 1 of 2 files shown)


app/views/application/_panel_flash_messages.html.slim
```diff
@@ -17,15 +17,15 @@
 - [:alert, :notice, :warning].each do |name|
   - {alert: 'icon_circled_i', warning: 'icon_circled_x'}[name].tap do |icon_partial|
     - if flash[name]
-      div class="single-panel--content single-panel--content--#{name} icon-and-text--wrapper"
+      .single-panel--content.icon-and-text--wrapper class="single-panel--content--#{name}"
```
mixonic: This also changes the order of application for these classes, and possibly the precedence of some styles.

AFAICT it should not make a difference in this case, but it seems wise for a developer to be sure they know how the classes interact before swapping the order simply for code readability.

kpfefferle: This is a good observation. Should we encounter a case where the classes must be applied in a certain order, that could certainly trigger an exception to this convention. I'd also consider that a code smell that something fishy is going on with the CSS (as I would consider the need to apply classes in a particular order to be a CSS anti-pattern).

PR #519, opened 2018-01-24
Verification flow and API updates

lgebhardt:
Updates for eyeshade multichannel compatibility
- update api endpoints (note: removes api/owners controller, which was assumed to be needed, but was not)
- update serializers for eyeshade compatibility
- update site verification flow
- user is now taken to the verification_background page after verify
- verification_background will be modified to poll and return the user
  to the dashboard when verification is complete

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


db/schema.rb
```diff
@@ -28,6 +28,18 @@
     t.index ["publisher_id"], name: "index_channels_on_publisher_id", using: :btree
   end
 
+  create_table "legacy_publisher_statements", id: :uuid, default: -> { "uuid_generate_v4()" }, force: :cascade do |t|
```
ayumi: i think this PR doesn't need schema.rb updates (from #494 ?)

lgebhardt:
@ayumi The schema changes should be out of the log now.

PR #523, opened 2018-01-25
Bootstrap4: Convert the 2FA flows

mixonic:
This converts the 2FA flows to Bootstrap 4.

* Adds `pages/` folder to the styles. This is for utility classes specific to a single page, not ones re-used across other pages on the site.
* Add `.btn-wide` for full-width buttons
* Tweak `.col-small-centered` and the single panel styles to have nicer responsiveness.
* Include a commit to add padding to the bottom of a page if the panel is taller than the viewport.
* Bring the error-group styles over from legacy
* Bring the `.js-feature-u2f-available` styles over from legacy

**Security key attempt**

screen shot 2018-01-24 at 9 02 50 pm

**Security key error**

screen shot 2018-01-24 at 9 03 33 pm

**TOTP**

screen shot 2018-01-24 at 9 02 23 pm

**TOTP error**

screen shot 2018-01-24 at 9 02 39 pm

**User only has security key, but is using an unsupported browser**

screen shot 2018-01-24 at 9 03 48 pm

**User has security key and TOTP, but is using an unsupported browser**

screen shot 2018-01-24 at 9 04 59 pm

TODO

* [ ] Use the proper flash style for TOTP errors to make them appear as warnings.
* [ ] Ensure the "security key unsupported" headline is not inside the small column.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -10,3 +10,4 @@
 @import "application-config";
 @import "components/**/*";
 @import "theme/**/*";
+@import "pages/**/*";
--- a/app/assets/stylesheets/components/btn.scss
+++ b/app/assets/stylesheets/components/btn.scss
@@ -1,0 +1,3 @@
+.btn-wide {
+  width: 100%;
+}
--- a/app/assets/stylesheets/pages/two_factor_authentications.scss
+++ b/app/assets/stylesheets/pages/two_factor_authentications.scss
@@ -1,0 +1,9 @@
+.totp-authentication {
+  svg {
+    margin-top: 4px;
+    margin-right: 16px;
+  }
+}
+.two-factor-authentication-alternative {
+  margin-top: $spacer;
+}
--- a/app/assets/stylesheets/theme/columns.scss
+++ b/app/assets/stylesheets/theme/columns.scss
@@ -4,6 +4,10 @@
     @include make-col(10);
     @include make-col-offset(1);
   }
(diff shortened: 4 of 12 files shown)


jenn-rhim:
my comment on overall:

1. error messages need to be trimmed. too verbose at the moment. i will provide the copy.

2. need to give more margins between the content and the container and the where the container starts vertically.

ayumi:
@jenn-rhim sorry for the quick merge

jenn-rhim:
@ayumi no problem!  we can keep improving. :)

@mixonic -- here is my input regarding the error messages (only)
1. Security key attempt: NA
2. Security key error
- msg: "Timed out detecting your security key."
3. TOTP: NA
4. TOTP error
- msg: "Invalid authentication code."
- wrong icon and color: x and red
- looks like the error message is not middle v-aligned for this one

5. User only has security key, but is using an unsupported browser (wrong title for the screen)-- i'm going to go by the screen:
User has security key and TOTP, but is using an unsupported browser
- "Your preferred method of two-factor authentication is security key. But it's not supported by your current browser. Switch to Google Chrome or Opera or login using authentication code below."
- yellow is fine, but we need the warning icon "!"

6. User has security key and TOTP, but is using an unsupported browser -- wrong title
For the screen, User only has security key, but is using an unsupported browser
- title -- "Security key not supported by your browser."
-msg:  "Switch to Google Chrome or Opera to log in. We recommend setting up authenticator app as a backup method for two-factor authentication for this situation."

PR #547, opened 2018-01-26
Show modal when channel has already been claimed

### LINKED ISSUE #478: Multichannel Support: Attempt to claim an already taken domain results in blank page
**Problem:** A user attempts to add a site channel that has already been claimed. System initially shows the Choose Verification Method page. Choosing either option results in a blank page (on the multi channel WIP).

-----

The production site has a bland message shown in this situation:

screen shot 2018-01-11 at 6 07 57 pm

Where as for already taken Youtube Channels a nice modal dialog is shown:

screen shot 2018-01-11 at 6 08 38 pm

I'd like to implement a similar modal dialog when the blank page issue is fixed.

dgeb:
Restores the modal dialog that’s shown when a YouTube channel has already been registered. This dialog has been enhanced to work with web site channels as well. Both versions also have a `noscript` equivalent.

For YouTube channels which have already been registered, the person who registered the channel and the registration date are shown. This has been seen as acceptable since the user will not see this message without already authenticating themselves as a manager for the channel.

For web sites, the registration information is omitted because no authorization step is performed prior to showing the “channel taken” dialog. I'm concerned that showing registration information with no authorization would allow users to discover everyone who has registered a site on Brave.

Closes #478 

/cc @ayumi @jenn-rhim 

Site channel version:

screen shot 2018-01-26 at 4 24 26 pm

YouTube channel version:

screen shot 2018-01-26 at 4 24 42 pm

noscript version:

screen shot 2018-01-26 at 4 23 22 pm

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


jenn-rhim:
@dgeb for the web site, the user would receive this warning message at the first step of verifying the site when they entered the domain name. I'd like to give them a chance to re-enter the domain or enter another domain in the same page. They can always cancel out the process. Here is how the message would come up.

**note:** let's make sure that we're not using the term 'registration' for adding the channels: it's either 'adding' or 'verifying' depending on the context.

step1 domain info-taken

dgeb:
@jenn-rhim Thanks for the feedback. I'm working on the tweaks now!

dgeb:
@jenn-rhim I've updated this PR based on your feedback. As we discussed, the behavior and new verbiage you've requested has been updated. However, the styling of the alert notification on the `Enter Domain Info` page won't be updated until Bootstrap4 is applied to this zone of the site.

The duplicate domain message is still shown with the old style flash, which doesn't look great but at least gets the message across:

screen shot 2018-01-29 at 2 47 15 pm

Issue #509, opened 2018-01-23
Add the 2fa confirmation screen instead of the set up screen for smooth transition.


jenn-rhim:
@mixonic 

For the first time 2fa setup, user will be doing this as part of the sign up process, not by going to the Account page. 

The attached screen seems more appropriate and also provides clear direction to the dashboard. If it's easier, a full page rather than an overlay is fine by me as long as 'Close' button takes the user to the dashboard. 

- Current screen
![screen shot 2018-01-22 at 3 52 31 pm](https://user-images.githubusercontent.com/4369856/35251506-546b549a-ff90-11e7-9e23-0a433fac1c61.png)

- Proposing content change
![2fa-confirm](https://user-images.githubusercontent.com/4369856/35251524-6627021a-ff90-11e7-8f2b-47b44a5d98bc.png)

mixonic:
@jenn-rhim this is only for first-time setups, correct?

jenn-rhim:
@mixonic yep, correct.  the user would feel lost as to what to do next given that they just signed up.

mixonic:
For reference the initial implementation of this feature was in https://github.com/brave-intl/publishers/pull/457/files

PR #565, opened 2018-01-30
Add a success modal for signup 2FA registration

mixonic:
Closes https://github.com/brave-intl/publishers/issues/509, initial implementation in https://github.com/brave-intl/publishers/pull/457.

* A commit to ignore `.DS_Store` files
* A commit to fix some style regressions on the 2FA prompt during account setup. Screenshots below.
* Add a success modal to `/publishers/home` if the registration is the result of a prompt.
* Add an abstraction for `flash[:modal_partial]` which opens that partial into the same modal used for the confirmation flows.
* These pages are legacy Bootstrap 3 pages, and this work does not convert them over to Bootstrap 4 pages. The styles seem close enough for the Bootstrap 3 pages, as we convert them to Bootstrap 4 we can probably improve them (font sizes, fonts, panel sizes). Thus I suggest shipping these styles as is.

TODO:

* [x] Add documentation for `flash[:modal_partial]`
* [x] Add a test confirming this behavior
* [x] Swap icon to inline SVG, drop un-used assets

Screenshots follow.

**Prompt to register during signup, desktop**

localhost_3000_publishers_two_factor_registrations_prompt

**Prompt to register during signup, mobileish**

localhost_3000_publishers_two_factor_registrations_prompt 1

**Success message after signup 2FA registration, desktop**

localhost_3000_publishers_home

**Success message after signup 2FA registration, mobileish**

localhost_3000_publishers_home 1


kpfefferle:
@mixonic I updated the styles for the prompt view to Bootstrap 4 as part of my 2FA Registrations work. I plan to PR everything in the AM after I do a final review of responsiveness (while grabbing screenshots for the PR), but you can check out what I've done at https://github.com/kpfefferle/publishers/tree/bootstrap4-2fa-registration

.gitignore
```diff
@@ -15,6 +15,7 @@
 clifton.sh
 dump.rdb
 yarn-error.log
+.DS_Store
```
kpfefferle: I'm generally of the opinion that system-specific `.gitignore` content should be in the developer's global `.gitignore` file rather than the project file (that's where I ignore `.DS_Store`). But if the team is mostly on macOS, it's not hurting to have it here I suppose.

app/assets/images/2fa-good.svg
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="60px" height="72px" viewBox="0 0 60 72" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 48.2 (47327) - http://www.bohemiancoding.com/sketch -->
```
kpfefferle: Did this export not use the SVGO Compressor plugin? This content usually gets removed.

app/views/modal_partials/_two_factor_registration_complete.html.slim
```diff
@@ -0,0 +1,10 @@
+.col.col-center.col-xs-12.col-sm-8.col-md-8.col-lg-9.text-center
+  = image_tag "2fa-good@2x.png", width: 60, height: 72
```
kpfefferle: Are we not using the SVG then? This PR includes both SVG and PNG icons...

mixonic: I'm going to swap this PR over to use an inline SVG 👍

mixonic:
@ayumi @jenn-rhim I think this is a good step forward. All legacy styles, but looking reasonable. Per https://github.com/brave-intl/publishers/issues/509 it improves the setup UX by sending the user to the dashboard after a first 2FA registration instead of sending them to the 2FA dashboard. 🎉

PR #571, opened 2018-01-31
Change 2FA auth unsupported messages to use panel

### LINKED ISSUE #551: 2FA auth flows cleanup
Mostly followup to https://github.com/brave-intl/publishers/pull/523

* [x] Use the proper flash style for TOTP errors to make them appear as warnings. (https://github.com/brave-intl/publishers/pull/559)
* [x] Ensure the "security key unsupported" headline is not inside the small column. (https://github.com/brave-intl/publishers/pull/562)
* [x] Middle align error messages when they are a single line (https://github.com/brave-intl/publishers/pull/559)
* [x] If user has U2F and TOTP but no U2F support change the message to: "Your preferred method of two-factor authentication is security key. But it's not supported by your current browser. Switch to Google Chrome or Opera or login using authentication code below." (https://github.com/brave-intl/publishers/pull/562)
* [x] If user has U2F and no TOTP, but no U2F support change the message to: "Switch to Google Chrome or Opera to log in. We recommend setting up authenticator app as a backup method for two-factor authentication for this situation." and the title to "Security key not supported by your browser." (https://github.com/brave-intl/publishers/pull/562)
* [x] Migrate our flash message type options to: `alert -> red` `warning -> yellow` `notice -> blue`. Yellow is new, use the mock below for it an its icon. Update the docs and existing messages. Remove the green state entirely. (https://github.com/brave-intl/publishers/pull/559)

I'll need a few things from @jenn-rhim 

* [x] ~Regarding the unsupported screens on https://github.com/brave-intl/publishers/pull/523 you said "yellow is fine, but we need the warning icon (!)" - I'll need a mock to proceed. I'm concerned that the text in that yellow box would then be extremely narrow and difficult to read.~ After the mock below was added @jenn-rhim and I discussed this over Slack. She is going to mock up the two use cases in detail: a) The user has security key and code, but their browser does not support the key. b) The user only has security key, but their browser does not support the key. (https://github.com/brave-intl/publishers/pull/571)

mixonic:
Closes #551. Also, drop un-used styles for 2FA auth from the legacy folder.

**User has security key and authenticator app, security key unsupported**

localhost_3000_publishers_two_factor_authentications 6

**User has only security key, security key unsupported**

localhost_3000_publishers_two_factor_authentications 5


app/views/two_factor_authentications/_u2f.html.slim
```diff
@@ -19,22 +19,24 @@ div.js-feature-u2f-available
...
 
-    = render "totp" do
-      .col-small-centered
-        p.alert.alert-warning = t("two_factor_authentications.u2f.unavailable.totp_available")
-        br
+    .col-small-centered
+        p = render "application/usb"
+        hr.two-factor-authentication--alternative-hr
+        p
+          = link_to "Use Authentication Code Instead", two_factor_authentications_path(request_totp: true), class: 'btn btn-primary btn-wide'
```
mixonic: Whoops, need translation here

app/views/two_factor_authentications/index.html.slim
```diff
@@ -4,47 +4,56 @@
...
+      .single-panel--content.single-panel--content--alert.icon-and-text--wrapper
         .icon-and-text--icon
-          = render 'icon_circled_x'
+          = render "icon_circled_x"
         .icon-and-text--text
           = t ".u2f-error.bad-request"
 
     .js-u2f-error-configuration-unsupported
-      .single-panel--content.single-panel--content--warning.icon-and-text--wrapper
+      .single-panel--content.single-panel--content--alert.icon-and-text--wrapper
```
mixonic: These were incorrectly in warning styles before. Should be alert/error/red style.

PR #582, opened 2018-01-31
Use an HTML modal for channel removal confirmation

### LINKED ISSUE #576: inconsistent dialog style
1. this dialog style is from the current browser. let's use the brave payments dialog style.
2. in the dialog content: the title should typically echo the action/operation the user invoked. in this case, 'Remove Channel' (see attached below for e.g.)

image

2fa-removekey-1

mixonic:
Closes #576. Stop using `window.confirm`, instead use our HTML dialog.

TODO:

* [x] Add a test to confirm this.

**Confirming a channel removal, desktop**

localhost_3000_publishers_home 2

**Confirming a channel removal, mobileish**

This page is not really responsive, but here is what  it looks like:

localhost_3000_publishers_home 3

### DIFF
--- a/app/assets/stylesheets/legacy/application/publishers.sass
+++ b/app/assets/stylesheets/legacy/application/publishers.sass
@@ -303,7 +303,7 @@ body[data-controller="publishers"]
     .container
       .col-details
         margin-top: 30px
-      h4
+      h4.balance-header
         font-size: 13px
         font-weight: 300
         letter-spacing: 0.5px
--- a/app/javascript/modal.js
+++ b/app/javascript/modal.js
@@ -66,7 +66,7 @@ var MODAL_SHOW_CLASS = 'md-show';
 /*
  * On demand open a modal.
  */
-function openModal(html, confirmCallback, denyCallback) {
+self.openModal = function openModal(html, confirmCallback, denyCallback) {
   var modalElement = document.querySelector('.js-shared-modal');
   var contentElement = modalElement.querySelector('.md-content');
   var containerElement = modalElement.querySelector('.md-container');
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/legacy/application/publishers.sass
```diff
@@ -303,7 +303,7 @@ body[data-controller="publishers"]
     .container
       .col-details
         margin-top: 30px
-      h4
+      h4.balance-header
```
mixonic: This has been driving me nuts. The `h4` style here I believe is only for the balance header, but it leaks out to all publisher pages. The long-term fix is migrating this page to bootstrap 4, this is a cheap hack.

app/javascript/modal.js
```diff
@@ -66,7 +66,7 @@ var MODAL_SHOW_CLASS = 'md-show';
 /*
  * On demand open a modal.
  */
-function openModal(html, confirmCallback, denyCallback) {
+self.openModal = function openModal(html, confirmCallback, denyCallback) {
```
mixonic: Expose `openModal` to the `window` so we can invoke it from the template JS

app/javascript/modal.js
```diff
@@ -85,15 +85,14 @@ function openModal(html, confirmCallback, denyCallback) {
       return;
     }
 
+    modalElement.removeEventListener('click', confirmationEventDelegate);
+    contentElement.innerHTML = '';
+    containerElement.classList.remove(MODAL_SHOW_CLASS);
```
mixonic: Earlier uses of this modal utility always resulted in a page reload/change after success. In this case, there is no page reload and so we change the implementation here to always clear the modal.

mixonic:
I've added tests for this flow. Good to go!

ayumi:
a merge conflict

mixonic:
@ayumi rebased, pushed and passed!

Issue #508, opened 2018-01-22
Treat log in the same as sign up in the workflow and don't make them go to sign up screen


jenn-rhim:
@lgebhardt 

Fundamentally, our log in and sign up processes are the same: a simple email roundtrip. The reason we have them separate on the UI is to support the user's current mental model and accommodate the familiarity. If the user selected log in screen and entered an email that is not in our database, let's just treat it as the sign up request. 

![screen shot 2018-01-22 at 3 38 11 pm](https://user-images.githubusercontent.com/4369856/35250299-cd715e62-ff8a-11e7-976d-a44c3c3db2c4.png)

mixonic:
Per a conversation on Slack: This should be automatically log in a user if their email address cannot be found. Instead it should prompt them in the flash notice with a link to register instead of logging in. For example:

> Email not recognized. Would you like to sign up with the email _email-address@was-not-found.com_ instead?

For reference: https://bravesoftware.slack.com/archives/C6KTLB6G4/p1516741882000199?thread_ts=1516730974.000200&cid=C6KTLB6G4

mixonic:
We actually already kind of do this:

<img width="578" alt="screen shot 2018-01-31 at 1 39 54 pm" src="https://user-images.githubusercontent.com/8752/35648940-45ea6750-068c-11e8-81ab-c68f03087260.png">

I can change the text and make the link actually create the account. Currently it links to the sign up page.

PR #588, opened 2018-01-31
Prompt the user to create an account if they have a bad email on login

mixonic:
Closes https://github.com/brave-intl/publishers/issues/508

**Before**

brave payments 2018-01-31 14-03-57

**After**

brave payments 2018-01-31 14-02-50

/cc @ayumi @jenn-rhim

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -38,7 +38,7 @@ class PublishersController < ApplicationController
     only: %i(home)
 
   def sign_up
-
+    @publisher = Publisher.new(publisher_create_auth_token_params)
   end
 
   def create
@@ -162,6 +162,7 @@ class PublishersController < ApplicationController
   end
 
   def create_auth_token
+    email = publisher_create_auth_token_params[:email]
     @publisher = Publisher.new(publisher_create_auth_token_params)
     @should_throttle = should_throttle_create_auth_token? || params[:captcha]
     throttle_legit =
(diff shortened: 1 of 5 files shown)


app/controllers/publishers_controller.rb
```diff
@@ -173,13 +174,16 @@ def create_auth_token
...
+    emailer = PublisherLoginLinkEmailer.new(email: email)
 
     if emailer.perform
       # Success shown in view #create_auth_token
     else
       # Failed to find publisher
-      flash.now[:alert_html_safe] = t('.unfound_alert_html', link: sign_up_publishers_path)
+      flash.now[:alert_html_safe] = t('.unfound_alert_html', {
+        create_publisher_path: publishers_path(email: email),
+        email: ERB::Util.html_escape(email)
```
ayumi: nice

config/locales/en.yml
```diff
@@ -152,8 +152,9 @@ en:
       access_throttled: You've attempted to sign up too many times and your access had been throttled. Try again later.
     create_auth_token:
       unfound_alert_html: |
-        Couldn't find a publisher with that email address. Please try again or
-        <a href="%{link}">sign up for an account</a>.
+        Couldn't find a publisher with that email address. Please try again, or
+        you may
+        <a data-method="post" href="%{create_publisher_path}">create an account with the email %{email}</a>.
```
ayumi: minor: this doesn't work without javascript and falls back to a 404 at `[GET] "/publishers"`

is there a way to make it work?

mixonic:
@ayumi I've added and documented a way to pass a URL for UJS to override `href`. This allows a link to have two URLs, an `href` that is no-JS safe and a `data-href` which, if present, overrides `href` for UJS links.

Testing this is awkward, we don't have a good way to turn off JS for a single test. I suggest landing it without a test for now and we can try to add no-JS support to the test suite elsewhere.

PR #594, opened 2018-02-01
Regression fix: Sign up was requiring an email query param

mixonic:


mixonic:
FYI I force pushed to this branch at least once to make a small improvement to the test.

PR #601, opened 2018-02-02
Refactor i18n strings to use lazy (organized) form

kpfefferle:
This PR completes the i18n string cleanup that we had started during Bootstrap 4 migration. It reorganizes the strings file to use [Rails i18n lazy lookup](http://guides.rubyonrails.org/i18n.html#lazy-lookup) as much as possible, which results in a much more organized locales file as well as cleaner templates.

This completes the remainder of the i18n transition work so that the rest of Bootstrap 4 migration can focus wholly on template and CSS updates.


dgeb:
Thanks for tackling this @kpfefferle!

PR #605, opened 2018-02-06
use case insensitive find by email for logins

### LINKED ISSUE #604: Login lookups by email are case sensitive
Lookups by email need to be case insensitive.

lgebhardt:
Fixes #604

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -24,7 +24,7 @@ class Publisher < ApplicationRecord
   phony_normalize :phone, as: :phone_normalized, default_country_code: "US"
 
   validates :email, email: { strict_mode: true }, presence: true, unless: -> { pending_email.present? }
-  validates :email, uniqueness: true, allow_nil: true
+  validates :email, uniqueness: {case_sensitive: false}, allow_nil: true
   validates :pending_email, email: { strict_mode: true }, presence: true, if: -> { email.blank? }
   # validates :name, presence: true, if: -> { brave_publisher_id.present? }
   validates :phone_normalized, phony_plausible: true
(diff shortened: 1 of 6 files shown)


app/models/publisher.rb
```diff
@@ -43,6 +43,8 @@ class Publisher < ApplicationRecord
   before_create :build_default_channel
   before_destroy :dont_destroy_publishers_with_channels
 
+  scope :case_insensitive_find_by_email, -> (email_to_find) { where('lower(publishers.email) = :email_to_find', email_to_find: email_to_find.downcase) }
+
```
ayumi: Can we also add update the email validation to include case_insensitive: true

L27
```
validates :email, uniqueness: {case_sensitive: false}, allow_nil: true
```

app/models/publisher.rb
```diff
@@ -43,6 +43,8 @@ class Publisher < ApplicationRecord
   before_create :build_default_channel
   before_destroy :dont_destroy_publishers_with_channels
 
+  scope :case_insensitive_find_by_email, -> (email_to_find) { where('lower(publishers.email) = :email_to_find', email_to_find: email_to_find.downcase) }
```
ayumi: could we rename the method to `#find_by_email_case_insensitive` ?
it seems like that would be consistent with other finder methods.

lgebhardt:
@ayumi Using `find_by_email_case_insensitive` caused problems when running `rake db:setup`, which causes the travis setup step to fail and makes rebuilding the database fail on dev systems. So I took the easy way out and named the scope `by_email_case_insensitive`. I'm open to other names.

I also changed the email index since your last review so it's case insensitive.

PR #614, opened 2018-02-14
Adding youtube login back in

### LINKED ISSUE #610: Youtube login for publishers that do not have verified email addresses
There was a change that allowed some publishers to sign up with only a youtube channel authentication. When we implemented multichannel this was over looked.  Now these users can not login because Youtube login was disabled. It will soon be replaced with google login (#274), but that will not allow access to these publishers.

In order to allow these users to again access the site we need to allow them to login with their youtube channel (brand account) again.

From Slack:
- we want to prioritize getting users back into their accounts.
- calculating the flag is fine, based on funny email only is ok
- when they successfully google (youtube) login and have a funny email, at the dashboard we should show a modal that says "Please update your account with a valid email address.".  on the dashboard, we should highlight the email address area to grab the user's attention
- user should still be able to interact with the dashboard

lgebhardt:
This PR adds OAuth login for publishers that were created through the login attempt of a youtube channel.

Access to this login process is through a non public link "/publishers/auth/youtube_login".

The OAuth logins have also been reworked to handle login separately from channel authentication. The changes require the creation of two new "Authorized redirect URIs". For example these are the URIs for a development server, and must be [entered](https://console.developers.google.com/apis/credentials) before merging this PR:

`http://localhost:3000/publishers/auth/register_youtube_channel/callback`
`http://localhost:3000/publishers/auth/youtube_login/callback`

and we can remove the old one once the changes are deployed:
`http://localhost:3000/publishers/auth/google_oauth2/callback`

**Todo** Still waiting on some additional wording changes for the login screen, and on the screen after the change email has been sent. That is currently reusing the "An email is on its way!" which is sent after sign-up.

Fixes #610 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/publishers/omniauth_callbacks_controller.rb
+++ b/app/controllers/publishers/omniauth_callbacks_controller.rb
@@ -1,10 +1,14 @@
 module Publishers
   class OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
-    # ToDo: rework
-    before_action :require_publisher
+    include PublishersHelper
 
-    def google_oauth2
+    # ToDo: rework
+    before_action :require_publisher, only: [:register_youtube_channel]
+    before_action :require_publisher_not_created_through_youtube_auth,
+                  only: %i(register_youtube_channel)
+
+    def register_youtube_channel
       oauth_response = request.env['omniauth.auth']
       token = oauth_response.credentials.token
 
@@ -58,6 +62,38 @@ module Publishers
       return
     end
 
+    def youtube_login
+      if current_publisher
+        sign_out(current_publisher)
+      end
+
+      oauth_response = request.env['omniauth.auth']
+
+      channel_details = YoutubeChannelDetails.where(auth_user_id: oauth_response.uid).
+          where.not(youtube_channel_id: nil).first
+
+      if channel_details.nil?
+        redirect_to new_auth_token_publishers_path, notice: t(".channel_not_eligable_for_youtube_login")
+        return
+      end
+
+      publisher = channel_details.channel.publisher
+
+      # if publisher.email != oauth_response.dig('info', 'email')
+      unless youtube_login_permitted?(channel_details.channel)
+        redirect_to new_auth_token_publishers_path, notice: t(".channel_not_eligable_for_youtube_login")
+        return
+      end
+
+      session['google_oauth2_credentials_token'] = oauth_response.credentials.token
+
+      unless current_publisher
+        sign_in(:publisher, publisher)
+      end
+
+      redirect_to change_email_publishers_path
+    end
+
     def after_omniauth_failure_path_for(scope)
       publisher = current_publisher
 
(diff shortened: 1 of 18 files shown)


lgebhardt:
After using the youtube login for a channel:

screen shot 2018-02-16 at 8 25 13 am

Entering your email sends a message:

screen shot 2018-02-16 at 8 30 22 am

When you use the login link you get to this information page:
screen shot 2018-02-16 at 8 30 43 am

And then you get to the normal One last thing page, but the button is labeled Continue in this flow:
screen shot 2018-02-16 at 8 23 11 am

app/controllers/publishers/omniauth_callbacks_controller.rb
```diff
@@ -1,10 +1,14 @@
 module Publishers
   class OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
+    include PublishersHelper
+
     # ToDo: rework
-    before_action :require_publisher
+    before_action :require_publisher, only: [:register_youtube_channel]
+    before_action :require_publisher_created_through_youtube_auth,
+                  only: %i(register_youtube_channel)
```
ayumi: should this be `only: %i(youtube_login)`

app/controllers/publishers_controller.rb
```diff
@@ -26,6 +26,18 @@ class PublishersController < ApplicationController
   before_action :require_verified_email,
     only: %i(email_verified
              complete_signup)
+  before_action :require_publisher_email_not_verified_through_youtube_auth,
```
ayumi: minor: as we add new actions we might forget to whitelist them here.  maybe we could use `:except` here?

config/initializers/devise.rb
```diff
@@ -244,7 +244,16 @@
                   Rails.application.secrets[:google_client_id],
                   Rails.application.secrets[:google_client_secret],
                   {
-                    scope: "profile,email,https://www.googleapis.com/auth/youtube.readonly"
+                      name: 'register_youtube_channel',
```
ayumi: we should update the corresponding part in the README where we describe the oauth redirect URLs

app/views/publishers/change_email.html.slim
```diff
@@ -0,0 +1,24 @@
...
+  = render "panel_flash_messages"
+  .single-panel--content
+
+    h3.single-panel--headline= t ".login_email"
+
+    .col-small-centered
+      p
+        span= t ".login_email_message"
+
+      = form_for @publisher, url: update_email_publishers_path do |f|
```
ayumi: are you able to run this? i tried to change my email and nothing happened.

i think because there's no HTML handler within `Publishers #update`.

we should test that change_email is able to actually change emali

app/controllers/publishers/omniauth_callbacks_controller.rb
lgebhardt: I missed the word `not` in the method_name.

app/controllers/publishers_controller.rb
lgebhardt: Good idea

PR #628, opened 2018-02-19
JS refactoring, cleanup, and tests

dgeb:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/controllers/site_channels_controller.rb
```diff
@@ -4,22 +4,17 @@ class SiteChannelsController < ApplicationController
   before_action :authenticate_publisher!
   before_action :setup_current_channel,
                 except: %i(new
-               create)
+                           create)
   before_action :require_unverified_site,
-                only: %i(email_verified
```
mixonic: Why remove this from the list requiring an unverified site?

Seems harmless either way.

dgeb: @mixonic this list seems to have been carried over from the `PublishersController`, even though `email_verified` and `contact_info` don't exist as actions on the `SiteChannelsController`. I agree it's harmless, but it's also unnecessary noise.

mixonic: Ah thanks I had missed the controller and was reading like this was publishers. 👍 thanks!

PR #630, opened 2018-02-19
Job record not found exceptions

### LINKED ISSUE #629: During tests sidekiq jobs are getting run in the development process
When running the development stack with foreman I've found that after running tests I get a lot of RecordNotFound errors. These seem to be from the tests putting their data into the development redis server.

These RecordNotFound exceptions are not being captured and logged correctly. They should be handled by Raven.

lgebhardt:
This fixes the issue observer in #629 where tests jobs are entered into the development redis queue. it also sends the RecordNotFound errors to Raven. That shouldn't happen on production, but...

Fixes #629

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -6,6 +6,11 @@ class ApplicationJob < ActiveJob::Base
     ErrorHandlerDelegator.new(super)
   end
 
+  rescue_from(ActiveRecord::RecordNotFound) do |exception|
+    require "sentry-raven"
+    Raven.capture_exception(exception)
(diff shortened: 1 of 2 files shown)


ayumi:
would it make sense to move the RecordNotFound checking logic to the base job class?

lgebhardt:
@ayumi Yes, putting the handler in the base job class is much cleaner. I wasn't aware of the `rescue_from` method - thanks.

PR #633, opened 2018-02-20
Testing cleanup

mixonic:
* Use database cleaner for all tests, not just integration
* Update Chromedriver
* Force Chromedriver to use a recent version


test/test_helper.rb
```diff
@@ -25,24 +28,31 @@ def setup
 
 module ActiveSupport
   class TestCase
+    self.use_transactional_tests = false
```
ayumi: is this going to slow down tests by a bit?

mixonic: Non-scientific numbers: This run was 56 seconds. The run for https://github.com/brave-intl/publishers/pull/630 was 46. For https://github.com/brave-intl/publishers/pull/614 was 43.

If we wanted to speed things up I would look at the 30s rvm install time or disabling `sudo`.

But I care less about the specific strategy here and more about the `Chromedriver` which was trolling me, and I expect will start to troll others shortly.

mixonic:
@ayumi should be good to go 👍

PR #634, opened 2018-02-20
Twitch Support

mixonic:
This lands Twitch support behind a param. To run through the UI follow the instructions added to `README.md` then visit:

`/publishers/choose_new_channel_type?twitch=1`

And choose the Twitch option to move forward. This is a work-in-progress UI (not final UI, no translations), but it should be safe to land in staging.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,9 @@ gem "nokogiri", "~> 1.8.2"
 # Oauth client for google / youtube
 gem "omniauth-google-oauth2", "~> 0.5.2"
 
+# Oauth client for google / youtube
+gem "omniauth-twitch"
+
 # Model record auditing
 gem "paper_trail", "~> 5.2.2"
 
--- a/app/controllers/publishers/omniauth_callbacks_controller.rb
+++ b/app/controllers/publishers/omniauth_callbacks_controller.rb
@@ -48,7 +48,53 @@ module Publishers
 
       @current_channel.details = YoutubeChannelDetails.new(youtube_details_attrs)
 
-      @current_channel.save
+      @current_channel.save!
+
+      begin
+        PublisherChannelSetter.new(publisher: current_publisher).perform
+      rescue => e
+        # ToDo: What do we do if call to eyeshade fails
+        require "sentry-raven"
+        Raven.capture_exception(e)
+      end
+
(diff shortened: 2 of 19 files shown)


ayumi:
some merge conflicts from #614 i think

mixonic:
This is good to go 👍

Gemfile
```diff
@@ -30,6 +30,9 @@ gem "nokogiri", "~> 1.8.2"
 # Oauth client for google / youtube
 gem "omniauth-google-oauth2", "~> 0.5.2"
 
+# Oauth client for google / youtube
+gem "omniauth-twitch"
```
ayumi: minor: should we pin this?

app/controllers/publishers/omniauth_callbacks_controller.rb
```diff
@@ -48,7 +48,53 @@ def register_youtube_channel
...
+        else
+          redirect_to home_publishers_path, flash: { taken_channel_id: existing_channel.id }
+          return
+        end
+      end
+
+      @current_channel = Channel.new(publisher: current_publisher, verified: true)
+
+      twitch_details_attrs = {
+          twitch_channel_id: uid,
```
ayumi: minor: 2 sp vs 4 sp

app/models/twitch_channel_details.rb
```diff
@@ -0,0 +1,42 @@
...
+  validates :auth_user_id, presence: true
+  validates :display_name, presence: true
+
+  ## Begin methods to satisfy the Eyeshade integration
+
+  def channel_identifier
+    "twitch#channel:#{name}"
+  end
+
+  def authorizerEmail
```
ayumi: it's kind of inconsistent that we have snake_case and camelCase

app/models/youtube_channel_details.rb
```diff
@@ -9,10 +9,22 @@ class YoutubeChannelDetails < ApplicationRecord
   validates :thumbnail_url, presence: true
   validates :auth_user_id, presence: true
 
+  ## Begin methods to satisfy the Eyeshade integration
+
   def channel_identifier
     "youtube#channel:#{youtube_channel_id}"
   end
 
+  def authorizerEmail
```
ayumi: ditto the above -- a little funny that we're also using camelCase for methods

PR #646, opened 2018-02-22
Fix piwik by using window._paq

### LINKED ISSUE #636: Restore piwik/matomo stats
We lost piwik stats on Jan 12th, 2018 and we don't why.

## Leads
https://matomo.org/faq/troubleshooting/#faq_58
### Piwik/Matoma release broke the stats
Piwik (now Matomo) 3.3.0 was released on Jan 12th, 2018, this could be related  https://matomo.org/changelog/.

Update: We have not updated piwik on analytics.brave.com to use 3.3.0, and brave.com piwik stats were unaffected, so this is probably not the problem.  I haven't found any other piwik users who have reported this.

### We broke the stats with Jan 12th release
We did a release on Jan 12th, 2018 which could be related https://github.com/brave-intl/publishers/pull/482

Update: There is _nothing_ in this PR that touches JS or Piwik or the html which triggers them.

### We failed to include analytics.js with webpacker update
We could not be including it in the bundled JS which was released late on Jan 11th https://github.com/brave-intl/publishers/pull/470. 
> @w0ts0n I've just done a bit of spelunking and from what I understand. For piwik to work, you need to envoke this JS https://github.com/brave-intl/publishers/blob/master/app/javascript/analytics.js and I can't see that being called anywhere on https://publishers.basicattentiontoken.org/
@ayumi perhaps with the webpack migration we failed to include it in the bundled JS

Update:  analytics.js is being included in application.js.  I've checked the staging branch locally and analytics.js _is_ loading and executing on every page. 

### We dropped the config var on the server
Update: The `PIWIK_HOST` config var has always been correctly set.  On production:
```
irb(main):002:0> Rails.application.secrets[:piwik_host]
=> "https://analytics.brave.com/"
```

### Tracking code on analytics.brave.com does not match publishers app
Update: Nope, it does, it's 6.

### Also
For some reason, we still [get 1-3 actions recorded every few days](https://analytics.brave.com/index.php?module=CoreHome&action=index&idSite=6&period=range&date=last7&updated=1#?idSite=6&period=range&date=2018-01-09,2018-01-25&category=General_Visitors&subcategory=VisitorInterest_Engagement), and I cannot explain why.

cc @alexwykoff @ayumi @mrose17 @w0ts0n @mandar-brave

ayumi:
Fix #636

I tried to manually trigger a tracking event on prod with:

`_paq.push(['trackEvent', 'test', 'console', 42])`

This led to an error about a Piwik tracker not being configured.

I hypothesize that the migration to webpacker removed global scope from our JS files. This PR assigns `_paq` to `window` to be sure it's global. I've deployed it to staging2, and it fixed the above error with tracker configuration.

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/javascript/analytics.js
+++ b/app/javascript/analytics.js
@@ -1,12 +1,12 @@
 if (document.head.dataset.piwikHost) {
-  var _paq = _paq || [];
+  window._paq = window._paq || [];
   /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
-  _paq.push(['trackPageView']);
-  _paq.push(['enableLinkTracking']);
+  window._paq.push(['trackPageView']);
+  window._paq.push(['enableLinkTracking']);
   (function() {
     var u=document.head.dataset.piwikHost;
-    _paq.push(['setTrackerUrl', u+'piwik.php']);
-    _paq.push(['setSiteId', '6']);
+    window._paq.push(['setTrackerUrl', u+'piwik.php']);
+    window._paq.push(['setSiteId', '6']);
     var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
     g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
   })();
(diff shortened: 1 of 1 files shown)


mixonic:
@ayumi this makes sense to me. Good sleuthing 🔍

nvonpentz:
:clap:

PR #647, opened 2018-02-22
Referral promo fixes

### LINKED ISSUE #631: add DB constraint for promo id, referral code
we check uniqueness of promo referral codes at the app level, however we should also add a DB level constraint.
a unique multi-column index on promo_registrations, ["promo_id", "referral_code"] should do it

### LINKED ISSUE #642: Supply download links to promo assets in 'promo activated' emails
This was originally left out because we did not have the download links ready at the time of launch.

### LINKED ISSUE #645: Use https for underlying referral links in promo emails
The links are currently using the default`http`, we should change this to `https`.

cc @diracdeltas

nvonpentz:
Resolves #631 (add DB constraint for referral code, promo id)
Resolves #645 (use https for underlying links in promo emails)
Resolves #642 (supply download links to promo assets in promo emails) cc @jenn-rhim 

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/assets/stylesheets/mailer.scss
+++ b/app/assets/stylesheets/mailer.scss
@@ -194,6 +194,10 @@ body {
     .promo--ad-banner-youtube {
       float: left;
     }
+
+    .promo--download-link a{
+      float: right;
+    }
   }
 }
 
--- a/app/assets/stylesheets/mailer.scss
+++ b/app/assets/stylesheets/mailer.scss
@@ -195,7 +195,7 @@ body {
       float: left;
     }
 
-    .promo--download-link a{
+    .promo--download-link a {
       float: right;
     }
   }
--- a/app/views/promo_mailer/new_channel_registered_2018q1.html.slim
+++ b/app/views/promo_mailer/new_channel_registered_2018q1.html.slim
@@ -7,7 +7,7 @@ p = t("promo_mailer.promo_activated_2018q1_verified.body_one")
 
 a.promo--item-label.promo--item-label-link-name = @channel.details_type == "YoutubeChannelDetails" ? "#{@channel.publication_title.upcase} #{t("promo.shared.on_youtube")}" : "#{@channel.publication_title.upcase}"
 .promo--referral-link-container
-  .promo--referral-link = referral_url(@channel.promo_registration.referral_code)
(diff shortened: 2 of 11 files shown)


app/assets/stylesheets/mailer.scss
```diff
@@ -194,6 +194,10 @@ body {
     .promo--ad-banner-youtube {
       float: left;
     }
+
+    .promo--download-link a{
```
ayumi: minor: should have a space after `a`

app/views/promo_mailer/new_channel_registered_2018q1.html.slim
```diff
@@ -7,7 +7,7 @@ p = t("promo_mailer.promo_activated_2018q1_verified.body_one")
 
 a.promo--item-label.promo--item-label-link-name = @channel.details_type == "YoutubeChannelDetails" ? "#{@channel.publication_title.upcase} #{t("promo.shared.on_youtube")}" : "#{@channel.publication_title.upcase}"
 .promo--referral-link-container
-  .promo--referral-link = referral_url(@channel.promo_registration.referral_code)
+  .promo--referral-link = link_to(referral_url(@channel.promo_registration.referral_code), "https://" + referral_url(@channel.promo_registration.referral_code))
```
ayumi: this is a little hacky.. should we not fix referral_url() to return the full "https://..." url?

nvonpentz: this method is also used on the dashboard where we do not want to display.  I could write a new method or add an argument to this one if you prefer, but we do need to differentiate somehow.

ayumi: i see. the clean way would be to have the existing display method to be `#human_referral_url`, and the complete one with HTTPS:// to be `#referral_url`. consider also a string is not actually a url if it doesn't have the protocol, which is the existing behavior.

nvonpentz: cool, I will write a `#human_referral_url` and push the fix.  as far as strings not being urls, i'll have to think about that a bit more..

Issue #632, opened 2018-02-20
twitch channel support: allowing twitch publishers to receive Brave contributions 


jenn-rhim:
**Quick Summary**
The UX/UI for providing the ability for the user to verify their twitch channel is largely leveraged from the YT channel support as the verification process follows the same pattern: i.e. the third party (twitch) authorization of the user's channel ownership and sharing the channel meta data from our API call. 

**Key UI changes**
_Primarily 2 pieces for now. I'm sure we'll run into cases/errors we need to handle later..._

**1. Pick a channel** 
Allow our user to add & verify their twitch channel by providing the entry point: we're simply adding 'twitch' as the 3rd option on the 'Add Channel' dialog, prompted when the user clicks on Add Channel button from their dashboard.  (after which, the user goes through the screens served from twitch to authorize their channels before returning to our site). 
*note: the dialog is presented in an overlay format (compared to the full page format in the current version). 

![2-twitch-addch](https://user-images.githubusercontent.com/4369856/36451084-b83a91a0-1644-11e8-9bec-60c1d7e20340.png)
3rd box (on twitch) is in hover mode.

**2. Distinguish channels on the dashboard**
Indicate the channel type in the channel card on our dashboard once it's verified. The purpose of this is to distinguish channels that are hosted on different platforms but their attributes may be identical. 
*note: 1) by adding this, we're removing the attribution of the platform from the channel name on the card: i.e.   '[channel name] on YouTube'  --> it's simply '[channel name]' (see below)
2) unrelated but want to mention: channels are ordered reverse chronologically for now (newer on top).

![3-twitch-dash-added](https://user-images.githubusercontent.com/4369856/36451467-0727202a-1646-11e8-8af8-72741bdcbc06.png)

![image](https://user-images.githubusercontent.com/4369856/36451499-167bb554-1646-11e8-96c8-c888b3140071.png)

@mixonic @nvonpentz @ayumi @lgebhardt

ayumi:
:+1: 
for channel specific balances, I'm not sure we can get that info yet. cc @mrose17

mrose17:
@ayumi - we can use `GET /v2/publishers/{publisher}/balance` and `GET /v1/publishers/{publisher}/statement` !!

mixonic:
[...]

PR #653, opened 2018-02-23
Redesign the channel choosing workflow

mixonic:
References https://github.com/brave-intl/publishers/issues/632, specifically part 1 where the UI for choosing a channel is laid out.

A few things tech-level things are included in this PR that set it up for a good implementation:

* Fix code whitespace in `omniauth_callbacks_controller.rb`
* Add a new feature to modals, `modal.js`. They will add a class that includes the identifier to the `.md-container` wrapping element. See the code for documentation.
* Rubyize/Underscore `authorize_email` `authorize_name`
* Refactor `choose_new_channel_type.html.slim` to use bootstrap 4.

So on the actual feature front:

* Update the style of the `choose_new_channel_type` page. This becomes the no-JS fallback flow.
* Add a modal version from the `publishers/home` page that re-uses the same DOM. Because the home page has not been converted to bootstrap 4, there is a lot of CSS added just to make that re-used DOM work well in the legacy styles. Once the `publishers/home` page is converted to bootstrap 4 (not in this PR) the page should mostly Just Work (tm).

<details>
  <summary>Click to expand for screenshots</summary>

<b>A) 960px wide screen, with JS</b>
localhost_3000_publishers_home_twitch 1 1

**B) Pixel2, with JS**
localhost_3000_publishers_home_twitch 1 pixel 2 1

**C) 960px wide screen, no JS**
localhost_3000_publishers_choose_new_channel_type_twitch 1
**D) Pixel2, no JS**
localhost_3000_publishers_choose_new_channel_type_twitch 1 pixel 2
</details>
<br>

By default, these screens do **not** have Twitch included. In order to enable Twitch, you need to add `twitch=1` to the URL and reload the page. For example use these links in development:

* https://localhost:3000/publishers/home?twitch=1
* https://localhost:3000/publishers/choose_new_channel_type?twitch=1


app/models/twitch_channel_details.rb
```diff
@@ -15,11 +15,11 @@ def channel_identifier
     "twitch#channel:#{name}"
   end
 
-  def authorizerEmail
+  def authorize_email
```
ayumi: typo

app/views/publishers/_choose_channel_type.html.slim
```diff
@@ -0,0 +1,27 @@
+= link_to(new_site_channel_path, class: 'card block-link--card') do
```
ayumi: this link doesn't actually work for me. it dismisses the modal and prevents default

mixonic: Addressed. In https://github.com/brave-intl/publishers/commit/1c60fc7f8a877441ab48397f3c27cc5170906522#diff-fdc082bd84b5d4eea4238f264d49e385R91 the `preventDefault` call was moved to incorrectly be fired any time the user clicked anywhere that was not the background of the page. Instead, `preventDefault` and other closing logic should only be called if `js-confirm` or `js-deny` is clicked.

mixonic:
Rebased, also added a test for clicking on the website in the chooser.

PR #661, opened 2018-02-23
Redesign the channel choosing workflow v2

ayumi:
### This is #653 d95821f828e91ba78b0780b9b48f4c798b121668 with merge conflicts resolved against staging 1fd27ef14d626fd33afb858237c8c179f8e08bdf

@mixonic's original PR below from 2018-02-23 18:25Z:

References #632, specifically part 1 where the UI for choosing a channel is laid out.

A few things tech-level things are included in this PR that set it up for a good implementation:

Fix code whitespace in omniauth_callbacks_controller.rb
Add a new feature to modals, modal.js. They will add a class that includes the identifier to the .md-container wrapping element. See the code for documentation.
Rubyize/Underscore authorize_email authorize_name
Refactor choose_new_channel_type.html.slim to use bootstrap 4.
So on the actual feature front:

Update the style of the choose_new_channel_type page. This becomes the no-JS fallback flow.
Add a modal version from the publishers/home page that re-uses the same DOM. Because the home page has not been converted to bootstrap 4, there is a lot of CSS added just to make that re-used DOM work well in the legacy styles. Once the publishers/home page is converted to bootstrap 4 (not in this PR) the page should mostly Just Work (tm).
Click to expand for screenshots

By default, these screens do not have Twitch included. In order to enable Twitch, you need to add twitch=1 to the URL and reload the page. For example use these links in development:

https://localhost:3000/publishers/home?twitch=1
https://localhost:3000/publishers/choose_new_channel_type?twitch=1


ayumi:
i ran into this issue with the modal buttons not working (in Brave and Chrome), going to defer to Monday in the original PR

PR #665, opened 2018-02-25
Update username in topnav

### LINKED ISSUE #649: changing username doesn't dynamically change the username in the dropdown
When you change your username via the `Contact` section of the site, the dropdown name doesn't automatically change. Users need to refresh the website to see the change take affect. This isn't high priority, but would be nice if the name was automatically changed and didn't need a manual refresh.

STR:

* login onto Publishers
* once logged, change your name via `Edit Contact`
* change the name and select `Update`

You'll notice that the name under the menu isn't changed and requires a page refresh. We should probably change the name in the menu dynamically rather than requiring a refresh, example:

contactname

hahmed:
Update the username in the top nav upon updating user via the edit contact form at `/publishers/home`.

- [x] Fixes #649 

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/javascript/publishers/home.js
+++ b/app/javascript/publishers/home.js
@@ -184,7 +184,7 @@ document.addEventListener('DOMContentLoaded', function() {
         showPendingContactEmail(updatedEmail);
 
         let currentUserName = document.getElementById('current-user-name');
-        let userNameDropDown = document.getElementById('user-name-dd');
+        let userNameDropDown = document.getElementById('user-name-dropdown');
         currentUserName.innerText = updateContactName.value;
         userNameDropDown.innerText = updateContactName.value;
 
--- a/app/javascript/publishers/home.js
+++ b/app/javascript/publishers/home.js
@@ -183,6 +183,11 @@ document.addEventListener('DOMContentLoaded', function() {
         pendingContactEmail.innerText = updatedEmail;
         showPendingContactEmail(updatedEmail);
 
+        let currentUserName = document.getElementById('current-user-name');
+        let userNameDropDown = document.getElementById('user-name-dd');
+        currentUserName.innerText = updateContactName.value;
+        userNameDropDown.innerText = updateContactName.value;
+
         updateContactForm.classList.add('hidden');
         showContact.classList.remove('hidden');
         editContact.classList.remove('hidden');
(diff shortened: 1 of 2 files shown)


app/views/application/_nav.html.slim
```diff
@@ -8,13 +8,13 @@ nav.navbar.navbar-default.navbar-static-top.top-nav-collapse
...
               .user-avatar= render "avatar_svg"
-              span= current_publisher.name
+              span#current-user-name= current_publisher.name
               .user-lock= render two_factor_enabled?(current_publisher) ? "locked_svg" : "unlocked_svg"
               .user-arrow= render "arrow_svg"
               .user-dropdown
                 = link_to home_publishers_path, class: "user-info" do
                   .user-avatar= render "avatar_svg"
-                  .user-name= current_publisher.name
+                  .user-name#user-name-dd= current_publisher.name
```
ayumi: what is `dd`?

ayumi: ic it might be "dropdown", now i understand.
i think it'd be preferable to spell out "dropdown" to make it super clear for future devs.

hahmed:
Done

app/javascript/publishers/home.js
```diff
@@ -183,6 +183,11 @@ document.addEventListener('DOMContentLoaded', function() {
         pendingContactEmail.innerText = updatedEmail;
         showPendingContactEmail(updatedEmail);
 
+        let currentUserName = document.getElementById('current-user-name');
+        let userNameDropDown = document.getElementById('user-name-dropdown');
```
mixonic: Generally I suggest using classes of the style `js-user-name-dropdown`. When reading the templates this helps you identify where JS might rely on the DOM. Especially for un-tested behavior it can help prevent regressions.

hahmed: Will change that over to a class and prefix with `js-`. By the way, do you prefer adding parent scope `.user-menu` into the selector?

`let userNameDropDown = document.querySelector('.user-menu .js-user-name-dropdown');`

mixonic: Un-scoped class selectors are faster, and I don't think the scope adds much so I would leave it out 👍

Issue #664, opened 2018-02-23
Change email needs better error handling around duplicates


lgebhardt:
When an email address is changed from the dashboard or the new youtube login flow the pending email is not checked against the existing verified email addresses. This currently leads to an exception when validation fails as the email is being confirmed.

A two part fix is needed. First check for duplicates when the pending email is being entered. Second handle the validation error with an error message.

Issue #638, opened 2018-02-22
Allow bulk creation of Owners and Channels through the API


lgebhardt:
Support `POST /api/owners` and `POST /api/owners/:owner_id/channels` endpoints for eyeshade.

Issue #558, opened 2018-01-29
Changing email to email used by another account leads to error


ayumi:
From the dashboard, you can attempt to change your contact email to an email associated with another account. You receive a confirmation email, however when confirming it fails and shows a blank screen (500 error).

I don't think it's urgent as you can't change email due to Publisher model validation.
However we should figure out the best flow for handling this.

nvonpentz:
Potentially a duplicate of #664 

This has been resolved by #702

PR #666, opened 2018-02-25
Create owners via POST api/owners

hahmed:
Create owners via the api. This api assumes that you are creating owners, if a verified email exists or pending_email this will return a 422 `:unprocessable_entity`.

Related to #558 and #664. #669 resolves an issue with checking for verified/unverified emails 

- [x] #638

Reviewer Checklist:

Tests

Security:


hahmed:
@ayumi this is ready for review

ayumi:
thanks for your contribution! unfortunately, i don't think we can accept this PR right now. the spec in #638 isn't complete, and we need to discuss internally before we can proceed.

PR #668, opened 2018-02-26
Add Twitch support for the referral promo

### LINKED ISSUE #667: Add twitch support for the referral promo
This means
- updating the UI to include "on Twitch" in places where it read "on YouTube" for youtube channels on promo emails and views
- updating PromoRegistrar service with another method `twitch_request_body` that gives the information necessary for the promo server to register a twitch channel for the promo

nvonpentz:
Resolves #667 

* Add PromoRegistrar# twitch_request_body

* Add PromosHelper# on_channel_type to display 'on Twitch' or 'on YouTube' next to channel name

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/helpers/promos_helper.rb
+++ b/app/helpers/promos_helper.rb
@@ -58,4 +58,17 @@ module PromosHelper
   def base_human_referral_url
     Rails.application.secrets[:base_referral_url].to_s + "/"
   end
+
+  def on_channel_type(channel)
+    case channel.details_type
+    when "YoutubeChannelDetails"
+      "#{channel.publication_title.upcase} #{t("promo.shared.on_youtube")}"
+    when "TwitchChannelDetails"
+      "#{channel.publication_title.upcase} #{t("promo.shared.on_twitch")}"
+    when "SiteChannelDetails"
+      "#{channel.publication_title.upcase}"
+    else
+      raise
+    end
+  end
 end
--- a/app/services/promo_registrar.rb
+++ b/app/services/promo_registrar.rb
@@ -69,6 +69,8 @@ class PromoRegistrar < BaseApiClient
     case channel.details_type
     when "YoutubeChannelDetails"
       return youtube_request_body(channel)
+    when "TwitchChannelDetails"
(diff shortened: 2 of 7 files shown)


app/services/promo_registrar.rb
```diff
@@ -88,6 +90,18 @@ def youtube_request_body(channel)
...
 
+  def twitch_request_body(channel)
+    {
+      "owner_id": @publisher.id,
+      "promo": @promo_id,
+      "channel": channel.channel_id, 
+      "title": channel.publication_title,
+      "channel_type": "twitch",
+      "thumbnail_url": channel.details.thumbnail_url,
+      "description": nil
```
nvonpentz: we do not have a channel `description` for twitch channels like with youtube channels https://github.com/brave-intl/publishers/blob/1fd27ef14d626fd33afb858237c8c179f8e08bdf/db/schema.rb#L205, but the promo server takes a nullable one (like with youtube). for now we will always give `nil` for the description

app/helpers/promos_helper.rb
```diff
@@ -58,4 +58,17 @@ def facebook_url(referral_code)
   def base_human_referral_url
     Rails.application.secrets[:base_referral_url].to_s + "/"
   end
+
+  def on_channel_type(channel)
```
nvonpentz: There might be a better name for this method.  Eventually this should move the PublishersHelper, as this should also be displayed on the dashboard next to each channel name.

app/services/promo_registrar.rb
mixonic: @nvonpentz you could use `display_name`, which would be something like `MixonicSuperZapper312412` in the wild.

nvonpentz: @mixonic 
hmm, I think that would be the same as `channel.publication_title` which will be sent to the promo server as the `title`.  If that is correct, then I don't think we want to duplicate the title as the description.

ayumi:
let's make sure to verify on staging~

PR #670, opened 2018-02-26
Add channel selection popup from prompt

mixonic:
The channel selection popup was firing from the header, however it must also fire from the prompt when no channel has been added. This ensures it fires from both locations:

brave payments 2018-02-26 10-59-11


mixonic:
I think the build needs a kick here. 🤔

Issue #655, opened 2018-02-23
Rework flow for adding site channels


dgeb:
The current flow contains some elements leftover from before the multichannel refactor that don't make much sense now. I met with @jenn-rhim yesterday to discuss how to rework this flow into a more intuitive and cohesive experience.

We plan to make the following changes (referenced mocks appear below):

* The progress train needs to better align with the current steps. (a,b,c,d)
* Clicking "Cancel" at any step should remove the channel and return the user to the dashboard.
* Clicking "Verify" should return the user to the user to the dashboard instead of the "Now verifying your domain..." page (this page is removed entirely).
* If a domain can not be verified immediately, display a "Verifying your domain" modal over the dashboard. (e)
* While a domain is being verified, show the channel with a "Verification in progress..." message. (f)
* Perform async polling of domain verification status and notify the user with a flash message when there has been an update. (f)
* When a user has viewed the instruction page but has not clicked Cancel nor Verify, the domain should be considered "in progress" with the message "One more step to complete...". (f)
* When an error has been encountered during domain verification, display that error for the channel and provide actionable steps. These scenarios will be spelled out further in a separate issue.

-----

a) Enter domain info

![screen shot 2018-02-23 at 9 27 24 am](https://user-images.githubusercontent.com/29122/36599205-b06b4520-187c-11e8-9480-6f6985fca072.png)

b) Choose a verification method

![screen shot 2018-02-23 at 9 27 37 am](https://user-images.githubusercontent.com/29122/36599241-cf82d842-187c-11e8-9b47-7b2e43a4872e.png)

c) Trusted file instructions

![screen shot 2018-02-23 at 9 27 50 am](https://user-images.githubusercontent.com/29122/36599287-edcf14a0-187c-11e8-972b-29517f516a60.png)

d) Wordpress instructions

![screen shot 2018-02-23 at 9 28 00 am](https://user-images.githubusercontent.com/29122/36599298-f9341534-187c-11e8-88c7-6c142df7df39.png)

e) Dashboard with "Verifying your domain" modal

![screen shot 2018-02-23 at 9 29 31 am](https://user-images.githubusercontent.com/29122/36599376-3044c06e-187d-11e8-9f01-9200d4bf8e8b.png)

f) Dashboard with both a domain verification in progress, as well as a domain registration in progress

[...]

PR #674, opened 2018-02-27
Channel refactor

dgeb:
Refactors the flow for adding site channels as per #655.

I believe this is mergeable and an improvement over the status quo, but the following aspects are still TBD:

* When a channel is added, a flash alert is shown instead of the "Verifying your domain" modal that was spec'd by @jenn-rhim (see #655 - screenshot e). We still need the hourglass image to complete this modal.

* When a channel's verification status has started, it _is_ displayed in the listing as "Verification in progress". We still need the timer image spec'd by @jenn-rhim (see #655 - screenshot f).

* An endpoint for polling channel status has been configured and tested, but is not yet used from the dashboard. Thus, it is still necessary to refresh the page to get an updated status.

* The channel status needs to be sent from eyeshade which will display a more detailed error message than the default "Channel could not be verified!".

Here's a dashboard with some channels in the four different possible statuses (verified, started, failed, and incomplete):

screen shot 2018-02-27 at 9 38 41 am

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/controllers/site_channels_controller.rb
```diff
@@ -94,10 +94,11 @@ def check_for_https
     redirect_to(site_last_verification_method_path(@channel), alert: t(".alert"))
   end
 
-  # TODO: Rate limit
   def verify
     VerifySiteChannel.perform_later(channel_id: current_channel.id)
-    render(:verification_background)
+    current_channel.verification_started!
+
+    redirect_to(home_publishers_path, alert: t(".alert"))
```
ayumi: minor: alert/red is kind of intense, maybe this should be a gentler color

Issue #637, opened 2018-02-22
Duplicate site channels for the same owner and domain should be locked out


lgebhardt:
It's possible to create multiple unverified site channels for the same domain for the same publisher. This should produce an error and prevent the second channel from being created.

PR #678, opened 2018-02-27
[WIP] Do not allow dup channels for same owner and domain

hahmed:
Submitter Checklist:

- [ ] #637
- [ ] @ayumi 

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/models/site_channel_details.rb
+++ b/app/models/site_channel_details.rb
@@ -8,6 +8,8 @@ class SiteChannelDetails < ApplicationRecord
   # formats to support more publishers.
   validates :brave_publisher_id, uniqueness: { if: -> { !errors.include?(:brave_publisher_id_unnormalized) && brave_publisher_id.present? && brave_publisher_id_changed? && verified_publisher_id_exists? } }
 
+  validates :same_publisher_brave_publisher_id_unique, if: -> { !errors.include?(:brave_publisher_id_unnormalized) && brave_publisher_id.present? && brave_publisher_id_changed? }
+
   # - normalized and unnormalized domains
   # - normalized domains and domain-related errors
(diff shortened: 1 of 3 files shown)


app/models/site_channel_details.rb
```diff
@@ -109,4 +111,10 @@ def brave_publisher_id_not_changed_once_initialized
       errors.add(:brave_publisher_id, "can not change once initialized")
     end
   end
+
+  def same_publisher_brave_publisher_id_unique
+    if self.class.joins(:channel).where(brave_publisher_id: brave_publisher_id, "channels.publisher_id": channel.publisher_id).any?
```
yachtcaptain23: I believe `.where(brave_publisher_id: brave_publisher_id, "channels.publisher_id": channel.publisher_id)`

Can be re-written as `.where(brave_publisher_id: brave_publisher_id, channels: {publisher_id: channel.publisher_id})`

app/models/site_channel_details.rb
```diff
@@ -8,6 +8,8 @@ class SiteChannelDetails < ApplicationRecord
   # formats to support more publishers.
   validates :brave_publisher_id, uniqueness: { if: -> { !errors.include?(:brave_publisher_id_unnormalized) && brave_publisher_id.present? && brave_publisher_id_changed? && verified_publisher_id_exists? } }
 
+  validate :same_publisher_brave_publisher_id_unique, if: -> { !errors.include?(:brave_publisher_id_unnormalized) && brave_publisher_id.present? && brave_publisher_id_changed? }
```
yachtcaptain23: Uniqueness constraint should be done in database, not in Rails validation, as a multi-process system cannot guarantee uniqueness due to non-shared memory.

ayumi:
thanks again for your contribution! it's been a while since the last commit, so i'm going to close this. the original issue is something we'd like to get fixed sooner rather than later, so we're going to work on this internally.

PR #679, opened 2018-02-27
Show balance by channel on publishers/home

mixonic:
Closes https://github.com/brave-intl/publishers/issues/632. Previous PRs that addressed part of that ticket are https://github.com/brave-intl/publishers/pull/653 and https://github.com/brave-intl/publishers/pull/671.

Show the BAT balance for each verified channel.

* When the balance is unknown the `0.00` number would be replaced by the work `Unknown` in the same font/size.
* The promo screens are unchanged. I will confer with @nvonpentz on these.
* ~Introduces [Typhoeus](https://github.com/typhoeus/typhoeus) to allow the publisher wallet and channel requests to Eyeshade to run in parallel.~ I've removed typhoeus and moved to work to another branch. It requires some new testing infra (webmock doesn't support the easy path). WIP (works in dev) is at: https://github.com/mixonic/publishers/tree/show-balance-in-parallel
* Use a high-res image for the BAT logo on near the publisher wallet balance.

localhost_3000_publishers_home 2

TODO:

* [x] Confirm offline mode still works well
* [x] Add some testing
* [x] Confirm that "CURRENT PERIOD" is always the correct text on this page

### DIFF
--- a/app/assets/stylesheets/legacy/application/publishers.sass
+++ b/app/assets/stylesheets/legacy/application/publishers.sass
@@ -646,9 +646,10 @@ body[data-controller="publishers"]
 
         &.channel-verified
           .channel-status
-            background: url(asset-path("icn-verified.png")) no-repeat top left
-            padding: 3px 0 12px 38px
-            margin-top: -4px
+            background: url(asset-path("bat-logo@1x.png")) no-repeat top left
+            background-size: 30px 28.5px
+            padding: 1px 0 12px 38px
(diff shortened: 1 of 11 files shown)


ayumi:
i didn't get a chance to fully diligence this today, so i was thinking of deploying it soon after twitch

mixonic:
@ayumi sounds good. Doing some diligence and giving it time on staging seems like a smart move.

PR #688, opened 2018-02-28
Improve messaging when claiming a claimed Twitch account

mixonic:
Because Twitch offers no way for publishers to force the Twitch OAuth approval prompt to appear, we need to message to users that they may want to log out of their Twitch account to claim the correct account.

I confirmed Twitch does not honor the `prompt=consent` argument.

The improved messaging follows:

**User claims a Twitch account owned by someone else**

screen shot 2018-02-28 at 11 14 54 am

**User claims a Twitch account they have already claimed**

screen shot 2018-02-28 at 11 06 03 am

### DIFF
--- a/app/controllers/publishers/omniauth_callbacks_controller.rb
+++ b/app/controllers/publishers/omniauth_callbacks_controller.rb
@@ -72,7 +72,7 @@ module Publishers
 
       if existing_channel
         if existing_channel.publisher == current_publisher
-          redirect_to home_publishers_path, notice: t(".channel_already_registered")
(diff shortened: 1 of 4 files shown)


nvonpentz:
About the messaging:
> The following twitch channel has already been added to a different account

Will that appear if the twitch channel is owned by the same user that attempted to add it?

mixonic:
@nvonpentz no. In the case where a user has already claimed a channel they are attempting to claim, the second UI with an `info` blue message is shown.

ayumi:
i think this broke a test

https://travis-ci.org/brave-intl/publishers/builds/347450771

tried rerunning it twice

PR #705, opened 2018-03-09
Add Create API calls for Owners and Site Channels

### LINKED ISSUE #638: Allow bulk creation of Owners and Channels through the API

lgebhardt:
Implemented from slack conversations with @mrose17. @mrose17 should review before merging.

The `owner` `create` call creates verified Publishers and accepts `email`, `name`, and `phone` values. `phone` will be normalized. The created publisher will be email verified.

The `channel` `create` call creates a new site channel for the publisher specified in the route, and accepts the single `brave_publisher_id` value. After creating the new channel it is sent to eyeshade using the `SiteChannelDomainSetter`.

Payloads must be json.

Fixes #638

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/api/channels_controller.rb
+++ b/app/controllers/api/channels_controller.rb
@@ -1,9 +1,12 @@
 class Api::ChannelsController < Api::BaseController
   before_action :require_owner,
-                only: %i(verify notify show)
+                only: %i(verify notify show create)
 
   before_action :require_channel,
                 only: %i(verify notify show)
+
+  before_action :ensure_json_content_type,
+                only: %i(create)
 
   include PublishersHelper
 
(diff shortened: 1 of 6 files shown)


mrose17:
@nvonpentz - this looks good to me. if you concur, can you approve and merge? thanks!

app/controllers/api/owners_controller.rb
```diff
@@ -2,11 +2,28 @@ class Api::OwnersController < Api::BaseController
...
+    render(json: owner, status: :ok)
+  end
+
+  private
+
+  def owner_params
+    owner_params = params[:owner].permit(:email, :name, :phone)
+    owner_params[:visible] = ActiveRecord::Type::Boolean.new.deserialize(params[:owner].permit(:show_verification_status)[:show_verification_status])
+    owner_params
+  end
```
nvonpentz: should we also do `owner_params[:created_via_api] = true`?

app/controllers/api/channels_controller.rb
```diff
@@ -54,21 +57,41 @@ def verify
   end
 
   def show
-    render(json: @channel.details)
+    render(json: @channel.details, status: :ok)
+  end
+
+  def create
+    channel = Channel.new(publisher: @owner, verified: true)
```
nvonpentz: should we also set `created_via_api`?
```
channel = Channel.new(publisher: @owner, verified: true, created_via_api: true)
```

app/controllers/api/channels_controller.rb
```diff
@@ -54,21 +57,41 @@ def verify
...
+
+    channel.details = SiteChannelDetails.new(site_channel_details_params)
```
nvonpentz: why are we only supporting site channels? what's the context for this endpoint / when do we see ourselves using it?

mrose17: i have been waiting to use it since before the transition for one multi-site publisher (-;

lgebhardt:
@nvonpentz Good catch on the `created_via_api`. It's been fixed.

nvonpentz:
@mrose17 Let's test this on staging first if possible

PR #709, opened 2018-03-10
Prevent creating duplicate site channels for a publisher

### LINKED ISSUE #637: Duplicate site channels for the same owner and domain should be locked out

lgebhardt:
This adds a validation for a publisher trying to create a duplicate site channel to one already in process (verified channels were already prevented). The odd part of this PR is the validation is on the Channel, not on the SiteChannelDetails since it needs access to the Publisher which isn't available when the SiteChannelDetails haven't been saved.

Fixes #637 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -17,9 +17,15 @@ class Channel < ApplicationRecord
 
   accepts_nested_attributes_for :details
 
+  validates :publisher, presence: true
+
+  validates :details, presence: true
+
   validate :details_not_changed?
 
   validates :verification_status, inclusion: { in: %w(started failed) }, allow_nil: true
+
+  validate :site_channel_details_brave_publisher_id_unique_for_publisher, if: -> { details_type == 'SiteChannelDetails' }
 
   after_save :register_channel_for_promo, if: :should_register_channel_for_promo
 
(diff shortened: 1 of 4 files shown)


app/models/channel.rb
```diff
@@ -17,10 +17,16 @@ class Channel < ApplicationRecord
...
 
+  validates :publisher, presence: true
+
+  validates :details, presence: true
+
   validate :details_not_changed?
 
   validates :verification_status, inclusion: { in: %w(started failed) }, allow_nil: true
 
+  validate :site_channel_details_brave_publisher_id_unique_for_publisher, if: -> { details_type == 'SiteChannelDetails' }
```
ayumi: minor: it feels a little funny to have SiteChannelDetail specific validations on the general Channel

test/controllers/site_channels_controller_test.rb
```diff
@@ -93,6 +93,48 @@ class SiteChannelsControllerTest < ActionDispatch::IntegrationTest
...
+      last_channel = Channel.order(created_at: :asc).last
+      last_channel.details.verification_method = "wordpress"
+      last_channel.save!
+
+      refute_difference("Channel.count") do
+        post site_channels_url, params: create_params
+      end
+
+      assert_select('div#flash') do |element|
+        assert_match("newsite.org is already present.", element.text)
```
ayumi: minor: we could improve resiliency by using translation to get the string

config/locales/en.yml
```diff
@@ -449,6 +449,8 @@ en:
         verify: Make sure the verification file is placed in the folder and click verify.
         note: "Note: this process may take a few minutes to several hours depending on where your site is hosted."
       verify_button: Verify
+    create:
+      duplicate_channel: "%{domain} is already present."
```
ayumi: minor: It might be nice to explain what the user should do like

"Your account has a pending verification for %{domain} and you can continue it %{here}."

PR #723, opened 2018-03-15
Correctly add sentry debugging information for PublisherMailer # verify_email

nvonpentz:
Progresses towards #404 

As [anticipated](https://github.com/brave-intl/publishers/issues/404#issuecomment-353000312) by @ayumi, the debugging information I put in the `PublisherMailer#verify_email` method to address this [bug](https://sentry.io/brave-software/publishers-production-t2/issues/435856730/), does not work.  We're still getting this error so I've moved the publisher identifying information into the error message.

If there is a better way to do this, or better debugging information besides the `{publisher}.id` please comment.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/mailers/publisher_mailer.rb
+++ b/app/mailers/publisher_mailer.rb
@@ -71,11 +71,10 @@ class PublisherMailer < ApplicationMailer
 
     if @publisher.pending_email.blank?
       begin
-        raise "SMTP To address must not be blank for PublisherMailer#verify_email"
+        raise "SMTP To address must not be blank for PublisherMailer#verify_email for publisher #{@publisher.id}"
       rescue => e
         require 'sentry-raven'
-        Raven.capture_exception(e,
-                                publisher: @publisher)
+        Raven.capture_exception(e)
       end
     end
     mail(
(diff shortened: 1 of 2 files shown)


app/mailers/publisher_mailer.rb
```diff
@@ -71,11 +71,10 @@ def verify_email(publisher)
...
-        raise "SMTP To address must not be blank for PublisherMailer#verify_email"
+        raise "SMTP To address must not be blank for PublisherMailer#verify_email for publisher #{@publisher.id}"
       rescue => e
         require 'sentry-raven'
-        Raven.capture_exception(e,
-                                publisher: @publisher)
+        Raven.capture_exception(e)
       end
     end
     mail(
```
lgebhardt: Won't mail still be called after the exception is caught?

nvonpentz: Yes, should we eliminate the `mail` call in this case?

lgebhardt: I'd say yes. Throw it in an `else` block, and maybe swap the `if` logic so the error condition is the `else`.

PR #729, opened 2018-03-19
Use a shared emailed_auth_token.html.slim for post sign_up and log_in

### LINKED ISSUE #574: Need to fix the email sent confirmation page
- the user gets to this page when they try to sign up with an email that is already registered. since we already have the user's email, we go ahead and send them the login access link instead of routing them to the login page to perform the exact same task. 

1. given that we auto-reroute user to the login path, the alert type being error feels conflicting.
2. the message content is supposed to be already provided in the body, which is missing (see below): the only difference between sign up and log in is the wording referring to the right corresponding process.
3. this includes the paper plane image (i.e. missing)
4. (lower priority item): we have pages without the header nav bar. this page is one of them (all the pre-login pages).

image

email-sent-signup

nvonpentz:
Resolves #574 

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/assets/stylesheets/theme/panels.scss
+++ b/app/assets/stylesheets/theme/panels.scss
@@ -57,13 +57,6 @@ $brave-panels-borderRadius: 8px;
     justify-content: center;
     flex: 1;
 
-    &--email-sent {
-      @include media-breakpoint-up(md) {
-        background-image: url(asset-path("email-2@1x.png"));
-        background-position: top 59px right 60px;
-        background-repeat: no-repeat;
-      }
-    }
     &--platforms {
       padding-top: $spacer*2;
       padding-bottom: $spacer;
@@ -123,6 +116,16 @@ $brave-panels-borderRadius: 8px;
     padding: $spacer*4 $spacer*2;
     margin-top: auto;
     margin-bottom: auto;
+
+    &--email-sent {
+      padding-top: 125px;
+      margin-top: 30px;
+      @include media-breakpoint-up(md) {
+        background-image: url(asset-path("email-2@1x.png"));
+        background-position: center 20px;
+        background-repeat: no-repeat;
+      }
+    }
 
     align-items: center;
     justify-content: center;
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -72,7 +72,7 @@ class PublishersController < ApplicationController
     if verified_publisher
       @publisher = verified_publisher
       PublisherLoginLinkEmailer.new(email: email).perform
-      flash.now[:alert] = t(".email_already_active", email: email)
+      flash.now[:notice] = t(".email_already_active", email: email)
       render :emailed_auth_token
     elsif @publisher.save
       PublisherMailer.verify_email(@publisher).deliver_later
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -16,7 +16,7 @@ class PublishersController < ApplicationController
                new
                new_auth_token
                expired_auth_token
-               resend_email_verify_email)
+               resend_auth_email)
   before_action :require_unauthenticated_publisher,
     only: %i(sign_up
              create
@@ -56,6 +56,7 @@ class PublishersController < ApplicationController
     end
 
     @publisher = Publisher.new(pending_email: email)
+    @publisher_email = @publisher.pending_email
 
     @should_throttle = should_throttle_create? || params[:captcha]
     throttle_legit =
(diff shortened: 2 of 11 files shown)


app/controllers/publishers_controller.rb
```diff
@@ -189,9 +190,9 @@ def new_auth_token
   end
 
   def create_auth_token
-    email = publisher_create_auth_token_params[:email]
+    @publisher_email = publisher_create_auth_token_params[:email]
```
nvonpentz: `@publisher_email` used in emailed_auth_token.rb can be `publisher.email` or `publisher.pending_email` so it makes sense to set this here

app/helpers/publishers_helper.rb
```diff
@@ -418,4 +418,14 @@ def youtube_login_permitted?(channel)
   def publisher_created_through_youtube_auth?(publisher)
     publisher && publisher.channels.visible.count == 1 && youtube_login_permitted?(publisher.channels.visible.first)
   end
+
+  def emailed_auth_token_action
```
nvonpentz: The purpose of this method is to determine whether the text should say
"we've sent you a log in link"
or
"we've sent you a sign up link"

we could remove this method if didn't care about differentiating, and just said
"we've sent you an access link"

dgeb: I just tested this with both an existing and a new email address, and both displayed "We just sent a sign up link to ..." in emailed_auth_token.html.slim, despite receiving a login email for one and a signup email for the other. I'll investigate what's happening in this helper.

dgeb: Ah, this happened because the session was not cleared before performing the login and still contained `session[:created_publisher_id]`, which is what the helper is keying off of.

nvonpentz: Thanks.  I think this method may be more trouble than it's worth.

This method originally used `params[:action]` to determine what to display there.  If the action was `create_done` it used "log in" and if it was `create_auth_token` it used "sign up".  However, if they click "send again" at that point, the action that loads this view is `resend_email_verify_email`, and you can no longer differentiate.  That's why I tried grabbing the user from the session.

If you have a good way to choose whether to display "log in" vs "sign up" i'll use that.  Otherwise I'll remove this method and I'll just say "access link".

dgeb: @nvonpentz perhaps the mode could be passed in from the controller?

```
render(:emailed_auth_token, locals: { mode: :signup })
```

and 

```
render(:emailed_auth_token, locals: { mode: :login })
```

And then the helper could be passed the `mode` instead of relying on a session var?

nvonpentz: @dgeb Hmm, I think we would run in the same problem as checking the `params[:action]`.  That is, when we click "send email again" it will always display "sign up" since `resend_email_verify_email` redirects to `create_done` which will pass the `:sign_up` mode.

https://github.com/brave-intl/publishers/blob/aa69e380817f4e612511c8a59b94d7fbcc4b14a1/app/controllers/publishers_controller.rb#L100

dgeb: @nvonpentz I see. Yeah, the `mode` would have to be passed back in that "send email again" request to complete the cycle. Alternatively, you could clear `session[:created_publisher_id]` in `PublishersController#create_auth_token` before rendering and continue to use your current helper.

app/assets/stylesheets/theme/panels.scss
```diff
@@ -124,6 +117,16 @@ $brave-panels-borderRadius: 8px;
...
+    &--email-sent {
+      padding-top: 125px;
+      margin-top: 30px;
+      @include media-breakpoint-up(md) {
+        background-image: url(asset-path("email-2@1x.png"));
+        background-position: center 20px;
+        background-repeat: no-repeat;
+      }
+    }
+
```
nvonpentz: These styles were moved here because the flash messages in `emailed_auth_token.html.slim` were pushing the paper plane image under the "Email was sent text".  Now the image moves with the with text, so it will always stay above.

app/controllers/publishers_controller.rb
```diff
@@ -72,12 +73,13 @@ def create
...
       flash.now[:alert] = t(".email_already_active", email: email)
-      render :create_auth_token
+      params[:publisher_id] = @publisher.id
+      render :emailed_auth_token
     elsif @publisher.save
       PublisherMailer.verify_email(@publisher).deliver_later
       PublisherMailer.verify_email_internal(@publisher).deliver_later if PublisherMailer.should_send_internal_emails?
-      session[:created_publisher_id] = @publisher.id
-      redirect_to create_done_publishers_path
+      params[:publisher_id] = @publisher.id
```
nvonpentz: Now we rely on the publisher id being in the params to identify the publisher in PublishersController#resend_auth_email

dgeb: @nvonpentz It seems that the `emailed_auth_token` view relies on `@publisher.id` and not `params[:publisher_id]` to form the resend auth email path, so I don't think you need to set `params[:publisher_id]` here or on line 76 above.

nvonpentz: @dgeb You are correct, making changes

PR #730, opened 2018-03-19
Add "Complete Verification Later" button

### LINKED ISSUE #710: Adding Site Channel: No clear way to "pause" channel addition
Prior to #655 pressing cancel during adding a site channel would put channel into a state where the user could come back to it. They would see a "One more step to complete..." message and the "Let's Finish" button:

screen shot 2018-03-12 at 10 41 49 am

Now the `cancel` button deletes the channel and there isn't a clear way to get back to the dashboard to show the channel in the "One more step to complete..." state. Note, it is still possible by navigating to "/publishers/home", but there's no UI for this option.

There are legitimate use cases I can think of where a user is not prepared to click verify, but does not want to lose the verification token that results from pressing cancel. For example assigning the task of setting the token in the `.well-known` or dns TXT record to another person. Without a clear way to "pause" the process the user will be tempted to click "Cancel" and break the process.

I propose we add a new button such as "Continue later..." that takes them back to the dashboard. We could revert to the previous behavior (`cancel` does not delete the channel once a verification token was shown to the user), but this seems less clear and I suspect is why the change in #655 was made.

lgebhardt:
Fixes #710 

Link takes you back to the dashboard. When you resume you will be back to the previous page.

Wordpress:
screen shot 2018-03-19 at 4 36 14 pm

Hover state on link:

screen shot 2018-03-19 at 4 39 04 pm

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/assets/stylesheets/pages/site-channels.scss
+++ b/app/assets/stylesheets/pages/site-channels.scss
@@ -26,3 +26,13 @@
 .site-channels--padded-group {
   margin: $spacer*2 0;
 }
+
+.site-channels--finish-later {
+  padding-left: 40px;
+  color: $braveBrand;
+  &:hover,
+  &:focus {
+    color: $braveBrand-Dark;
+    text-decoration: none;
+  }
+}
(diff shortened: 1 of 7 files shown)


nvonpentz:
Works great! Thanks.

Issue #536, opened 2018-01-26
in-page notification/alert message needs to be event specific


jenn-rhim:
someone 'channel created' message is stuck, being displayed at the wrong time. see below the screenshot after removing a channel. should be 'Channel removed.' 

![screen shot 2018-01-25 at 4 34 03 pm](https://user-images.githubusercontent.com/4369856/35419774-cef21a90-01ee-11e8-9f96-5c04a4efbd4f.png)

ayumi:
Let's review after multi-chan to see if it changes

nvonpentz:
@jenn-rhim @ayumi I've never experienced this; is this still happening?

hahmed:
To reproduce:

Create a twitch account and the notification "Channel created" appears at the top.

*Do not refresh the page* 

Then click *Remove Channel* link next to the linked twitch account and the notification remains at the top of the page. 

Looks like the notification is persisting from whatever the last action, if you refresh the page - notification disappears.

nvonpentz:
Thanks @hahmed I'm able to reproduce it now.  Would you like to take this issue?

PR #740, opened 2018-03-21
Correct flash message on channel remove

hahmed:
Fixes https://github.com/brave-intl/publishers/issues/536

/cc @nvonpentz I pulled out the removed channel message from the locale then added to the element below, so I can set the flash message correctly inside the javascript. 

`.row.channel-row id=("channel_row_#{channel.id}") data-remove-message= t("shared.channel_removed")`

https://github.com/brave-intl/publishers/blob/8109d1ed6780a6c14a2f1cc674683de3244e3318/app/views/publishers/home.html.slim#L144

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/javascript/publishers/home.js
+++ b/app/javascript/publishers/home.js
@@ -4,6 +4,7 @@ import {
   submitForm
 } from '../utils/request';
 import dynamicEllipsis from '../utils/dynamicEllipsis';
+import flash from '../utils/flash';
 
 function showPendingContactEmail(pendingEmail) {
   let pendingEmailNotice = document.getElementById('pending_email_notice');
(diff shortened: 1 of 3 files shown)


app/javascript/publishers/home.js
```diff
@@ -51,6 +52,8 @@ function removeChannel(channelId) {
         let addChannelPlaceholder = document.getElementById("add_channel_placeholder");
         addChannelPlaceholder.classList.remove("hidden");
       }
+      flash.clear();
+      flash.append('warning', channelRow.getAttribute('data-remove-message'));
```
nvonpentz: Can we change this to use 'info' / blue?

```js
flash.append('info', channelRow.getAttribute('data-remove-message'));
```

PR #753, opened 2018-03-28
WIP - Add MailChimpRegistrar service

lgebhardt:
WIP - This PR was made for discussion. I'd like to get the Rake Task in as well so we can initialize the mailchimp account.

There are a few things I'd like to change, but looking for feedback:

 The main one is the configuration for the mailchimp api. We're using a single list, and Publishers will track interests under it's own category. Currently the ids for the list and category need to be specified. It's easy to get them with an api call, but not from the MailChimp UI. And this is an additional config step.

Failures communicating with the MailChimp api do not result in the service trying again. These are only logged.

Fixes ##738

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,9 @@ gem "email_validator", "~> 1.6"
 # HTTP library wrapper
 gem "faraday", "~> 0.9.2", require: false
 
+# MailChimp API Wrapper
+gem "gibbon"
+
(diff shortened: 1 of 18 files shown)


lgebhardt:
Moving to Sendgrid instead

PR #754, opened 2018-03-28
Add max_site_age envar

### LINKED ISSUE #752: Some valid channels are never enqueued for verification
We attempt to verify site channels every hour, however, if the channel is older than the `MAX_AGE` hardcoded in the `EnqueueSiteChannelVerifications` job, they will never be enqueued.
https://github.com/brave-intl/publishers/blob/b216b8d344bc74cc500372f2df8a3a68d325e9b5/app/jobs/enqueue_site_channel_verifications.rb#L3
If there is a bug that prevents an otherwise valid verification attempt from being processed for longer than `MAX_AGE`, these channels will never be verified.  This is precisely what happened with the recent bout of verification problems.

We should change `EnqueueSiteChannelVerifications` to depend on an environment variable instead of being hardcoded.  This way we can enqueue old (but valid) site verifications by adjusting the environment variable and not the code.

nvonpentz:
Resolves #752 

Open to other name suggestions!

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -66,6 +66,7 @@ default: &default
   api_promo_key: <%= ENV["API_PROMO_KEY"] %>
   active_promo_id: <%= ENV["ACTIVE_PROMO_ID"] %>
   base_referral_url: <%= ENV["BASE_REFERRAL_URL"] %>
+  max_site_age: <%= ENV["MAX_SITE_AGE"] %> # Maximimum age in weeks of sites enqueued for verification
 
 development:
   <<: *default
@@ -84,6 +85,7 @@ development:
   bat_twitter_url: "https://twitter.com/@attentiontoken"
   bat_reddit_url: "https://www.reddit.com/r/BATProject/"
   bat_rocketchat_url: "https://basicattentiontoken.rocket.chat/"
+  max_site_age: 6
 
 test:
   <<: *default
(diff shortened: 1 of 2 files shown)


config/secrets.yml
```diff
@@ -84,6 +85,7 @@ development:
   bat_twitter_url: "https://twitter.com/@attentiontoken"
   bat_reddit_url: "https://www.reddit.com/r/BATProject/"
   bat_rocketchat_url: "https://basicattentiontoken.rocket.chat/"
+  max_site_age: 6
```
lgebhardt: Needs to be `6.weeks`

config/secrets.yml
```diff
@@ -108,6 +110,7 @@ test:
...
```
lgebhardt: Needs to be `6.weeks`

nvonpentz:
@lgebhardt Thanks for taking a look.  I do specify `weeks` in when `MAX_AGE` is set in the job, but I could  instead specify it in the environment variable.  What do you think?

lgebhardt:
@nvonpentz Sorry, I missed that. Ignore my comments - it's right the way it is.

nvonpentz:
@lgebhardt I see no problem with being limited to weeks.

PR #759, opened 2018-03-30
Feature/modal verification failed

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/mailers/publisher_mailer.rb
+++ b/app/mailers/publisher_mailer.rb
@@ -68,7 +68,7 @@ class PublisherMailer < ApplicationMailer
   def verify_email(publisher)
     @publisher = publisher
     @private_reauth_url = generate_publisher_private_reauth_url(@publisher)
-    
+
     if @publisher.pending_email.present?
       mail(
           to: @publisher.pending_email,
@@ -83,7 +83,7 @@ class PublisherMailer < ApplicationMailer
       end
     end
   end
-  
+
   # TODO: Refactor
   # Like the above but without the verify_email link
   def verify_email_internal(publisher)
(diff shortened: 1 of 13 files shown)


app/models/channel.rb
```diff
@@ -98,15 +109,20 @@ def promo_enabled?
...
+    update!(verified: false, verification_status: STARTED, verification_details: nil)
   end
 
   def verification_failed!(details = nil)
     # Clear changes so we don't bypass validations when saving without checking them
     self.reload
 
+    if manual_verification_running
+      PublisherMailer.verification_failed(self).deliver_later
+      self.update(manual_verification_running: false)
```
ayumi: minor: `#save!` runs below so `#update` could be replaced with a simple assignment.

app/mailers/publisher_mailer.rb
```diff
@@ -163,6 +163,16 @@ def verified_no_wallet_internal(publisher, params)
     )
   end
 
+  def verification_failed(channel)
+    @publisher = channel.publisher
+    @channel = channel
+    mail(
+      to: @publisher.email,
+      subject: "#{channel.details.publication_title} Verification Failed",
```
ayumi: `subject: default_i18n_subject`

test/features/publishers_home_test.rb
```diff
@@ -65,13 +65,37 @@ class PublishersHomeTest < Capybara::Rails::TestCase
...
+    publisher = publishers(:small_media_group)
+    channel = channels(:small_media_group_to_verify)
+
+    sign_in publisher
+    visit home_publishers_path
+    assert_content page, "Verification Failed"
+
+    # Close out of partial
+    find("#modal-close-button").click
+    assert_no_content page, "Verification Failed"
```
ayumi: it would also be good to confirm the modal is dismissed permanently, i.e. `#disable_verification_failed_modal` runs correctly and `#shown_verification_failed_modal` is updated.

db/migrate/20180403031030_add_shown_verification_failed_modal_to_channels.rb
```diff
@@ -0,0 +1,6 @@
+class AddShownVerificationFailedModalToChannels < ActiveRecord::Migration[5.0]
+  def change
+    add_column :channels, :shown_verification_failed_modal, :boolean, default: false
```
ayumi: minor: consider adding a `null: false` constraint

Also this manual failure case may only apply to Site channels, so it could be cleaner to move these two flags onto [SiteChannelDetails](https://github.com/brave-intl/publishers/blob/staging/app/models/site_channel_details.rb).

PR #764, opened 2018-04-03
Use Eyeshade naming conventions for promo stuff

### LINKED ISSUE #691: Remove {channel}.channel_id and use {channel}.details.channel_identifier in PromoRegistrar
As the discussion in #673, we have two methods that return a channel's identifier:

`{channel}.channel_id`
`{channel}.details.channel_identifier`

The first method was created in error when the initial referral promo code was written, when I didn't realize second method was available. The only difference is that `{channel}.channel_id `does not include the `youtube#channel` or `twitch#channel:` prefix. ex:
```
irb(main):005:0> YoutubeChannelDetails.first.channel.channel_id
=> "UCXmw_D6tX85F0y3DcDws5Hg"
irb(main):006:0> YoutubeChannelDetails.first.channel_identifier
=> "youtube#channel:UCXmw_D6tX85F0y3DcDws5Hg"
```

We should remove `{channel}.channel_id` and use `{channel}.details.channel_identifier` instead in the `PromoRegistrar`.  However, since this method has been already used in production with promo services, we will need to update the promo services db to reflect this change for channels that have already been registered for the promo.

nvonpentz:
Resolves #691 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/services/promo_registration_getter.rb
+++ b/app/services/promo_registration_getter.rb
@@ -24,7 +24,7 @@ class PromoRegistrationGetter < BaseApiClient
     referral_code = referral_code_for_promo_id(registrations)
     
     if should_update_channel_owner_on_promo_server(registrations)
-      PromoChannelOwnerUpdater.new(publisher_id: @publisher.owner_identifier, referral_code: referral_code).perform
+      PromoChannelOwnerUpdater.new(publisher: @publisher, referral_code: referral_code).perform
     end
     
     referral_code
--- a/app/services/promo_registration_getter.rb
+++ b/app/services/promo_registration_getter.rb
@@ -18,13 +18,13 @@ class PromoRegistrationGetter < BaseApiClient
     response = connection.get do |request|
       request.headers["Authorization"] = api_authorization_header
       request.headers["Content-Type"] = "application/json"
-      request.url("/api/1/promo/publishers?channel=#{@channel.channel_id}")
+      request.url("/api/1/promo/publishers?channel=#{@channel.details.channel_identifier}")
     end
     registrations =  JSON.parse(response.body)
     referral_code = referral_code_for_promo_id(registrations)
     
     if should_update_channel_owner_on_promo_server(registrations)
-      PromoChannelOwnerUpdater.new(publisher_id: @publisher.id, referral_code: referral_code).perform
+      PromoChannelOwnerUpdater.new(publisher_id: @publisher.owner_id, referral_code: referral_code).perform
     end
     
     referral_code
@@ -43,13 +43,13 @@ class PromoRegistrationGetter < BaseApiClient
       {
       "referral_code" => offline_referral_code,
       "promo" => "#{@promo_id}",
-      "publisher" => "#{@channel.channel_id}",
+      "publisher" => "#{@channel.details.channel_identifier}",
       "name" => "#{@channel.publication_title}",
       },
       {
       "referral_code" => offline_referral_code,
       "promo" => "free-bats-2018q2",
-      "publisher" => "#{@channel.channel_id}",
+      "publisher" => "#{@channel.details.channel_identifier}",
       "name" => "#{@channel.publication_title}",
       }
     ]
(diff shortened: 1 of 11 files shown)


app/services/promo_registration_getter.rb
```diff
@@ -18,13 +18,13 @@ def perform
...
       request.headers["Content-Type"] = "application/json"
-      request.url("/api/1/promo/publishers?channel=#{@channel.channel_id}")
+      request.url("/api/1/promo/publishers?channel=#{@channel.details.channel_identifier}")
     end
     registrations =  JSON.parse(response.body)
     referral_code = referral_code_for_promo_id(registrations)
     
     if should_update_channel_owner_on_promo_server(registrations)
-      PromoChannelOwnerUpdater.new(publisher_id: @publisher.id, referral_code: referral_code).perform
+      PromoChannelOwnerUpdater.new(publisher_id: @publisher.owner_id, referral_code: referral_code).perform
```
dgeb: Should this be `@publisher.owner_identifier`? Or maybe we should pass `@publisher.id` and then find the corresponding `@publisher` in `PromoChannelOwnerUpdater`.

Looking at `PromoChannelOwnerUpdater`, I see we're setting `@publisher_id` to this value. However, in some places `@publisher.id` is also referenced, without instantiating `@publisher`.

app/services/publisher_promo_stats_fetcher.rb
```diff
@@ -13,7 +13,7 @@ def perform
     response = connection.get do |request|
       request.headers["Authorization"] = api_authorization_header
       request.headers["Content-Type"] = "application/json"
-      request.url("/api/1/promo/owners/#{@publisher.id}/statsByTime?promo_id=#{@promo_id}")
+      request.url("/api/1/promo/owners/#{@publisher.owner_identifier}/statsByTime?promo_id=#{@promo_id}")
```
dgeb: Since `owner_identifier` is of the format `"publishers#uuid:#{id}"`, won't the hash (`#`) cause an issue in a URL?

app/services/promo_registration_getter.rb
nvonpentz: @dgeb Good catch, you are correct on both counts.

app/services/publisher_promo_stats_fetcher.rb
dgeb: @nvonpentz yeah, using `URI.escape` is the right answer if the full owner identifier is expected here.

nvonpentz: @dgeb Thanks!  I will update here as well as the other places I use `#{channel.details.channel_identifier}`

test/tasks/launch_promo_test.rb
```diff
@@ -8,30 +8,15 @@ class LaunchPromoTest < ActiveJob::TestCase
...
-  #   #     Rake::Task["promo:launch_promo"].invoke
-  #   #   end
-  #   # end
+  # test "generates a promo token and sends email to each publisher" do
+  #   assert_difference("Publisher.where.not(promo_token_2018q1: nil).count", Publisher.where.not(email: nil).count) do
+  #     assert_enqueued_jobs(Publisher.where.not(email: nil).count) do
+  #       Rake::Task["promo:launch_promo"].invoke
+  #       Rake::Task["promo:launch_promo"].reenable
+  #     end
+  #   end
```
nvonpentz: This test fails no matter what I do.  We're not launching another promo any time soon, and we'd have to revisit this rake task anyway if/when we do.

nvonpentz:
tests pass locally 🤔

nvonpentz:
the problem was that I left defined the auth token in the `stub_request`, which must be different then the on the travis build.

nvonpentz:
Closing because this branch is stale, and this functionality will be encapsulated in a later PR.

PR #772, opened 2018-04-07
Push default currency changes directly to eyeshade

### LINKED ISSUE #767: Tell eyeshade that a user has updated their default currency
We should use the PublisherWalletSetter `PUT /v1/owners/{owner}/wallet` to update eyeshade when a new currency is selected.  Currently it’s only called once, right after they create their uphold wallet.

cc @evq

dgeb:
When the `default_currency` is successfully updated, a new `UploadDefaultCurrencyJob` will be triggered that will upload the change to eyeshade.

The UX has been tweaked slightly so that, for users with no default currency, a select option `— Select currency —` will appear as selected by default. This allows any other selection to trigger a change (since there’s no default chosen as was previously the case).

screen shot 2018-04-06 at 5 09 29 pm

Fixes #767 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/helpers/publishers_helper.rb
+++ b/app/helpers/publishers_helper.rb
@@ -47,8 +47,8 @@ module PublishersHelper
   end
 
   def publisher_converted_balance(publisher)
-    currency = publisher_default_currency(publisher)
-    return if currency == "BAT"
+    currency = publisher.default_currency
+    return if currency == "BAT" || currency.blank?
     if balance = publisher.wallet && publisher.wallet.contribution_balance
       converted_amount = '%.2f' % balance.convert_to(currency)
       I18n.t("helpers.publisher.balance_pending_approximate", amount: converted_amount, code: currency)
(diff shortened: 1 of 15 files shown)


dgeb:
I'll investigate the Travis failure tomorrow.

app/helpers/publishers_helper.rb
```diff
@@ -111,13 +111,15 @@ def terms_of_service_url
...
 
-  def publisher_default_currency(publisher)
-    publisher.default_currency.present? ? publisher.default_currency : 'BAT'
-  end
-
   def publisher_available_currencies(publisher)
     available_currencies = publisher.wallet.try(:wallet_details).try(:[], 'availableCurrencies')
-    available_currencies.blank? ? ['BAT'] : available_currencies
+    if available_currencies.blank?
+      available_currencies = ['BAT']
```
dgeb: @nvonpentz I think this default may cause more problems than it solves, now that I understand that not all publishers can choose `BAT`. It's probably better to leave the options blank and have `default_currency` remain `nil` until publishers can choose from actual available currencies pulled from eyeshade.

PR #773, opened 2018-04-09
Rate limit attempts to resend emails 

### LINKED ISSUE #755: We do not force a captcha when spamming resend login & sign up email
We force a captcha for > 3 attempts to create a new publisher in `PublishersController#create`
https://github.com/brave-intl/publishers/blob/dae4643d24214c480aa45f86a6db999d5e095aa1/app/controllers/publishers_controller.rb#L60

and for > 3 attempts to log in in `PublishersController#create_auth_token`
https://github.com/brave-intl/publishers/blob/dae4643d24214c480aa45f86a6db999d5e095aa1/app/controllers/publishers_controller.rb#L200

but we don't when users spam the resend login/sign up link in `Publishers#resend_auth_email`.  We should add similar constraints here.

We may want to take this opportunity to refactor how these emails are sent.  Do we want 3 separate controller actions to send emails?  Do we need `PublisherLoginLinkEmailer`?  Should we be changing application state in the helper method `PublishersHelper#generate_publisher_private_reauth_url`?

nvonpentz:
Resolves #755 

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -95,7 +95,7 @@ class PublishersController < ApplicationController
   def resend_auth_email    
     @publisher = Publisher.find(params[:publisher_id])
 
-    @should_throttle = should_throttle_create_auth_token?
+    @should_throttle = should_throttle_create_auth_token? || params[:captcha]
     throttle_legit =
       @should_throttle ?
         verify_recaptcha(model: @publisher)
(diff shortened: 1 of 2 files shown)


app/controllers/publishers_controller.rb
```diff
@@ -94,7 +94,17 @@ def create_done
 
   def resend_auth_email    
     @publisher = Publisher.find(params[:publisher_id])
-    
+
+    @should_throttle = should_throttle_create_auth_token?
```
dgeb: Not a blocker - but let's consider adding ` || params[:captcha]` for consistency and easier testing.

PR #788, opened 2018-04-17
Allow Uphold account to be disconnected / changed

### LINKED ISSUE #133: Allow Uphold accounts to be disconnected / changed
After an Uphold account has been associated with a publisher, it seems desirable to allow a publisher to switch to using a different Uphold account.

Similarly, it may also be desirable to allow publishers to disconnect from Uphold entirely for a time, thereby allowing balances to accrue without being transferred.

/cc @alexwykoff @mrose17

dgeb:
This introduces a new UX and new capabilities to the Uphold panel on the dashboard.

A status message and status change link now always appear at the top of the panel.

When connected, the user now has the option to disconnect:

screen shot 2018-04-17 at 9 41 58 am

This link will raise the following confirmation prompt:

screen shot 2018-04-17 at 9 42 10 am

If the user chooses to disconnect, Uphold will show as "Not Connected", and eyeshade will be notified:

screen shot 2018-04-17 at 9 42 24 am

If the user then reconnects to Uphold, the "Connecting" status will pulse until a connection has been confirmed:

screen shot 2018-04-17 at 9 43 02 am

If communication problems occur and the browser is unable to get an updated status, the following timeout message will eventually appear: 

screen shot 2018-04-17 at 9 40 07 am

Closes #133 

/cc @nvonpentz @jenn-rhim 

Submitter Checklist:

- [X] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/assets/stylesheets/legacy/application/publishers.sass
+++ b/app/assets/stylesheets/legacy/application/publishers.sass
@@ -506,18 +506,6 @@ body[data-controller="u2f_registrations"]
           margin-bottom: 40px
       .attribute-value
         margin-bottom: 24px
-
-      #publisher_verification_status
-        font-size: $font-size-base
-        vertical-align: middle
-        padding: 5px 0 5px 35px
-        background-position: center left
-        background-repeat: no-repeat
-        background-size: 25px
-        &.verified
-          background-image: url(asset-path("verified-icon@1x.png"))
-        &.unverified
-          background-image: url(asset-path("unverified-icon.png"))
       #publisher_status
         vertical-align: top
         margin-top: 20px
--- a/app/assets/stylesheets/legacy/application/publishers.sass
+++ b/app/assets/stylesheets/legacy/application/publishers.sass
@@ -1,5 +1,9 @@
 $subnav-active: #8F9397
 
+@keyframes blinker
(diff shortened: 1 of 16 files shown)


nvonpentz:
This looks solid.  

* @jenn-rhim Can you confirm that having two "Reconnect to Uphold" buttons (one small, one large) is the intended behavior?

* If a publisher has never created a uphold wallet, their dashboard will say "unconnected".  However if a publisher has been uphold connected any time in the past and they either click "disconnect" or deauthorize us from uphold.com, their dashboard will always read "connection problems...we are experiencing communication problems with uphold".  There might not be any communication problems here to begin with, a publisher might have just wanted to disconnect.  Should we always be displaying "connection problems" if a publisher disconnects?

jenn-rhim:
- yes
- when the user disconnects their uphold explicitly, we display ‘Not Conncted’ (unconnected sounds awkward to me). 
- Connection problem status is when we are not getting the proper response from eyeshade that the user’s uphold account is not connected: eyeshade may be having an issue, or other connection issues that may result in us not knowing the exact status. i.e. User’s uphold account connection may be good.

PR #810, opened 2018-04-26
Email token handling

### LINKED ISSUE #733: 'Your login link has expired' on freshly generated login links
> I'm trying to verify my email and when I receive the link. The Brave payments webpage says my account is already verified. Then when I try to login sometimes I don't receive the emails to log in and when I get them the Brave payments accounts says the link is already expired.

I've been experiencing this on development for a while now.

https://github.com/brave-intl/publishers/pull/440 is possibly related.

There could be a race condition somewhere in the process of sending an email where we aren't generating an `authentication_token` fast enough to be included in the email. 

cc @jenn-rhim

lgebhardt:
* Reworks login token handling to keep token generation synchronous
* Emails that send login tokens are now sent from MailerServices that generate the required token, and also handle sending internal and confirmation emails
* Some test cleanup was done
* Unused Welcome emails were removed
* Ensure that a publisher exists for twitch channel authentication
* When creating a new publisher attempt to reuse previous unverified publishers with the same pending email

The existing tests cover the functionality of the MailerServices, but I'd like to eventually create unit tests for these. I think that can be deferred, but let me know if we need them to merge.

This may fix #733, but since we can't reliably reproduce it it also may not. It should however remove one potential cause which is asynchronous writing of the publisher record by the mailers.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


test/mailers/previews/publisher_mailer_preview.rb
```diff
@@ -18,14 +18,6 @@ def verification_done_internal
...
   end
 
-  def welcome
-    PublisherMailer.welcome(Publisher.first)
-  end
-
-  def welcome_internal
-    PublisherMailer.welcome_internal(Publisher.first)
-  end
-
```
nvonpentz: (Outside the scope of this PR) I noticed that a lot of methods are missing from the `PublisherMailerPreview` and some like `verification_done` raise an error.  I will probably open a separate issue to address this.

app/controllers/publishers_controller.rb
```diff
@@ -72,12 +72,11 @@ def create
     verified_publisher = Publisher.by_email_case_insensitive(email).first
     if verified_publisher
```
nvonpentz:
```ruby
 @publisher = Publisher.find_or_create_by(pending_email: email, email: nil)
```
In this action we first look to see if there is a valid publisher with the correct `pending_email` that also has `email: nil`.  If there is a such a publisher we assign `@publisher` to it.  If not, we assign `@publisher` to a new publisher.

Then we check if there is a valid publisher with that `email`.  If that's the case we overwrite `@publisher` to be the publisher with the valid `email`.

Shouldn't we be checking if there exists a valid publisher with the `email` first, and only run `Publisher.find_or_create_by(pending_email: email, email: nil)` if that's not the case?

app/controllers/publishers_controller.rb
```diff
@@ -219,7 +212,6 @@ def create_auth_token
       return redirect_to new_auth_token_publishers_path
     end
 
-    @publisher = Publisher.new(publisher_create_auth_token_params)
     @should_throttle = should_throttle_create_auth_token? || params[:captcha].present?
```
nvonpentz: Should we move `|| params[:captcha].present?` into `should_throttle_create_auth_token?` here and throughout?

app/controllers/publishers_controller.rb
```diff
@@ -56,7 +56,7 @@ def create
       return redirect_to sign_up_publishers_path
     end
 
-    @publisher = Publisher.new(pending_email: email)
+    @publisher = Publisher.find_or_create_by(pending_email: email, email: nil)
```
nvonpentz: Can we add a comment explaining why we're not just doing
```ruby
Publisher.find_or_create_by(pending_email: email)
```
Something along the lines of
```
    # Requiring `email: nil` ensures we do not select a publisher with the same pending_email
    # as a publisher in the middle of the change email flow
```

app/mailers/publisher_mailer.rb
```diff
@@ -2,19 +2,27 @@ class PublisherMailer < ApplicationMailer
   include PublishersHelper
   add_template_helper(PublishersHelper)
 
-  def login_email(publisher)
+  # Note: token_fresh is a reminder that you must refresh the token before sending this email.
+  # Best practice is to use the MailerServices::PublisherLoginLinkEmailer service
+  def login_email(publisher, token_fresh = false)
+    raise "token not fresh" unless token_fresh
```
nvonpentz: I like how this restricts the the use of token generation to only the MailerServices .  What do you think of also checking that the token exists and hasn't expired to be extra sure?

```ruby
raise if publisher.authentication_token == nil || publisher.authentication_token_expires_at <= Time.now
```

app/controllers/publishers_controller.rb
yachtcaptain23: This might cause a double create in the future. Since the values for `pending_email` and `email` are the same, shouldn't we have a flag on whether or not the email is verified? Let's talk about it more in person.

app/mailers/publisher_mailer.rb
nvonpentz: I'll add the before filters and remove `token_fresh`.

nvonpentz:
Sorry for notifications on this one.  I'm having trouble getting travis to accept this, but tests do pass locally.

The issue is with the change I made to the `completed` publisher fixture.  I changed their default `authentication_token` value from `5.days.from_now` to `PublisherTokenGenerator::TOKEN__TTL.from_now`.

Now any test that involves signing in the `completed` publisher is failing on travis because the it thinks the token is expired.  I don't know why that's the case.

https://github.com/lgebhardt/publishers/blob/64a478189a9f4667e52a574afdd9e364a0cad14c/test/fixtures/publishers.yml#L29

PR #813, opened 2018-05-01
Provide default UPHOLD_AUTHORIZATION_ENDPOINT

### LINKED ISSUE #608: set default UPHOLD_AUTHORIZATION_ENDPOINT to facilitate development
This is the remaining env var so the app works out of the box, without README Step 8 (sourcing an additional dev file with env vars).
We should update the README accordingly.

UPHOLD_AUTHORIZATION_ENDPOINT="https://sandbox.uphold.com/authorize/<UPHOLD_CLIENT_ID>?scope=<UPHOLD_SCOPE>&intention=signup&state=<STATE>"

ayumi:
Fix #608

This lets you run tests and the app without special environment variables.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -81,6 +81,12 @@ development:
   from_email: Brave Payments Publishers Dev<brave-publishers@localhost.local>
   attr_encrypted_key: 499a4c51df667b4edfab40c1f8b813b7ed6ce02096d59f23b5dcb095369375f6
   secret_key_base: dd7b12788a804315fd75f1ff97fae33310b451c39d83d1d015543d25d0ba034c02fd312e83735f0d0aeaaf70131f26e614629e3c40531b949445b4dfacd3bad3
+  uphold_authorization_endpoint: "https://uphold.example.com/authorize/<UPHOLD_CLIENT_ID>?scope=<UPHOLD_SCOPE>&intention=signup&state=<STATE>"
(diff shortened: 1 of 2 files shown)


config/secrets.yml
```diff
@@ -81,6 +81,7 @@ development:
   from_email: Brave Payments Publishers Dev<brave-publishers@localhost.local>
   attr_encrypted_key: 499a4c51df667b4edfab40c1f8b813b7ed6ce02096d59f23b5dcb095369375f6
   secret_key_base: dd7b12788a804315fd75f1ff97fae33310b451c39d83d1d015543d25d0ba034c02fd312e83735f0d0aeaaf70131f26e614629e3c40531b949445b4dfacd3bad3
+  uphold_authorization_endpoint: "https://uphold.example.com/authorize/<UPHOLD_CLIENT_ID>?scope=<UPHOLD_SCOPE>&intention=signup&state=<STATE>"
```
nvonpentz: I got an error on Publishers#home, I think we need to add
```yml
uphold_client_id: test_client_id
uphold_client_secret: test_client_secret
uphold_scope: a:read,b:read,c:deposit
```
to development like we did with test.

ayumi: good catch. i updated it and tested out Publishers #home

Issue #805, opened 2018-04-25
SendGrid data initialization rake task


lgebhardt:
A rake task to initialize the Publishers list at SendGrid used by the SendGrid Update Service #804. This task should be able to be run more than once and will ensure all publishers are loaded into the list.

nvonpentz:
Resolved by #824

PR #824, opened 2018-05-03
Register email addresses with SendGrid

### LINKED ISSUE #804: SendGrid Update Service
Publishers will maintain a list at the Brave SendGrid account. When publisher records are created or updated this service will ensure the data kept at SendGrid will be updated. This needs to support new records, and changes to email addresses and contact information, such as names.

This service will need to communicate with the SendGrid API. SendGrid has the sendgrid-ruby gem (https://github.com/sendgrid/sendgrid-ruby) which provides a ruby interface to the the API and looks like a good place to start.

lgebhardt:
Uploads publishers as contacts to SendGrid as they are created and edited. Also adds them to a specified list. When emails are changed the old email is removed from the list and the new one added.

Includes a rake task to upload all email verified publishers to SendGrid. This rake task is safe to run more than once.

Fixes #804, #805

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/lib/send_grid/api_helper.rb
+++ b/lib/send_grid/api_helper.rb
@@ -1,0 +1,116 @@
+require 'digest'
+require "base64"
+
+module SendGrid
+  class Error < StandardError; end
+  class NotFoundError < StandardError; end
+  class Unauthorized < StandardError; end
+  class RateLimitExceeded < StandardError; end
+  class PageLimitExceeded < StandardError; end
+
+  class ApiHelper
+    class << self
+      def find_contact_by_email(email:)
+        params = { "email": email }
+        result = sg.client.contactdb.recipients.search.get(query_params: params)
+        check_result(result: result, sucess: '200')
+
+        json_result = JSON.parse(result.body)
+
+        if json_result['recipient_count'] == 0
+          raise NotFoundError.new
+        end
+
+        json_result['recipients'][0]
+      end
+
+      def get_contact_by_email(email:)
+        result = sg.client.contactdb.recipients._(id_from_email(email)).get
+        check_result(result: result, sucess: '200')
+        JSON.parse(result.body)
+      end
+
+      def upsert_contacts(publishers:)
+        raise PageLimitExceeded.new if publishers.length > 1000
+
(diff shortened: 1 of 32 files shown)


lgebhardt:
To install this on Production and Staging we should have two separate SendGrid accounts. 

The setup steps needed are:
- Create SendGrid account
- Generate an API key under [Settings / API Keys](https://app.sendgrid.com/settings/api_keys) with full access. Give it a name and record the secret key
- At Heroku add a config var for SENDGRID_API_KEY (recorded above)
- run the rake task `sendgrid:init` and record the list id it creates or returns
- At Heroku add a config var for SENDGRID_PUBLISHERS_LIST_ID (list id recorded above)

lib/send_grid/api_helper.rb
```diff
@@ -0,0 +1,129 @@
...
+        if json_result['recipient_count'] == 0
+          raise NotFoundError.new
+        end
+
+        json_result['recipients'][0]
+      end
+
+      def get_contact_by_email(email:)
+        result = sg.client.contactdb.recipients._(id_from_email(email)).get
+        check_result(result: result, sucess: '200')
```
yachtcaptain23: `success`

lib/send_grid/api_helper.rb
```diff
@@ -0,0 +1,129 @@
...
+          raise NotFoundError.new(result.body)
+        when '429'
+          raise SendGrid::RateLimitExceeded.new
+        else
+          raise "Unknown Error"
+        end
+      end
+
+      def sg
+        SendGrid::API.new(api_key: Rails.application.secrets[:sendgrid_api_key])
```
yachtcaptain23: Would you cache this value?
`@sg ||= SendGrid::API.new(api_key: Rails.application.secrets[:sendgrid_api_key])`

lib/send_grid/api_helper.rb
lgebhardt: Fixed

PR #825, opened 2018-05-03
Update promo stats sync logic

### LINKED ISSUE #821: Remove publisher parameter in the sentry call in PublisherPromoStatsFetcher
When promo services went down, the `PublisherPromoStatsFetcher` service would error.  Since we send publisher information to sentry, we end up creating new sentry error reports for each publisher.  This cluttered up the sentry feed.

https://github.com/brave-intl/publishers/blob/11a9f86d7854aade5d00d3cfd7701736945d0508/app/services/publisher_promo_stats_fetcher.rb#L23

We should remove the individual publisher information from the sentry call so we combine all of the `PublisherPromoStatsFetcher` reports into a single error.

Edit:  We are also synchronously calling `PublisherPromoStatsFetcher` in `PublishersController#home`.  We should make this async so if the call fails the user doesn't get the heroku error screen.

nvonpentz:
Resolves #821 

* Use SyncPublisherPromoStatsJob in `PublishersController#home` to make stats refresh async

* Adjust SyncPublisherPromoStatsJob to be able to be used for a single publisher

* Adjust SyncPublisherPromoStatsJobTest

* Remove `publisher` from sentry call in PublisherPromoStatsFetcher

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -321,9 +321,8 @@ class PublishersController < ApplicationController
   # Domain verified. See balance and submit payment info.
   def home
     if current_publisher.promo_stats_status == :update
-      PublisherPromoStatsFetcher.new(publisher: current_publisher).perform
+      SyncPublisherPromoStatsJob.perform_later(publisher: current_publisher)
     end
-
     # ensure the wallet has been fetched, which will check if Uphold needs to be re-authorized
     # ToDo: rework this process?
     current_publisher.wallet
--- a/app/jobs/sync_publisher_promo_stats_job.rb
+++ b/app/jobs/sync_publisher_promo_stats_job.rb
@@ -3,12 +3,17 @@ class SyncPublisherPromoStatsJob < ApplicationJob
   include PromosHelper
   queue_as :transactional
 
-  def perform(promo_id: active_promo_id)
-    # Might be able to select fewer with a more specific query
-    publishers = Publisher.where(promo_enabled_2018q1: true)
+  # Syncs promo stats for all publishers by default
+  # If a publisher is provided, sync stats for only that publisher
 
-    publishers.find_each do |publisher|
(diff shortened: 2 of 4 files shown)


app/controllers/publishers_controller.rb
```diff
@@ -321,9 +321,8 @@ def show
   # Domain verified. See balance and submit payment info.
   def home
     if current_publisher.promo_stats_status == :update
-      PublisherPromoStatsFetcher.new(publisher: current_publisher).perform
+      SyncPublisherPromoStatsJob.perform_later(publisher: current_publisher)
```
yachtcaptain23: I don't think this matches with the parameters of `perform`

nvonpentz: Hm, not sure I follow.  `SyncPublisherPromoStatsJob` takes two optional parameters, `active_promo_id` and `publisher`.  Unless specified, `active_promo_id` will be the default set by the config var, and if `publisher` is not supplied, the job will run for all publishers.

app/jobs/sync_publisher_promo_stats_job.rb
```diff
@@ -3,12 +3,17 @@ class SyncPublisherPromoStatsJob < ApplicationJob
...
   queue_as :transactional
 
-  def perform(promo_id: active_promo_id)
-    # Might be able to select fewer with a more specific query
-    publishers = Publisher.where(promo_enabled_2018q1: true)
+  # Syncs promo stats for all publishers by default
+  # If a publisher is provided, sync stats for only that publisher
 
-    publishers.find_each do |publisher|
+  def perform(promo_id: active_promo_id, publisher: false)
```
lgebhardt: I think it would be clearer to leave the `false` out as the default value and have `publisher` set to nil by default. The code should run the same way, but with `false` I think my other choice is `true` and not a Publisher.

PR #833, opened 2018-05-04
[WIP] SiteChannelVerifier always verifies locally

### LINKED ISSUE #815: Verify SiteChannels within Publishers
Currently SiteChannel verification relies on [Eyeshade](https://github.com/brave-intl/bat-ledger/tree/master/eyeshade). While this is good because it protects a sensitive part of our system, it's bad because it adds a layer of indirection and duplicates channel and publisher data, thus impeding development speed.

This is blocked on publisher settlement generation within Publishers (Issue TBD).

ayumi:
#### TODO: Verify if Eyeshade changes are required.
Publishers will handle Site Channel verification. Upon verification, it will call PublisherChannelSetter to sync verified channels to Eyeshade. TODO: Determine if Eyeshade needs prior knowledge of channels.

Fix #815

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/services/site_channel_verifier.rb
```diff
@@ -87,33 +56,41 @@ def verified_channel_post_verify
       PublisherMailer.verification_done_internal(verified_channel).deliver_later
     end
     SlackMessenger.new(message: "*#{verified_channel.publication_title}* verified by #{verified_channel.publisher.name} (#{verified_channel.publisher.email}); id=#{verified_channel.id}").perform
+
+    # Let eyeshade know about the new Publisher
+    begin
```
yachtcaptain23: I don't think we need to wrap it around in `begin` `perform`. If it crashes, my hunch is that we want it to crash loudly.

app/services/site_channel_verifier.rb
```diff
@@ -87,33 +56,41 @@ def verified_channel_post_verify
...
+      PublisherChannelSetter.new(publisher: @publisher).perform
```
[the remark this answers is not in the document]
yachtcaptain23: I agree: I personally don't want any publisher business logic to be a blocking call that's on the same thread as what the customer uses (with exception to our Postgres database and Redis database).

app/services/site_channel_verifier.rb
```diff
@@ -87,33 +56,41 @@ def verified_channel_post_verify
...
+    end
   end
 
-  def verify_offline_publisher_id
-    Rails.logger.info("PublisherVerifier offline by #{channel.details.verification_method}")
+  def verify_publisher_id
+    Rails.logger.info("PublisherVerifier by #{channel.details.verification_method}")
 
     case channel.details.verification_method
       when "dns_record"
```
yachtcaptain23: I'd prefer if these were constants so pushing into staging would throw an error if there was a typo.

app/services/site_channel_verifier.rb
```diff
@@ -140,25 +117,25 @@ def verify_offline_publisher_id_dns
...
-  def verify_offline_publisher_id_public_file
+  def verify_publisher_id_public_file
     generator = SiteChannelVerificationFileGenerator.new(site_channel: channel)
     uri = URI("https://#{channel.details.brave_publisher_id}/.well-known/#{generator.filename}")
     response = fetch(uri: uri)
     if response.code == "200"
       token_match = /#{channel.details.verification_token}/.match(response.body)
       if token_match
-        Rails.logger.warn("verify_offline_publisher_id_public_file: Token Found")
+        Rails.logger.warn("verify_publisher_id_public_file: Token Found")
```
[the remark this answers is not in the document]
lgebhardt: I look at the log statements in development. So if we don't want them in production we could set to a lower priority.

app/services/site_channel_verifier.rb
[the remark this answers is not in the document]
ayumi: i agree i think async channel setting is a good idea. here i copied the way it's done with OAuth callbacks for consistency.

i think fixing this is out of scope for this PR so i created #842 to track it.

app/services/site_channel_verifier.rb
ayumi: Failures are detected with this code by sending events to Sentry, and we know reasonable fallback solutions:
- Synchronous user initiated verification: Show failure message. Verify again in the future, which happens automatically every hour.
- Async verification: Fail and retry the Eyeshade channel setting. Actually, we don't have a logic path for this -- if this job reruns at this point it will do nothing since channel.verified is true. It would make sense to add this to #842.

app/services/site_channel_verifier.rb
ayumi: This was existing code so I opted not to change it. However, I agree with the sentiment to constantize and will update the code.

ayumi: just kidding i'm not sure the optimal way to do this

app/services/site_channel_verifier.rb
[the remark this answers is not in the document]
ayumi: have downgraded to `#debug`

ayumi:
I've updated this PR to support restricted site channel domains.

In case someone verifies a restricted channel, the app sets `Channel #verification_status` to `"verification_awaiting_admin_approval"`, it sends an internal email to admins, and notifies the user an admin is looking into it.

To finalize, an admin needs to open a rails console and run:
```ruby
SiteChannelVerifier.new(admin_approval: true, channel: <Channel instance>).preform
```

@evq

ayumi:
I successfully locally tested this against Eyeshade, which has the [relevant patch to POST /v2/owners](https://github.com/brave-intl/bat-ledger/pull/259) to support adding verified site channels.

Site channel restriction also works, blocking until an admin manually runs SiteChannelVerifier.

I'm reasonably confident this code will work, so next step is to test on Publishers staging against Eyeshade staging.
Then we should deploy the Eyeshade patch to prod and Publishers to prod.

@evq @nvonpentz @yachtcaptain23

Issue #844, opened 2018-05-08
Bug: Broken redirect


yachtcaptain23:
https://sentry.io/brave-software/publishers-production-t2/issues/497125846/?query=is:unresolved

PR #845, opened 2018-05-08
Fix an issue where a redirect had the wrong parameter type for last v…

yachtcaptain23:
…erification

Submitter Checklist:

- [x] Ticket: https://github.com/brave-intl/publishers/issues/844
- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/site_channels_controller.rb
+++ b/app/controllers/site_channels_controller.rb
@@ -135,7 +135,7 @@ class SiteChannelsController < ApplicationController
 
   def require_https_enabled_site
     return if current_channel.details.supports_https?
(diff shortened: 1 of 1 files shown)


app/controllers/site_channels_controller.rb
```diff
@@ -135,7 +135,7 @@ def require_unverified_site
 
   def require_https_enabled_site
     return if current_channel.details.supports_https?
-    redirect_to(site_last_verification_method_path(current_channel.details), alert: t(".alert"))
+    redirect_to(site_last_verification_method_path(channel=current_channel), alert: t(".alert"))
```
nvonpentz: This change will yield a missing translation.

image

PR #846, opened 2018-05-08
 Resolve Uphold disconnection / reauthorization issues

### LINKED ISSUE #841: Disconnecting from Uphold may show "Connection problems" status
It's been observed on production that disconnecting from Uphold occasionally results in displaying a status of "Connection problems". This status is triggered when eyeshade returns a wallet status of `re-authorize`. This appears to the user as:

screen shot 2018-05-07 at 3 37 19 pm

Instead, the "Not connected" status should always be displayed when a user has explicitly disconnected from Uphold. Any backend connection issues are irrelevant at this point.

dgeb:
When eyeshade responds that Uphold authorization / reauthorization is needed, the `uphold_verified` status should not be revoked immediately, because this status can be resolved either by a user or by Uphold without involving Brave. Instead `:reauthorization_needed` should be viewed as a a substate of `:verified`.

This approach also fixes the case in which a user has explicitly disconnected their Uphold account. They will always be shown a status of “Not connected” and never be shown “Connection problems”.

This PR also removes the now unused `Publisher#uphold_complete?` method.

Fixes #841

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -146,16 +146,6 @@ class Publisher < ApplicationRecord
     save!
   end
 
-  def uphold_complete?
-    # check the wallet to see if the connection to uphold has been been denied
-    action = wallet.try(:status).try(:[], 'action')
-    if action == 're-authorize' || action == 'authorize'
-      false
-    else
-      self.uphold_verified || self.uphold_access_parameters.present?
-    end
-  end
-
   def uphold_status
     if self.uphold_verified
       :verified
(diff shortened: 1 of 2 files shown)


app/models/publisher.rb
```diff
@@ -79,13 +79,6 @@ def wallet
 
     # if the wallet call fails the wallet will be nil
     if @_wallet
```
yachtcaptain23: Would you add early termination like `return nil if @_wallet.nil?` to not have nested if statements?

PR #851, opened 2018-05-08
Addresses an issue where a redirect had the wrong parameter type for …

### LINKED ISSUE #844: Bug: Broken redirect

yachtcaptain23:
…last verification

Fixes #844

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/site_channels_controller.rb
+++ b/app/controllers/site_channels_controller.rb
@@ -135,7 +135,7 @@ class SiteChannelsController < ApplicationController
 
   def require_https_enabled_site
     return if current_channel.details.supports_https?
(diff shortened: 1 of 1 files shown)


yachtcaptain23:
I'm sort of stumped by this bug (I don't know RoR's I18n implementation that well). Basically the string needs to be explicitly set instead of relative, otherwise I18n will look from the caller. 

The before behavior lookup was against `en.site_channels.download_verification_file.alert`

Issue #819, opened 2018-05-02
GET API for channel verification_status


yachtcaptain23:
This is a public API for channel verification_status and possibly include relevant data such as verification time. 

We may need to verify marketing status to see if the user opted-in.

PR #857, opened 2018-05-09
Fix broken tests in staging.

### LINKED ISSUE #856: Fix broken test regarding GET /api/channels/<id>/verification_status
I broke tests on staging to quickly get something tested. Next time I'll test by pushing to staging2.

yachtcaptain23:
Closes #856

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/api/channels_controller.rb
+++ b/app/controllers/api/channels_controller.rb
@@ -81,6 +81,7 @@ class Api::ChannelsController < Api::BaseController
     render(json: channel.details, status: :ok)
   end
 
+  # Copied from eyeshade/controllers/publishers#getToken
   def verification_status
     if @current_channel.verified?
       render(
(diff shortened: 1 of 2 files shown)


nvonpentz:
For some background, this was the original issue https://github.com/brave-intl/publishers/issues/819. The goal is to expose verification status to the ledger so publishers becomes the source of truth for the in-browser green verification checkmarks beside the url bar.

https://github.com/brave-intl/publishers/pull/853 was the first PR to address #819 

https://github.com/brave-intl/publishers/pull/855 is the second PR that that made some adjustments after some feedback from from the ledger, but this broke some tests.

This PR (#857) fixes the broken tests.

@yachtcaptain23 is that an accurate summary?

I'm having trouble following all the changes between these.  I'm thinking it might be better to just revert #855 then #853, then introduce a new PR with all those changes.  What do you think @yachtcaptain23 @dgeb?

Also, have we confirmed with the ledger team the api interaction is successful?

yachtcaptain23:
Yes I'm fine with reverting. I'll push a separate PR.

PR #866, opened 2018-05-14
Feature/get api channels identity

yachtcaptain23:
Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/models/json_builders/identity_json_builder.rb
+++ b/app/models/json_builders/identity_json_builder.rb
@@ -76,7 +76,6 @@ class JsonBuilders::IdentityJsonBuilder
   end
 
   def build_twitch_identity_json
-    # TODO with real values
     Jbuilder.encode do |json|
       json.publisher      @publisher_name
       json.publisherType  'provider'
@@ -94,9 +93,12 @@ class JsonBuilders::IdentityJsonBuilder
   end
 
   def build_properties(json)
+    require 'publishers/excluded_channels'
     if @channel_detail.present? && @channel_detail.channel.verified?
       json.verified true
-      json.timestamp (Time.now.to_i << 32)
+      # (Albert Wang): To satisfy backwards compatibility in Ledger's v3.identity
+      json.timestamp (@channel_detail.channel.updated_at.to_i << 32)
+      json.exclude Publishers::ExcludedChannels.excluded?(@channel_detail)
     end
   end
 end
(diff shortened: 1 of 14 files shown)


app/models/json_builders/identity_json_builder.rb
```diff
@@ -0,0 +1,102 @@
...
+      json.properties     do
+        build_properties(json)
+      end
+    end
+  end
+
+  def build_properties(json)
+    if @channel_detail.present? && @channel_detail.channel.verified?
+      json.verified true
+      json.timestamp (Time.now.to_i << 32)
```
evq: you should leave a comment explaining why we had to do this

app/models/json_builders/identity_json_builder.rb
```diff
@@ -0,0 +1,104 @@
...
+    @parsed_publisher_name = URL_REGULAR_EXPRESSION.match(@publisher_name)
+    find_channel_detail
+  end
+
+  def find_channel_detail
+    @channel_detail = if @parsed_publisher_name.present? && @parsed_publisher_name[1] == Channel::YOUTUBE
+      YoutubeChannelDetails.find_by(youtube_channel_id: @parsed_publisher_name[3])
+    elsif @parsed_publisher_name.present? && @parsed_publisher_name[1] == Channel::TWITCH
+      TwitchChannelDetails.find_by(twitch_channel_id: @parsed_publisher_name[3])
+    else
```
lgebhardt: Consider having a case where @parsed_publisher_name is present does not match youtube or twitch. As it is that would flow to finding the site (and failing) and then the `build_site_identity_json` logic. If that's acceptable a comment about why would help.

app/models/json_builders/identity_json_builder.rb
```diff
@@ -0,0 +1,104 @@
...
+      end
+    end
+  end
+
+  def build_properties(json)
+    require 'publishers/excluded_channels'
+    if @channel_detail.present? && @channel_detail.channel.verified?
+      json.verified true
+      # (Albert Wang): To satisfy backwards compatibility in Ledger's v3.identity
+      json.timestamp (@channel_detail.channel.updated_at.to_i << 32)
```
lgebhardt: What is the purpose of the timestamp? A note on the logic would be good here as well. That would help me understand why it's based on `updated_at` vs `verified_at`.

The timestamp logic doesn't seem to be tested. The reason for the bit shift isn't apparent so a note would be helpful as well as a test.

PR #869, opened 2018-05-15
Keep track of a publisher's status

### LINKED ISSUE #850: Publisher Status
Before we implement `status` on publishers, I want to make sure we get the states down correctly:

`created` as opposed to `create`
`onboarding` as opposed to `updating`
`active` 
`suspended`
`deleted` which is just `archived`

@nvonpentz let me know if you agree or disagree.

nvonpentz:
Resolves #850 

Note: This PR does nothing to address existing publishers.  Please see my [comment](https://github.com/brave-intl/publishers/issues/850#issuecomment-388955125) on the challenges and possible half solutions for that.

* Add new table PublisherStatusUpdates

* Automatically create a PublisherStatusUpdate (created) upon creation of a publisher

* Create a PublisherStatusUpdate (onboarding) upon verifying email

* Create a PublisherStatusUpdate (active) upon agreeing to TOS and addressing 2fa prompt

* Add new Publisher & PublisherStatusUpdate fixtures

* Add tests

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/models/publisher.rb
```diff
@@ -198,7 +203,30 @@ def has_verified_channel?
...
   end
 
+  def last_status_update
+    status_updates.first
+  end
+
   private
+  
+  def set_created_status
+    created_publisher_status_update = PublisherStatusUpdate.new(publisher: self, status: "created")
```
yachtcaptain23: Would you convert the statuses into constants? I recommend keep track of statuses in something similar to:

`STATUSES = [ONBOARDING, CREATED, ACTIVE, SUSPENDED]`

nvonpentz: I added a list a constant `ALL_STATUSES` on the `PublisherStatusUpdates` model, and added a validation that the status must be in the list of all statuses.  Please let me know if you had something else in mind @yachtcaptain23.

app/models/publisher_status_update.rb
```diff
@@ -0,0 +1,11 @@
+class PublisherStatusUpdate < ApplicationRecord
+  has_paper_trail
+  
+  ALL_STATUSES = ["created", "onboarding", "active"]
```
yachtcaptain23: Would you add the `suspended` state to this list?

Additionally, let's make these constants:

```
CREATED = 'created'
ONBOARDING = 'onboarding'
ACTIVE = 'active'
SUSPENDED = 'suspended'
ALL_STATUSES = [CREATED, ONBOARDING, ACTIVE, SUSPENDED]
``

File: app/models/publisher_status_update.rb [hunk 3]
```diff
@@ -0,0 +1,11 @@
+class PublisherStatusUpdate < ApplicationRecord
+  has_paper_trail
```
yachtcaptain23: Make sure you delete this line. This model essentially is its own paper_trail.

I realize it seems silly to run our own paper trail, but essentially we don't want to bloat 1 table too hard to affect the rest of the database which depends on a `versions` table.

PR #887, opened 2018-05-18
Migration/channel verified at

yachtcaptain23:
Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -115,7 +115,7 @@ class Channel < ApplicationRecord
     self.reload
 
     self.verified = false
-    self.verified_at = false
+    self.verified_at = nil
     self.verification_status = 'failed'
     self.verification_details = details
     self.save!(validate: false)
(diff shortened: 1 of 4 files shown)


dgeb:
I agree with @lgebhardt's recommendation to add a validation and then clear `verified_at` whenever `verified` is set to `false`.

I'm on the fence as to whether we should initialize `verified_at` for all already verified accounts via a db migration. If we did so, the initial value would have to be a best guess like the `updated_at` timestamp. It also seems fine to leave `verified_at` as `nil` for these pre-existing accounts.

yachtcaptain23:
Cool I added an `after_save` hook to clear. 

And I prefer we just only look at `verified_at` as when a Channel was verified and keep it at that rather than depend on it implicitly. Ideally if I were to rebuild this, we'd get rid of `verified` column and keep the logic in `verified_status`.

app/models/channel.rb
```diff
@@ -114,6 +115,7 @@ def verification_failed!(details = nil)
     self.reload
 
     self.verified = false
+    self.verified_at = false
```
dgeb: `self.verified_at = nil`

PR #894, opened 2018-05-21
Maintain a list of excluded channels from payout. Closes #891

### LINKED ISSUE #891: Have Channels that are excluded from payout
<strike> Going to add the field `excluded_from_payout` as part of a Channel </strike>

Update: Spoke with eV. We're going to store the list in a config file.

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


yachtcaptain23:
I don't think we'll be updating the exclusion list anytime soon. I prefer this goes out earlier. I'll go ahead and update the count.

nvonpentz:
I'd also prefer we store this in a db for the sake of not having sketchy/offensive domains explicitly referenced in our source.  Open `excluded_site_channels.yml` and search for `# adult` and you'll see what I mean.

evq:
This list is very infrequently updated (though we seriously need to go through and revise/prune the list atm, it's a carry over from prototype days and does not align with our current intent - adult sites being a perfect example).

The advantage of storing it in git is revision control and having the PR process for the infrequent updates that do happen. I also think it is more transparent. Our unhappiness with the list as is should drive that revision process as a priority IMO

PR #908, opened 2018-05-24
Update attr_encrypted gem

### LINKED ISSUE #907: Encrypted columns nil if ActiveRecord DB connection down when app loaded
I found that if the database is not created when the app was loaded the encrypted columns were not getting saved. This was resulting in nil tokens when the app was first started in docker. It also mirrors behavior seen but not diagnosed on production.

Propose updating the `attr_encrypted` gem to 3.1.0. https://github.com/attr-encrypted/attr_encrypted/commit/98769e94af199a7ea8ae6804db8bb2d9fbc8999e fixes this behavior.

lgebhardt:
Updates the `attr_encrypted` gem to 3.1.0.

Captures an exception to sentry if we encounter a nil token.

Fixes #907 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/mailers/publisher_mailer.rb
+++ b/app/mailers/publisher_mailer.rb
@@ -208,6 +208,16 @@ class PublisherMailer < ApplicationMailer
   private
 
   def ensure_fresh_token
-    raise if @publisher.authentication_token.nil? || @publisher.authentication_token_expires_at <= Time.now
+    # Check if we are missing the token and capture to sentry if we are. This should not happen.
+    begin
+      raise "missing token" if @publisher.authentication_token.nil?
+    rescue => e
+      require 'sentry-raven'
+      Raven.capture_exception(e)
+      raise
+    end
+
+    # Expired tokens are expected and will not be logged
+    raise "expired token" if @publisher.authentication_token_expires_at <= Time.now
   end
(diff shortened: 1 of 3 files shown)


app/mailers/publisher_mailer.rb
```diff
@@ -208,6 +208,16 @@ def verified_invalid_wallet_internal(publisher, params)
   private
 
   def ensure_fresh_token
-    raise if @publisher.authentication_token.nil? || @publisher.authentication_token_expires_at <= Time.now
+    # Check if we are missing the token and capture to sentry if we are. This should not happen.
+    begin
+      raise "missing token" if @publisher.authentication_token.nil?
+    rescue => e
+      require 'sentry-raven'
```
yachtcaptain23: Do we need these lines to capture errors in Sentry? I had always thought when the app loads, it loads Sentry as an exception handler before it gets re-raised even higher.

lgebhardt: I'm not actually sure. @ayumi set things up like this and I've been following the convention. I would like to know the reasoning, so I'll look into it.

PR #910, opened 2018-05-24
Replace v1.identity with a domain normalizer. Closes #817

### LINKED ISSUE #817: Normalize SiteChannel domains
currently we request Eyeshade /v1/publishers/identity 

as part of the publishers and channels refactor we would like to remove the call to Eyeshade and consolidate most channel logic in Publishers.

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/services/site_channel_domain_setter.rb
```diff
@@ -13,22 +13,23 @@ def perform
   private
 
   def normalize_domain
-    require "faraday"
+    require 'addressable'
 
-    channel_details.brave_publisher_id = SiteChannelDomainNormalizer.new(domain: channel_details.brave_publisher_id_unnormalized).perform
+    unless channel_details.brave_publisher_id_unnormalized.starts_with?("http")
```
lgebhardt: Explicitly check for "http://" or "https://" otherwise domains that start with "http", i.e "http-tips.com" will not get the required protocol added.

lgebhardt:
> I'll make the changes, but because we're handling this locally, I don't think it makes sense to run this logic as a subclass of `BaseApiClient`.

@yachtcaptain23 I agree, but I do think it makes sense to keep it encapsulated in a service derived from `BaseService`.

yachtcaptain23:
From what I can tell, the two things that the service does is 
(1) Normalize unnormalized_url into the domain 
(2) Check if the domain is a valid `canonical_tld?`.

As a result, this implementation is done in ~7 lines.

yachtcaptain23:
Refactoring this to use PublicSuffix

test/services/site_channel_domain_setter_test.rb
```diff
@@ -78,7 +103,13 @@ def teardown
     channel_details = SiteChannelDetails.new
     channel_details.brave_publisher_id_unnormalized = "https://bad url.com"
     SiteChannelDomainSetter.new(channel_details: channel_details).perform
+    assert_equal 'invalid_uri', channel_details.brave_publisher_id_error_code
+  end
 
+  test "raises exception with invalid url without protocol" do
```
lgebhardt: minor nit - test name indicates raises an exception but sets brave_publisher_id_error_code

The next test name is the same

test/services/site_channel_domain_setter_test.rb
```diff
@@ -33,6 +34,30 @@ def teardown
     assert channel_details.host_connection_verified
   end
 
+  test "normalizes domain for http" do
```
lgebhardt: Could we add an http://http-lib.com case as well?

PR #915, opened 2018-05-29
update list of excluded channels

evq:
Per discussion, list was pruned to adhere to the general rule of thumb is that we try to not auto-pay sites that users already pay, such as commerce sites, financial services sites like banks, SaaS vendors and government sites.

This list should more or less have the above minus gov sites


test/unit/excluded_channels_test.rb
```diff
@@ -3,15 +3,15 @@
 
 class ExcludedChannelsTest < ActiveSupport::TestCase
   test "excluded_brave_publisher_id? for excluded" do
-    assert Publishers::ExcludedChannels.excluded_brave_publisher_id?("google.com")
+    assert Publishers::ExcludedChannels.excluded_brave_publisher_id?("amazon.com")
```
evq: Al can we change this test to not depend on the list at all? e.g. a setup step that runs before any of these tests and replaces the list with one for testing

PR #917, opened 2018-05-30
Feature/v3 timestamp port

yachtcaptain23:
Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


yachtcaptain23:
I don't want to add an index if we don't have to. This API is for the browser to determine if it's local cache is expired or not based on the channels being updated. @evq would I be able to get your opinion on this?

PR #918, opened 2018-05-30
Feature: Logging service into NewRelic and Sentry

### LINKED ISSUE #909: Add LogException Class
The current way of loading exceptions is cumbersome. We should use a dedicated service for logging metrics as to make lines used down to 1 as opposed to 4 or more.

yachtcaptain23:
Closes #909

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/services/log_exception.rb
+++ b/app/services/log_exception.rb
@@ -1,13 +1,16 @@
 class LogException
   def self.perform(error, publisher: {}, params: {}, force: false)
-    if Rails.env.production? || force
+    if Rails.env.production? || Rails.env.staging? || force
+
+      require 'newrelic_rpm'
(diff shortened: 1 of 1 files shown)


yachtcaptain23:
I haven't figured out how to write tests around this. In the event that New Relic or Sentry goes down, I'm also considering to make this an ActiveJob to prevent outbound calls from stopping any code from working.

app/services/log_exception.rb
```diff
@@ -0,0 +1,24 @@
+class LogException
+  def self.perform(error, publisher: {}, params: {}, force: false)
+    if Rails.env.production? || force
```
nvonpentz: Can we also send exceptions to sentry on staging?

Issue #542, opened 2018-01-26
Uphold balance and statements


jenn-rhim:
Edited by Al Wang:

Action Items:
- [x] Add Settlements (DONE)
- [ ] See if eyeshade API's are done for charts
- [ ] Add still loading while the information is being fetched
- [ ] Render charts
- [ ] Fetch statements during the service

ayumi:
more info needed

ayumi:
Possibly fixed by #567

nvonpentz:
Related to #265

@jenn-rhim What needs to be done here?

jenn-rhim:
This one requires a bit of work (initially intended to be part of second half of Pub 2.0 design, of which we only finished Account model change + multi-channel support).

Here are the key screens involving Statements 
1. Dashboard main (I'm not proposing that we make this change for the time being)/
2. Blankslate of Statements page
3. Statements page: **This is where we need to add "**Promo**" parameter


![1-dash-2](https://user-images.githubusercontent.com/4369856/37498549-3a92e514-287c-11e8-9b42-9597507d5d49.png)

![2-statement-blankslate](https://user-images.githubusercontent.com/4369856/37498553-3f130eca-287c-11e8-94da-1ee36e7e7583.png)

![3-2statement-long](https://user-images.githubusercontent.com/4369856/37498563-435466d2-287c-11e8-8327-517927402be6.png)

jenn-rhim:
@nvonpentz  Ideally, I’d like to see at least 3 dimensions that the user can use to filter the data records: 1. Channels (‘All’ by default), 2. Promotion, 3. Time period. Promotion has its own time period that may not coincide with the regular settlement cycle, choosing a specific promotion (e.g. February Referral Promo) would override the Time Period parameter.

Minimally, we transfer over the current time period dropdown, and allow user to generate the statement, and we include the promo results inside of that report if the time period overlaps.  
@mandar-brave

dgeb:
@jenn-rhim @mandar-brave how would you like the "Deposit to Uphold" panel to appear when there is no Uphold wallet connected?

Currently this panel appears as:

![34235980-0e2fdd66-e5aa-11e7-8820-233a3336c43b](https://user-images.githubusercontent.com/29122/37914174-997c38da-30e4-11e8-9772-cd3767d6a0a1.png)

[...]

PR #922, opened 2018-05-30
Introduce separate statements page

dgeb:
This is an intermediate implementation of #542. It introduces a new statements page and moves all the existing statement generation and viewing code from the dashboard to that page. Complete implementation of #542 is blocked on eyeshade generating statements automatically.

This PR also introduces new styling for `select` elements that's consistent with what @jenn-rhim mocked in #542.

The dashboard will now show the "Statements" panel as follows before Uphold has been connected:

screen shot 2018-05-30 at 2 44 21 pm

Once Uphold has been connected, the user will be able to link to the statements page:

screen shot 2018-05-30 at 4 55 41 pm

And once statements have been generated, the most recent one will be shown as a link:

screen shot 2018-05-30 at 4 45 58 pm

This is the statements page before any statements have been generated:

screen shot 2018-05-30 at 2 56 22 pm

And while a statement is being generated:

screen shot 2018-05-30 at 2 56 29 pm

Here it is with a couple statements ready to be viewed:

screen shot 2018-05-30 at 2 56 40 pm

Finally, the statement generation links are hidden when all possible periods have been requested:

screen shot 2018-05-30 at 2 57 04 pm

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/controllers/publishers_controller.rb
```diff
@@ -325,6 +326,9 @@ def home
     current_publisher.wallet
   end
 
+  def statements
```
yachtcaptain23: Deadcode?

app/helpers/publishers_helper.rb
```diff
@@ -231,7 +231,23 @@ def statement_period_description(period)
   end
 
   def statement_period_date(date)
-    date.strftime('%B %e, %Y')
+    date.strftime('%B %e')
+  end
+
+  def link_to_most_recent_statement
+    most_recent_statement = current_publisher.statements.order(created_at: :desc).first
+    if most_recent_statement
```
yachtcaptain23: Nit pick: I prefer using `most_recent_statement.present?`, even though the results are the same. The reviewer has to be certain that the type is not a collection, otherwise an empty collection would enter this block.

dgeb: Good call 👍 I just amended that commit.

dgeb:
@yachtcaptain23 thanks for your review. I've amended the last commit with the change you requested. 

I also added download icons to the statements page, as shown below (and as specified in #542):

screen shot 2018-05-31 at 8 52 43 am

PR #935, opened 2018-06-04
Remove PublisherNotifierUnverified service

### LINKED ISSUE #785: PublisherNotifierUnverified service is not used anywhere and is missing tests
The `PublisherNotifierUnverified` service used to be referenced in`API::PublishersController`, but was removed during multi channel, along with all of the tests. This is the original pull request https://github.com/brave-intl/publishers/pull/429.

The `PublisherNotifier` is used by the `API::ChannelsController`, so that's where the `PubisherNotifierUnverified` service should be used as well.

We should also consider changing the name of the service, since there is no notion of an unverified publisher.  Maybe `UnregisteredPublisherNotifier`?

Eyeshade has not implemented this yet.

Related to https://github.com/brave-intl/bat-ledger/pull/107

nvonpentz:
Resolves #785 

* Remove the service class

* Remove the whois gems from gemfile

Submitter Checklist:

- [x] Tagged reviewers.
- [x] Integrated piwik/matomo (for code that adds new buttons).

Test Plan:

Reviewer Checklist:

Tests

Security:


yachtcaptain23:
I believe `app/services/get_whois_emails_for_domain.rb` requires `whois`

nvonpentz:
Good catch @yachtcaptain23.  That's only being used in the PublisherNotifierUnverified service, so I've removed that file as well.

PR #941, opened 2018-06-05
[WIP] BAT by default

### LINKED ISSUE #834: Automatically create uphold BAT card for new publishers
Blocked on https://github.com/brave-intl/bat-ledger/issues/229

Action Items:
(1) Make an API call to the Eyeshade API once the API is created

### LINKED ISSUE #835: Have publishers explicitly set their default currency as part of uphold attachment

nvonpentz:
Resolves #834 
Resolves #835 

This code will automatically prompt all users that have verified with uphold, and have completed the uphold sign up (`authorized`) with the confirm default currency screen.  The user will select from a list of possible currencies, and hit confirm.  The default currency will be recorded and `default_currency_updated_at` will be set.

If the publisher has the correct scope they will be redirected to Publishers#home.  If the publisher does not have the correct scope, they will be redirected to the uphold authorization endpoint, then after authorizing, they will be redirected to Publishers#home.

Publishers#home will create the card if their confirmed currency is not available on uphold, but is a possible currency.  If BAT is not an available currency, Publishers#home will create that card as well (assuming they did not confirm this as their default currency).  Publishers#home will also update eyeshade to use the confirmed default currency.

### TO DO
* Use correct modal styles
* Add piwik buttons 
* Resolve feedback

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/controllers/publishers_controller.rb
```diff
@@ -500,6 +546,21 @@ def should_throttle_resend_auth_email?
...
+  def should_create_bat_card?
+    current_publisher.available_uphold_currencies.exclude?("BAT") && 
+    current_publisher.possible_uphold_currencies.include?("BAT") &&
+    current_publisher.default_currency != "BAT"
+  end
+
+  def should_update_eyeshade_default_currency?
+    current_publisher.eyeshade_default_currency != current_publisher.default_currency
+  end
+
```
nvonpentz: Would these be more appropriate in the Publisher model?

dgeb: I think that would be preferable, since the logic is wholly based on the model and thus it could be unit tested there.

app/controllers/publishers_controller.rb
```diff
@@ -318,12 +344,28 @@ def show
...
+      end
+
+      if should_create_bat_card?
+        UpholdServices::CardCreationService.new(publisher: current_publisher, currency_code: "BAT").perform
+      end      
+
+      if should_update_eyeshade_default_currency?
+        PublisherDefaultCurrencySetter.new(publisher: current_publisher).perform
+      end
+    end
```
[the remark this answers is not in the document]
dgeb: Instead of blocking here, can we use a background job to check all publishers and create uphold cards as needed?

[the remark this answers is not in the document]
lgebhardt: This looks like it won't update the wallet in the `current_publisher` model instance. The `CardCreationService` and `PublisherDefaultCurrencySetter` will both cause changes in the wallet we may want to fetch.

app/controllers/publishers_controller.rb
```diff
@@ -318,12 +344,28 @@ def show
...
+    wallet = current_publisher.wallet.wallet_json
+
+    # Create Uphold cards if they need to be created
+    if current_publisher.uphold_verified && current_publisher.default_currency_confirmed_at.present?
+      if should_create_default_currency_card?
+        UpholdServices::CardCreationService.new(publisher: current_publisher,
+                                                currency_code: current_publisher.default_currency).perform
+      end
+
+      if should_create_bat_card?
```
lgebhardt: As discussed in a DM we might want to move the flow logic from `should_create_bat_card?` to this method in the case of a BAT card not being the default.

app/models/publisher.rb
```diff
@@ -164,6 +164,26 @@ def set_uphold_updated_at
...
+  end
+
+  def available_uphold_currencies
+    wallet.try(:wallet_details).try(:[], 'availableCurrencies')
+  end
+
+  def possible_uphold_currencies
+    wallet.try(:wallet_details).try(:[], 'possibleCurrencies')
+  end
+
```
lgebhardt: I'd like to move these methods to the wallet class. It would make it cleaner to remove the wallet from the publisher if we choose to move to a caching scheme. That could help clean up the flow when wallet modifying actions (like the `CardCreationService` and `PublisherDefaultCurrencySetter`) are taken in the same request (we could invalidate the cache afterwards and then the next action would pull a fresh wallet).

app/views/publishers/_confirm_default_currency_modal.html.slim
```diff
@@ -0,0 +1,22 @@
+javascript:
```
lgebhardt: We need to ensure that the cards are created and a default currency is chosen even if javascript is off.

config/locales/en.yml
```diff
@@ -376,6 +375,10 @@ en:
       confirmation: Are you sure you want to disconnect your account with Uphold?
       deny: Cancel
       confirm: Disconnect
+    confirm_default_currency_modal:
+      headline: Confirm your preferred deposit currency
+      leadin: You will receive monthly contributions from your Brave fans in this currency in your Uphold account.
```
lgebhardt: The wording is slightly awkward "from your Brave fans in this currency in your Uphold account".

app/controllers/publishers_controller.rb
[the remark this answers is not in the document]
dgeb: Perhaps we should introduce a new `uphold_status` value on the model that represents that these backend changes are needed and/or being processed? The status could then be polled on the controller via `PublishersController#uphold_status` and the dashboard rendering would not need to be blocked.

app/controllers/publishers_controller.rb
nvonpentz: I will move these to the model.

app/controllers/publishers_controller.rb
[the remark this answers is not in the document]
nvonpentz: @dgeb I think that's a possible solution.  If a publisher confirms their currency and we do settlement before the job runs, they won't be paid in the correct currency, though.

@lgebhardt I don't think we will need an updated wallet after this point, since the `default_currency` in the rails database is what will be displayed on the dashboard - though I could be wrong.  Is there a specific case you were thinking of?  How could we keep the wallet up-to-date?

app/models/publisher.rb
nvonpentz: I will move these to the wallet class.

app/views/publishers/_confirm_default_currency_modal.html.slim
nvonpentz: I will remove all js and make it a static modal.

config/locales/en.yml
nvonpentz: cc @jenn-rhim @mandar-brave

dgeb:
This has been superceded by #985

PR #942, opened 2018-06-06
Feature/admin dashboard

yachtcaptain23:
Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


config/secrets.yml
```diff
@@ -132,7 +132,7 @@ test:
   # a real api_key is needed for recording new VCR based tests
   sendgrid_api_offline: true
   sendgrid_api_key: <%= ENV["SENDGRID_API_KEY"] || "fakeapikey" %>
-  sendgrid_publishers_list_id: 3648346
+  sendgrid_publishers_list_id: <%= ENV["SENDGRID_PUBLISHERS_LIST_ID"] || 3648346 %>
```
lgebhardt: I think we should leave this as a fixed value, 3986776, from what @nvonpentz mentioned in #955. This will probably require either rebuilding the cassettes one more time, or just doing a search and replace (my choice).

yachtcaptain23:
If we add the user directly into the database, they won't be affected by `update_sendgrid`

PR #955, opened 2018-06-07
Use env variable for SENDGRID_PUBLISHERS_LIST_ID

### LINKED ISSUE #954: Setup env variable for secrets to include PUBLISHERS_SENDGRID_LIST_ID

yachtcaptain23:
Closes #954

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -130,9 +130,9 @@ test:
   should_send_notifications: true
   # Sendgrid
   # a real api_key is needed for recording new VCR based tests
-  sendgrid_api_offline: true
+  sendgrid_api_offline: false
   sendgrid_api_key: <%= ENV["SENDGRID_API_KEY"] || "fakeapikey" %>
-  sendgrid_publishers_list_id: 3648346
+  sendgrid_publishers_list_id: <%= ENV["SENDGRID_PUBLISHERS_LIST_ID"] || 3648346 %>
 
 # Do not keep production secrets in the repository,
 # instead read values from the environment.


yachtcaptain23:
@nvonpentz let me know if we need this. I prefer merging this earlier than later, and in terms of merge conflicts, a rebase should be easy.

config/secrets.yml
```diff
@@ -130,9 +130,9 @@ test:
   should_send_notifications: true
   # Sendgrid
   # a real api_key is needed for recording new VCR based tests
-  sendgrid_api_offline: true
+  sendgrid_api_offline: false
```
lgebhardt: This should remain true unless we update all the tests which also trigger the sendgrid updates.

config/secrets.yml
```diff
@@ -130,9 +130,9 @@ test:
...
   sendgrid_api_key: <%= ENV["SENDGRID_API_KEY"] || "fakeapikey" %>
-  sendgrid_publishers_list_id: 3648346
+  sendgrid_publishers_list_id: <%= ENV["SENDGRID_PUBLISHERS_LIST_ID"] || 3648346 %>
```
lgebhardt: Given the `sendgrid_publishers_list_id` is in the urls and/or body of the request we'll get matching errors if this isn't consistent. 

If we're having issues it's probably because the cassettes were recorded against two different accounts. We should upgrade all cassettes and this value to use the id from the new shared dev sendgrid account.

nvonpentz: `3986776` is the list id used by the shared acct

PR #964, opened 2018-06-11
Enforce 2fa and IP whitelisting for admins

### LINKED ISSUE #961: (Admin) IP Whitelist and 2FA

nvonpentz:
Resolves #961 

Submitter Checklist:

- [] Tagged reviewers.
- [x] Integrated piwik/matomo (for code that adds new buttons).

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -18,6 +18,10 @@ class ApplicationController < ActionController::Base
     current_publisher
   end
 
+  def current_ability
+    @current_ability ||= Ability.new(current_user, request.remote_ip)
+  end
+  
   def set_locale
     I18n.locale = params[:locale] || I18n.default_locale
   end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -3,9 +3,15 @@ class Ability
   include PublishersHelper
 
   ROLES = %i(admin publisher)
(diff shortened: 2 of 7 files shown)


app/controllers/application_controller.rb
```diff
@@ -18,6 +18,10 @@ def current_user
     current_publisher
   end
 
+  def current_ability
+    @current_ability ||= Ability.new(current_user, request.remote_ip)
```
nvonpentz: I followed these instructions to write this method https://gist.github.com/davejamesmiller/2189934

app/models/ability.rb
```diff
@@ -25,7 +32,21 @@ def publisher
...
 
   def admin
+    raise AdminNotOnIPWhitelistError.new("Administrator must be IP whitelisted") unless admin_ip_whitelisted?
+    raise TwoFactorDisabledError.new("2fa must be enabled for administrators") unless two_factor_enabled?(@publisher)
     can :manage, :all
     can :access, :all
   end
+
+  def admin_ip_whitelisted?
+    return true if ADMIN_IP_WHITELIST.blank?
```
evq: this makes sense for dev, can we add an extra check against RAILS_ENV to return false if we're in production though?

Issue #838, opened 2018-05-07
Publishers#home is slow because of synchronous calls to eyeshade


yachtcaptain23:
We still need to deep dive, but this is a problem that will probably scale poorly:

https://files.slack.com/files-pri/T04PX1BU8-FAK0BM9FT/image.png

Dan brought up a good point that we move the UI portion to use asynchronous javascript. I suggested we could do Redis caching like so:

If Eyeshade stores all the values in Redis:
`REDIS.get("wallet_<id>_balance")`

If we call Eyeshade and then cache the result
```
Rails.cache.fetch("wallet_<id>_balance", expires_in: 24.hours) do
  # Last value should be the value of the wallet balance
  9001
end
```

nvonpentz:
I think we should reduce the total number of calls to get a publisher's wallet and per channel balances to one, then make this call async.  

This way we don't block page load, increase overall response time, and avoid duplicating eyeshade data on publishers (which @evq had some objections to).

We'd have to wait until eyeshade can support doing all of this in one call, though.

cc @dgeb @ayumi

dgeb:
@nvonpentz An async approach sounds good to me. Some data, like channel balances, seems undesirable to cache because a flash of stale content could hurt more than it helps.

> We'd have to wait until eyeshade can support doing all of this in one call, though.

I agree this would be ideal, but I don't think this has to block implementation. If we issue an async call from the dashboard, the request could be fulfilled by any number of sync requests on the backend from publishers to eyeshade. It may take longer than necessary, but at least it wouldn't completely block the display of the dashboard.

yachtcaptain23:
@nvonpentz We can close this right? I forget if this is a N+1 issue.

nvonpentz:
@yachtcaptain23 You created this issue to improve the dashboard page load.

This was partially addressed in #1095 when we switched to using 1 API call for all channel balances instead of 1 API per channel.

However, the wallet is still fetched synchronously in Publishers#home, and that's still a bottleneck.  I plan to remove the synchronous call when we switch to the new dashboard charts.

We should also consider if there is anything we can do to improve the speed of the /wallet endpoint on eyeshade since that takes ~3 seconds.

evq:
@nvonpentz currently the wallet endpoint still involves the pre-transaction table balance calculation. We can remove that, which should help with response time somewhat

yachtcaptain23:
Going to move conversation to https://github.com/brave-intl/publishers/issues/1852 as it has more concise details.

PR #970, opened 2018-06-12
WIP: Feature/settlement info

### LINKED ISSUE #768: Display settlement information on the dashboard
Currently, we display only the contribution balance on the dashboard and don't indicate the settled amount that the user receives in their Uphold account. Contribution balance is the total accrued BAT amount to date prior to the settlement. Once it's settled, the balance becomes zero. This leaves the user with no easy way to verify whether or not the Uphold deposit is accurate until they generate the statement, which requires a number of improvements to make it useful. 

The following UI update addresses this by displaying the settlement summary in the Uphold wallet box.
- Display last deposit date
- Display last deposited BAT total

@nvonpentz 

publisher dashboard-deposit

lgebhardt:
This creates a new javascript wallet class to handle the wallet logic on the client. This wallet is now returned by the `balance` class and handles contribution calculations and the last settlement date. This goes a good ways towards allowing for the asynchronous fetching of the wallet data (#838), but because some of the wallet info is still needed by the server it will not result in the desired speedup.

It also updates the Your Uphold Wallet section to match #835, which is an updated version of the UI from #768. However @dgeb is adding the dialog that will replace the Default Currency drop down as a separate PR. Merging this should be coordinated with this PR.

Further work is expected to reorganize the server side wallet, once eyeshade is updated to handle balance requests for multiple channels at once.

Fixes #768 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/test/javascript/wallet-test.js
+++ b/test/javascript/wallet-test.js
@@ -1,0 +1,51 @@
+import { Wallet } from 'wallet';
+
+const { module, test } = QUnit;
+
+module('Wallet tests', function() {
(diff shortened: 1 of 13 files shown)


test/javascript/wallet-test.js
```diff
@@ -0,0 +1,51 @@
...
+  let wallet_data_without_settlement = {"providerWallet":{"provider":"uphold","authorized":true,"defaultCurrency":"USD","rates":{"BTC":"0.00003514","ETH":0.0004749371464487743,"LTC":0.0022794117647058827,"USD":0.28188557439999995,"EUR":0.24163829927299837},"availableCurrencies":["USD","BAT","BTC","EUR","XAU"],"possibleCurrencies":["AED","ARS","AUD","BAT","BCH","BRL","BTC","BTG","CAD","CHF","CNY","DASH","DKK","ETH","EUR","GBP","HKD","ILS","INR","JPY","KES","LTC","MXN","NOK","NZD","PHP","PLN","SEK","SGD","USD","XAG","XAU","XPD","XPT","XRP"],"scope":"cards:read user:read transactions:transfer:others"},"channelBalances":{"youtube#channel:UC1CF3Uud5SZJAOtusRcaMZg":{"probi":17737693781040825362},"youtube#channel:UCpspxI6hEuK9AVxNMtZsObA":{"probi":17737693781040825362}}};
+  let wallet_data_not_uphold_verified = {};
```
dgeb: I would use camelCase for these JS vars (e.g. `walletDataWithSettlement`)

dgeb:
This has been superceded by #985

PR #984, opened 2018-06-15
Allow admins to create notes for a publisher

### LINKED ISSUE #975: Create a PublisherNotes table
It would be helpful to for the support staff to be able to read and write notes associated with a publisher.  We should create a PublisherNotes table that is only readable and writable by admins.

Once #966 has been merged, we should integrate this on the Publishers #show page of the admin dashboard.

Refs #959 

Update:
The notes table and model have been merged in #976.  Now we just need to integrate adding/displaying notes into the admin dashboard.

nvonpentz:
Resolves #975 
image

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -32,7 +32,7 @@ class PublishersController < ApplicationController
                change_email)
   before_action :require_publisher_email_verified_through_youtube_auth,
                 only: %i(update_email)
-  # before_action :protect, only: %i(show home)
+  before_action :protect, only: %i(show home)
   before_action :require_verified_publisher,
     only: %i(disconnect_uphold
              edit_payment_info
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -32,7 +32,7 @@ class PublishersController < ApplicationController
                change_email)
   before_action :require_publisher_email_verified_through_youtube_auth,
                 only: %i(update_email)
-  before_action :protect, only: %i(show home)
+  # before_action :protect, only: %i(show home)
   before_action :require_verified_publisher,
     only: %i(disconnect_uphold
              edit_payment_info
--- a/app/views/admin/publishers/show.html.slim
+++ b/app/views/admin/publishers/show.html.slim
@@ -1,56 +1,59 @@
 #publisher
-  #publisher-info
+  #publisher-section
     h2 = @publisher.name
-    .db-info-row
-      .db-field = "Email:"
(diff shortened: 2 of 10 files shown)


app/helpers/admins_helper.rb
```diff
@@ -10,4 +10,9 @@ def channel_url(channel)
       raise "Unsupported channel type"
     end
   end
+
+  def note_creator(note)
```
yachtcaptain23: I'd prefer to keep the admin helper skinnier and the caller more transparent, in other words, I'd prefer to remove this.

app/views/admin/publishers/show.html.slim
```diff
@@ -72,4 +75,22 @@
...
+    #create-note.rounded-box
+      = form_for @note, { method: :post, url: create_note_admin_publishers_path } do |f|
+        = f.text_area(:note, id: "create-note-content")
+        = f.hidden_field(:publisher, :value => @publisher.id)
+        = f.submit("Add note", id: "submit-create-note")
+    - @publisher.notes.order(created_at: :desc).each do |note|
+      .note.rounded-box
+        .note-header
+          .note-created-by
+            = note_creator(note)
```
yachtcaptain23: `note.created_by.name`

app/views/admin/publishers/show.html.slim
```diff
@@ -72,4 +75,22 @@
...
-              .db-value = channel.promo_registration.referral_code
\ No newline at end of file
+              .db-value = channel.promo_registration.referral_code
+  hr
+  h3.admin-header = "Notes"
+  #notes-section
+    #create-note.rounded-box
+      = form_for @note, { method: :post, url: create_note_admin_publishers_path } do |f|
+        = f.text_area(:note, id: "create-note-content")
+        = f.hidden_field(:publisher, :value => @publisher.id)
```
yachtcaptain23: Nitpicking, but let's avoid rocket syntax, e.g. `:value => @publisher.id` becomes `value: @publisher.id`

app/controllers/publishers_controller.rb
```diff
@@ -32,7 +32,7 @@ class PublishersController < ApplicationController
                change_email)
   before_action :require_publisher_email_verified_through_youtube_auth,
                 only: %i(update_email)
-  before_action :protect, only: %i(show home)
+  # before_action :protect, only: %i(show home)
```
yachtcaptain23: Would you uncomment?

PR #985, opened 2018-06-15
Improve currency and wallet management

### LINKED ISSUE #768: Display settlement information on the dashboard

### LINKED ISSUE #834: Automatically create uphold BAT card for new publishers

### LINKED ISSUE #835: Have publishers explicitly set their default currency as part of uphold attachment

### LINKED ISSUE #997: Update sprockets to at least 3.7.2
Security issue disclosed here: https://blog.heroku.com/rails-asset-pipeline-vulnerability

dgeb:
This PR builds off the work of @nvonpentz in #941 and @lgebhardt in #970. It takes a more dynamic approach to communicate changes and upcoming redirects to the user than #941. 

### Default currency management

Much of the description that follows was written by @nvonpentz in #941:

This code will automatically prompt all users that have verified with uphold, and have completed the uphold sign up (`authorized`) with the confirm default currency screen. The user will select from a list of possible currencies, and hit confirm. The default currency will be recorded and `default_currency_updated_at` will be set.

If the publisher has the correct scope the dialog will close and the dashboard will be updated. If the publisher does not have the correct scope, they will be shown a notification that they are about to be redirected to the uphold authorization endpoint. After they are redirected and provide authorization at Uphold, they will be redirected to Publishers#home.

Publishers#home will then create the card if their confirmed currency is not available on Uphold, but is a possible currency. If BAT is not an available currency, Publishers#home will create that card as well (assuming they did not confirm this as their default currency). Publishers#home will also update eyeshade to use the confirmed default currency.

screen shot 2018-06-19 at 8 44 52 am

screen shot 2018-06-19 at 8 45 19 am

### Wallet management

Much of the description that follows was written by @lgebhardt in #970:

This creates a new javascript wallet class to handle the wallet logic on the client. This wallet is now returned by the balance class and handles contribution calculations and the last settlement date. This goes a good ways towards allowing for the asynchronous fetching of the wallet data (#838), but because some of the wallet info is still needed by the server it will not result in the desired speedup.

It also updates the Your Uphold Wallet section to match #835, which is an updated version of the UI from #768. 

The default currency drop down has been replaced with a Change link that triggers the default currency selection modal.

screen shot 2018-06-19 at 8 44 12 am

---

Closes #970
Closes #941 
Resolves #768 
Resolves #834 
Resolves #835
Resolves #997 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


test/helpers/publishers_helper_test.rb
```diff
@@ -127,6 +127,43 @@ def initialize(wallet_json:)
     assert_equal "Unavailable", publisher_humanize_balance(publisher, "USD")
   end
 
+  test "publisher_possible_currencies should an array of possible currencies" do
```
lgebhardt: Missing a word (be)? "publisher_possible_currencies should an array of possible currencies" -> "publisher_possible_currencies should be an array of possible currencies"

dgeb:
Currently working to fix the dialog to dynamically populate the options in the currency list from the wallet. This will ensure that the currency list is up to date when the wallet is loaded _after_ the page.

app/views/publishers/home.html.slim
```diff
@@ -87,22 +98,44 @@ noscript
...
+                span= ")"
+            .last-deposit-date.field
+              label= t ".uphold.last_deposit_date"
+              .value.deposit-date#last_deposit_date= publisher_humanize_last_settlement_date(current_publisher)
+          .dashboard-panel--content#last_settlement class=last_settlement_class(current_publisher)
+            .last-deposit
+              = t ".uphold.last_deposit"
+              .amounts
+                .bat
+                  - last_depost_balance = publisher_humanize_last_settlement(current_publisher, "BAT")
```
yachtcaptain23: Nit picking: This mispelling

PR #998, opened 2018-06-19
Allow admins to search by channel

### LINKED ISSUE #974: Search by Channel

nvonpentz:
Resolves #974 

This PR uses a single search bar for searching by publisher and channel. 

I also removed an unused route.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/admin/publishers_controller.rb
+++ b/app/controllers/admin/publishers_controller.rb
@@ -20,25 +20,25 @@ def sql(query)
                              INNER JOIN site_channel_details 
                                      ON site_channel_details.id = channels.details_id 
                                         AND channels.details_type = 'SiteChannelDetails' 
-                                        AND site_channel_details.brave_publisher_id LIKE '%#{query}%' 
+                                        AND site_channel_details.brave_publisher_id ILIKE '%#{query}%' 
                       UNION ALL 
                       SELECT channels.* 
                       FROM   channels 
                              INNER JOIN youtube_channel_details 
                                      ON youtube_channel_details.id = 
                                         channels.details_id 
-                                        AND youtube_channel_details.title LIKE '%#{query}%' 
+                                        AND youtube_channel_details.title ILIKE '%#{query}%' 
                       UNION ALL 
                       SELECT channels.* 
                       FROM   channels 
                              INNER JOIN twitch_channel_details 
                                      ON twitch_channel_details.id = channels.details_id 
-                                        AND twitch_channel_details.NAME LIKE '%#{query}%') 
+                                        AND twitch_channel_details.NAME ILIKE '%#{query}%') 
                                        c 
                    ON c.publisher_id = publishers.id
     UNION ALL
     SELECT publishers.id
     FROM publishers
-    WHERE publishers.email LIKE '%#{query}%'
-          OR publishers.name LIKE '%#{query}%'}
+    WHERE publishers.email ILIKE '%#{query}%'
+          OR publishers.name ILIKE '%#{query}%'}
 end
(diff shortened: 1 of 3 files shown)


app/controllers/admin/publishers_controller.rb
```diff
@@ -2,14 +2,43 @@ class Admin::PublishersController < AdminController
...
+                             INNER JOIN twitch_channel_details 
+                                     ON twitch_channel_details.id = channels.details_id 
+                                        AND twitch_channel_details.NAME LIKE '%#{query}%') 
+                                       c 
+                   ON c.publisher_id = publishers.id
+    UNION ALL
+    SELECT publishers.id
+    FROM publishers
+    WHERE publishers.email LIKE '%#{query}%'
+          OR publishers.name LIKE '%#{query}%'}
```
yachtcaptain23: Would you use `ILIKE` on `name` for case-insensitive match? Same with the twitch channels and youtube titles.

app/controllers/admin/publishers_controller.rb
```diff
@@ -2,14 +2,43 @@ class Admin::PublishersController < AdminController
...
     @publishers = Publisher
     if params[:q].present?
-      @publishers = @publishers.where("email ILIKE :email OR name ILIKE :name",
-                      {
-                        email: "%#{params[:q]}%",
-                        name: "%#{params[:q]}%"
-                      }
-                    )
+      # Returns an ActiveRecord::Relation of publishers for pagination
+      @publishers = Publisher.where("publishers.id IN (#{sql(params[:q])})").distinct
```
yachtcaptain23: Instead of doing Publisher.where("publishers.id IN (#{sql(params[:q])})") which is essentially `select publishers.* from publishers where publishers.id in <YOUR SQL STATEMENT>`, you can simplify by doing `ActiveRecord::Base.connection.execute("<YOUR SQL CODE>")`

nvonpentz: `ActiveRecord::Base.connection.execute("<YOUR SQL CODE>")` would return a `PG::Result`, instead of an `ActiveRecord::Relation`, and we wouldn't be able to paginate.

`@publishers = Publisher.where("publishers.id IN (#{sql(params[:q])})").distinct` creates a relation of publishers sourced from the list of publisher ids returned by the `sql` method and allows for pagination.

I'm not sure if there is a better way of doing this.  I opted not to use Arel since this was quicker , and I'm not convinced it's more efficient.  Let me know

yachtcaptain23: Ah okay that makes sense.

app/controllers/admin/publishers_controller.rb
```diff
@@ -2,14 +2,43 @@ class Admin::PublishersController < AdminController
...
-
     @publishers = @publishers.paginate(page: params[:page])
   end
 end
+
+private
+
+# Returns an array of publisher ids that match the query
+def sql(query)
+  %{SELECT publishers.id
```
yachtcaptain23: I'm nitpicking, but would you change to `SELECT DISTINCT publishers.id` here? The `DISTINCT` at the top runs it at the top layer.

nvonpentz: 💯

nvonpentz: Actually, I think we want this to run at the top.  Assuming we don't, if there is a publisher has a channel that matches the query AND has a name/email that matches the query, I think we'd display them twice in the results.

PR #1018, opened 2018-06-22
Feature/admin suspended users

yachtcaptain23:
Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/controllers/admin/publishers/publisher_status_updates_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+    get_publisher
+    @publisher_status_updates = @publisher.status_updates
+  end
+
+  def create
+    @publisher.status_updates.create(status: params[:publisher_status])
+    @publisher.reload
+    flash[:notice] = "Updated publisher's status to #{@publisher.inferred_status}"
+    redirect_to admin_publisher_path(id: @publisher.id)
+  end
```
nvonpentz: Should we log the action and send it to slack?

I'm doing this in the admin-approve PR, and I anticipate we'll need to do it elsewhere, so it might make sense to have a generalized service.

yachtcaptain23: Nah it's already traceable in the database. I don't want to bloat the code.

PR #1030, opened 2018-06-26
FAQ page and admin management page

### LINKED ISSUE #962: FAQ Page
- [ ] Use the same UI as settlements
- [ ] Database for storing question + answer, which will then interface with Admin. Questions should fall under a type (e.g. Payment).
- [ ] Part of the dropdown on the top-right side of the dashboard, which will be a link to FAQ.

lgebhardt:
This adds an FAQ page. The page is accessible whether you are logged in or not. The header changes depending on the user's log-in state.

There are links to the FAQs page in the user menu and from the landing page. The links to the FAQ page can be disabled by setting the ENV variable `HIDE_FAQS`. The landing page FAQ link was moved below the login button, and the TOS was added next to it.

FAQs are organized by Categories and each category can have multiple FAQs. Categories have a rank, which is used to control sorting on the FAQ page. FAQs also have rank to control sort order under categories. FAQs can be marked as Published, which controls whether they appear on the public FAQ page.

The admin page is a bit busy since it shows ranks and the published state, but I think having the info makes setting the order easier. Ideally we could have a set of buttons to allow the order to be set in a more intuitive manner, but that could be saved for another day.

Both Questions and Answers support markdown (and html) using the redcarpet gem. I expect this to be used for simple formatting and not complex page design. There isn't a toolbar, so some training for admins may be needed. 

Closes #962 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


yachtcaptain23:
Looks good. I think we need to update a link to include the FAQ view. I couldn't find the link that was used to get to FAQ.

yachtcaptain23:
Looks good. We just security review on `redcarpet`

PR #1047, opened 2018-06-29
WIP: Introduce dashboard charts

dgeb:
This introduces the dashboard charts described in https://github.com/brave-intl/publishers/issues/542

* A donut chart of contributions
* A stacked bar chart of monthly deposits

These charts are written to be customizable, but still require data in a certain schema, which is documented with examples. A common array of channels and corresponding colors can be passed, as well as a number of options related to the display.

The charts themselves are rendered with D3.

An example data set is used to explore usage of the charts on the dashboard. To move beyond this WIP, we’ll need to invoke rendering based on live data coming from eyeshade.

In order to enable display of these charts on the dashboard, visit `/publishers/home?charts=true` with the temporary query parameter.

screen shot 2018-06-29 at 3 39 05 pm

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/assets/stylesheets/pages/home.scss
+++ b/app/assets/stylesheets/pages/home.scss
@@ -266,6 +266,10 @@
         background-repeat: no-repeat;
         background-position: top right;
       }
+
+      &.with-charts {
+        padding: 30px 30px 30px 40px;
+      }
     }
     &--header {
       margin: 0 0 0 -50px;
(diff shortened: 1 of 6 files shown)


yachtcaptain23:
Effort will continue in https://github.com/brave-intl/publishers/pull/1080

PR #1064, opened 2018-07-10
Tell publishers why their site channel failed to verify

### LINKED ISSUE #711: Adding Site Channel: Missing way to correct "Channel could not be verified!"
Since #674/#655 if the user enters the token wrong they will see "Channel could not be verified!" with no way to "reenter" the channel to either choose a new method or check the instructions or token:

screen shot 2018-03-12 at 10 59 22 am

I propose we restore the previous behavior where clicking the channel will take the user to viewing the channel where they last left off (when they clicked verify).

From #655: "When an error has been encountered during domain verification, display that error for the channel and provide actionable steps. These scenarios will be spelled out further in a separate issue." I'm not sure how this is to be implemented, but it could provide an alternative way for the user to correct the problem instead of the above solution.

nvonpentz:
Resolves #711 

**This PR includes changes to the SiteChannelVerifier that should be reviewed and tested closely.**

How channels in various states appear on the dashboard:
image

Publishers can click "what happened" for unverified sites and see an explanation for why the verification failed.  This information is stored in the `verification_details` and is set by the SiteChannelVerifier:
image

After they click "verify" or "try again" from the dashboard, the modal will pop up with an explanation and a suggestion on how to fix:
image

When this PR is merged into production, an admin should replace all site channel details with verification_method `support_queue` to `nil`, as well as replace all channel details with verification_status `started` with `nil` or `failed`.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/helpers/channels_helper.rb
```diff
@@ -64,12 +64,65 @@ def channel_verification_status(channel)
...
+  def should_display_verification_token?(channel)
+    return false if channel.verified? || channel.details_type != "SiteChannelDetails"
+    ["no_txt_records", "token_incorrect_dns", "token_not_found_dns"].include?(channel.verification_details)
+  end
+
+  def capitalize_first_letter(str)
+    return if str.nil?
+    str[0] = str[0].capitalize
+    str
+  end
```
nvonpentz: This method allows us to use the translations in `en.helpers.channels.verification_failure_explanation.*` either as a standalone sentence (dashboard) or a the second part of a sentence (verification failed modal) without having to duplicate the translations.

app/views/application/_icon_circled_x.html
```diff
@@ -1,4 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="42" height="42" viewBox="0 0 42 42">
+<svg xmlns="http://www.w3.org/2000/svg" width="42" height="42" viewBox="0 0 42 42" class="icon-circled-x">
```
nvonpentz: Adding a class here allows us to dynamically size the SVG with CSS

app/views/site_channels/verification_public_file.html.slim
```diff
@@ -4,6 +4,9 @@
     publisher: current_publisher \
   }
 
+script type="text/html" id="verification_failed_modal_wrapper" data-open-verification-failed-modal="#{current_channel.verification_failed?}"
```
nvonpentz: This script (here and on all of the verification pages) will load the modal if the channel's current `verification_status` has failed.  This has the unfortunate side effect of displaying the modal immediately after a publisher selects a new verification method.  I think ideally we show the failed verification modal only if the the failed attempt uses the method of the page the publisher is currently on.  I'll think about how to do this...

app/assets/stylesheets/pages/home.scss
```diff
@@ -657,12 +660,62 @@
...
+        display: flex;
+        background: url(asset-path("icn-warning@1x.png")) no-repeat top left / 24px 24px;
         padding: 0 0 12px 38px;
         font-weight: bold;
         .btn {
-          margin: -8px 0 0 12px;
           vertical-align: top;
+          margin-left: 10px;
+        }
+        .verification-failed {
```
yachtcaptain23: We may want to rename these to BEM style conventions.

app/helpers/channels_helper.rb
```diff
@@ -64,12 +64,65 @@ def channel_verification_status(channel)
...
+        I18n.t("helpers.channels.verification_failure_cta.generic", support_email: Rails.application.secrets[:support_email])
+    end
+  end
+
+  def should_display_verification_token?(channel)
+    return false if channel.verified? || channel.details_type != "SiteChannelDetails"
+    ["no_txt_records", "token_incorrect_dns", "token_not_found_dns"].include?(channel.verification_details)
+  end
+
+  def capitalize_first_letter(str)
```
yachtcaptain23: You can get rid of this method and just use `str.capitalize` according to https://stackoverflow.com/questions/3724913/capitalize-first-letter-in-ruby

config/locales/en.yml
```diff
@@ -85,9 +85,33 @@ en:
       title: Frequently Asked Questions
   helpers:
     channels:
-      generic_verification_failure: Channel could not be verified!
       verification_in_progress: Verification in progress
-      verification_awaiting_admin_approval: Awaiting admin approval
+      verification_awaiting_admin_approval: Under review
+      verification_failure_what_happend: What happened
```
yachtcaptain23: Misspelling

app/helpers/channels_helper.rb
nvonpentz: sweet

nvonpentz: @yachtcaptain23 Actually the reason why I didn't use this method is that it doesn't preserve the capitalization of the tail of the string.

```
irb(main):001:0> s = "we could not find the token in any of the TXT records."
irb(main):002:0> s.capitalize
=> "We could not find the token in any of the txt records."
```

I think we want to preserver the capitalization of "TXT".  Let me know if you'd like to keep the capitalize_first_letter method or create duplicate translations, or something else.

app/javascript/publishers/home.js
```diff
@@ -347,6 +357,25 @@ document.addEventListener('DOMContentLoaded', function() {
...
+
+  // Hide all verification failed bubbles when anywhere on DOM is clicked
+  document.body.addEventListener('click', function(event) {
+    for (var i=0; i<verificationFailureWhatHappenedElements.length; i++) {
+      // Do not hide if the clicked element is supposed to show the bubble
+      let e = verificationFailureWhatHappenedElements[i];
+      if (e === event.target) {
+        continue;
+      } else {
+        hideVerificationFailureWhatHappenend(e);
```
yachtcaptain23: Should be `hideVerificationFailureWhatHappened`

nvonpentz: happened is a tough one for me apparently

app/models/channel.rb
```diff
@@ -125,25 +119,30 @@ def verification_failed!(details = nil)
...
 
-  def verification_succeeded!(admin_approval)
-    if admin_approval
-      verification_status = 'approved_by_admin'
+  def verification_succeeded!(has_admin_approval)
+    if needs_admin_approval?
+      if has_admin_approval
+        verification_status = 'approved_by_admin'
+      else
+        raise VERIFICATION_RESTRICTION_ERROR
```
yachtcaptain23: I believe we want to append it to `errors` rather than raising an exception.

app/services/site_channel_verifier.rb
```diff
@@ -3,32 +3,31 @@
...
-    return false if channel.verification_awaiting_admin_approval? && !admin_approval
+    return false if channel.verification_awaiting_admin_approval? && !has_admin_approval
 
-    @verified_channel_id = verify_publisher_id
-
-    if verified_channel_id.blank?
-      channel.verification_failed!
+    verification_result = verify_site_channel
+    
+    if verification_result == false
```
yachtcaptain23: `if verification_result.nil?` is more conventional

app/services/site_channel_verifier.rb
```diff
@@ -3,32 +3,31 @@
...
-
-    update_verified_on_channel
-
-    unless channel.id == verified_channel_id
-      channel.verification_failed!
+    elsif channel.needs_admin_approval? && !has_admin_approval
+      channel.verification_awaiting_admin_approval!
       return false
+    else
+      channel.verification_succeeded!(has_admin_approval)
```
yachtcaptain23: You should check the errors resulting from `verification_succeeded!`

app/services/site_channel_verifier.rb
```diff
@@ -37,125 +36,107 @@ def perform
...
-      else
-        verified_channel.verification_awaiting_admin_approval!
-        InternalMailer.channel_verification_approval_required(channel: verified_channel).deliver_later
-        SlackMessenger.new(message: "*Admin Action Required:* *#{verified_channel.publication_title}* on the restriction list verified by #{verified_channel.publisher.owner_identifier}; id=#{verified_channel.details.channel_identifier}").perform
-        return
-      end
-    end
-
-    verified_channel.verification_succeeded!(admin_approval)
+    verified_channel.verification_succeeded!(has_admin_approval)
```
yachtcaptain23: Same here I believe

app/views/publishers/home.html.slim
```diff
@@ -210,32 +210,42 @@ script type="text/html" id="confirm_default_currency_modal_wrapper"
...
+              .channel-summary
+                h3= channel.publication_title
+                .channel-details
+                  .added-date
+                    = t ".channel.added", date: channel.created_at.to_date.iso8601
+            - elsif channel.verification_failed?
               .channel-status.float-right
-                = channel_verification_details(channel)
-                = link_to(t(".channel.try_again"), channel_next_step_path(channel), class: "btn btn-primary")
+                .verification-failed
```
yachtcaptain23: IMO we should start breaking up the `home.html.slim` and merge as part of a `dashboard_partials` or `home_partials` folder just to make the folder not bloated.

app/services/site_channel_verifier.rb
nvonpentz: `verify_site_channel` returns either returns `true` if the verification is successful, or `false` if not.  If it returns `false` it also sets `@verification_details` to the error, which is then later presented to the user.  So `verification_result.nil?` will always return `false` under this design.

We could have `verify_site_channel` return `nil` if the verification was successful and return the value of `@verification_details` if it wasn't.

config/locales/en.yml
```diff
@@ -85,9 +85,31 @@ en:
...
+        domain_not_found: "we could not find your domain."
+        connection_failed: |
+          we could not find the public file at https://%{domain}/.well-known/brave-payments-verification.txt.
+        too_many_redirects: "we were redirected too many times to verify your domain. "
+        no_txt_records: "there are no TXT records on your domain's DNS. "
+        token_incorrect_dns: "the token in your TXT records is incorrect. "
+        token_not_found_dns: "we could not find the token in any of the TXT records. "
+        token_not_found_public_file: "we could not find the token in your public file. "
+        no_https: "HTTPS is not enabled. "
+        generic: "of an unkown reason. "
```
yachtcaptain23: Misspelling.

config/locales/en.yml
```diff
@@ -85,9 +85,31 @@ en:
...
+      verification_failure_cta:
+        domain_not_found: "Please check if the domain is correct."
+        connection_failed: "Please make sure the file exists and is publicly available."
+        too_many_redirects: "Please reduce the number of redirects."
+        no_txt_records: "Please make sure you have entered the token in your DNS records:"
+        token_incorrect_dns: "Please make sure the token in your DNS entry is correct."
+        token_not_found_dns: "Please add this TXT entry to your DNS records:"
+        token_not_found_public_file_html: "Please make sure the verification token is present in the public file %{domain}/.well-known/brave-payments-verification.txt."
+        no_https: "Please enabled HTTPS on your domain, or choose another verification method."
+        generic: "Please contact %{support_email}."
```
yachtcaptain23: image

To be honest, the CTA is confusing since there's 2 conflicting CTAs (1. please email us and 2. Retry Verification). I think we should simplify this somehow, or even adjust the wording to say "If this verification failure continues to occur, please reach us at <email>".

PR #1074, opened 2018-07-16
Verified and non-verified publishers created_at per day

### LINKED ISSUE #1065: [Admin] API to measure # of account signups per day

yachtcaptain23:
Closes #1065

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/api/stats_controller.rb
+++ b/app/controllers/api/stats_controller.rb
@@ -12,7 +12,7 @@ class Api::StatsController < Api::BaseController
     render(json: fill_in_blank_dates(result).to_json, status: 200)
   end
 
-  def verified_signups_per_day
+  def email_verified_signups_per_day
     sql =
     """
       select created_at::date, count(*)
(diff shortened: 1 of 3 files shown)


app/controllers/api/stats_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      from publishers
+      where role = 'publisher'
+      group by created_at::date
+      order by created_at::date
+    """
+    result = ActiveRecord::Base.connection.execute(sql).values
+    render(json: fill_in_blank_dates(result).to_json, status: 200)
+  end
+
+  def verified_signups_per_day
```
nvonpentz: It might be better call this email-verified publishers, since there is no concept of a 'verified publisher' since the multichannel update.  We could also map this to the publisher status.  A publisher is 'onboarding' once they have verified their email address. https://github.com/brave-intl/publishers/blob/e47b83a994d764643f4acecb0190d7bbbc844ddd/app/models/publisher_status_update.rb#L7

PR #1080, opened 2018-07-18
Dashboard charts

yachtcaptain23:
This introduces the dashboard charts described in #542

* A donut chart of contributions
* A stacked bar chart of monthly deposits

These charts are written to be customizable, but still require data in a certain schema (i.e. format). A common array of channels and corresponding colors can be passed, as well as a number of options related to the display.

The charts themselves are rendered with D3.

An example data set is used to explore usage of the charts on the dashboard. To move beyond this WIP, we’ll need to invoke rendering based on live data coming from eyeshade.

In order to enable display of these charts on the dashboard, visit `/publishers/home?charts=true` with the temporary query parameter.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/views/publishers/dashboard_partials/_no_donut_chart.html.slim
```diff
@@ -0,0 +1 @@
+h4 No contributions yet
```
CoryMcDonald: I18n

app/views/publishers/home.html.slim
```diff
@@ -28,64 +28,28 @@ script type="text/html" id="confirm_default_currency_modal_wrapper"
...
-            = t ".add_channel_cta"
-    .row
+        .dashboard-panel--wrapper
+          h4.dashboard-panel--header#dashboard_brave_header
+            .row
+              .col.col-md-8
+                = t ".brave_fan_contributions"
+              .col.col-md-4#dashboard_info_box
+                = image_tag("small-calendar@1.5x.png")
+                = " #{Date.today.beginning_of_month.strftime('%m/%d')} - #{Date.today.strftime('%m/%d')}"
```
CoryMcDonald: I would argue that this should most likely be in a helper, and that the strftime should be internationalized because different countries display their dates in different ways.

diracdeltas:
security comment here: https://github.com/brave/internal/issues/395#issuecomment-409701571

yachtcaptain23:
Closing the PR to minimize noise: Continuing on work after 1.1 release.

PR #1086, opened 2018-07-25
Stats for no-script users

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/api/stats_controller.rb
+++ b/app/controllers/api/stats_controller.rb
@@ -71,18 +71,20 @@ class Api::StatsController < Api::BaseController
 
   def javascript_enabled_usage
 
-    active_users_with_javascript_enabled = Publisher
-      .distinct
-      .joins("inner join channels on channels.publisher_id = publishers.id")
-      .where.not(javascript_last_detected_at: nil)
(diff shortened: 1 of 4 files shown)


app/controllers/api/stats_controller.rb
```diff
@@ -26,6 +26,18 @@ def email_verified_signups_per_day
...
+    active_users_with_javascript_disabled = Publisher.distinct.joins("inner join channels on channels.publisher_id = publishers.id").where(javascript_last_detected_at: nil).where("publishers.created_at > ?", "2018-06-22 15:48:40").count
+    render(
+      json: {
+        active_users_with_javascript_enabled: active_users_with_javascript_enabled,
+        active_users_with_javascript_disabled: active_users_with_javascript_disabled,
+      },
+      status: 200
+    )
+  end
+
```
nvonpentz: According to our slack conversations I think the start date was a few days earlier `Tue, 19 Jun 2018 15:51:51 PDT -07:00`.

I think this will consider users who haven't logged in recently as no script users.

This will sort publishers into two buckets, those who have never used javascript on the site, and those who have used javascript at least once.  This might not be the ideal metric, but I don't think getting more detailed information is a blocker.

Also, could we break this into multiple lines on each `.where` clause?

PR #1087, opened 2018-07-26
Upgrade/rails5.2

### LINKED ISSUE #1083: Upgrade to Rails 5.2
Since we're making good headway on progress, let's see if we can get to Rails 5.2 to use the integrated features to support S3 upload.

yachtcaptain23:
Resolves #1083 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


Gemfile
```diff
@@ -112,14 +118,14 @@ gem "u2f", "~> 1.0"
 # One-time passwords for 2fa
 gem "rotp", "~> 3.3"
 
-gem 'webpacker', '~> 3.2'
-
-# YouTube API client
-gem 'yt'
```
nvonpentz: I think we need the `yt` gem

Gemfile
```diff
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 source "https://rubygems.org"
 
+gem 'actionmailer', '~> 5.2.0'
+gem 'actionpack', '~> 5.2.0'
+gem 'activemodel', '~> 5.2.0'
+gem 'activerecord', '~> 5.2.0'
+gem 'activesupport', '~> 5.2.0'
+
```
nvonpentz: Don't these automatically come with rails?

Procfile.dev
```diff
@@ -1,2 +1,3 @@
 web: bundle exec puma -C config/puma.rb -e ${RACK_ENV:-development}
-worker: bundle exec sidekiq -C config/sidekiq.yml -e ${RACK_ENV:-development}
\ No newline at end of file
+worker: bundle exec sidekiq -C config/sidekiq.yml -e ${RACK_ENV:-development}
+webpacker: ./bin/webpack-dev-server
```
nvonpentz: Do we need to add this to the production procfile?

Do you know why we need to specify this when in the past we didn't?

app/models/channel.rb
```diff
@@ -168,7 +168,7 @@ def update_last_verification_timestamp
   def should_register_channel_for_promo
     promo_running = Rails.application.secrets[:active_promo_id].present?  # Could use PromosHelper#active_promo_id
     publisher_enabled_promo = self.publisher.promo_enabled_2018q1?
-    promo_running && publisher_enabled_promo && verified_changed? && verified
+    promo_running && publisher_enabled_promo && saved_change_to_verified? && verified
```
nvonpentz: Is this change related to rails upgrade?

config/environments/development.rb
```diff
@@ -21,7 +21,7 @@
   # Enable/disable caching. By default caching is disabled.
   if Rails.application.secrets[:redis_url]
     config.action_controller.perform_caching = true
-    config.cache_store = :redis_store
+    config.cache_store = :redis_cache_store, { url: Rails.application.secrets[:redis_url] }
```
nvonpentz: What does this change do?

app/models/channel.rb
yachtcaptain23: Yes, it's related to https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Dirty.html

From what I understand, `changed?` was confusing because at-a-glance, it could be a change to the unsaved version or to the saved version.

config/environments/development.rb
yachtcaptain23: https://www.engineyard.com/blog/rails-5-2-redis-cache-store

TL;DR Rails comes bundled with their own version of https://github.com/redis-store/redis-rails

Procfile.dev
yachtcaptain23: Not sure if we need to add to the production Procfile. Webpacker detects changes to the javascript code so you don't need to refresh the page: https://webpack.js.org/concepts/hot-module-replacement/

However, since the assets are compiled during build time (and assets shouldn't change) I don't believe we're supposed to include webpacker. I'm not 100% confident though, so I'll test against staging-2.

Gemfile
yachtcaptain23: They do, but upgrading dependencies via `bundler` wasn't elegant.

I'll remove these lines, and I think we should be fine as long as these changes are remarked in `Gemfile.lock`.

PR #1096, opened 2018-08-10
[WIP] Use new eyeshade balance endpoint for wallet information

### LINKED ISSUE #1095: Use new balances API for the Eyeshade wallet

nvonpentz:
Resolves #1095 

This PR utilizes the new transaction table `/balances` API for the channel & owner balances.  This PR should not change anything from the users perspective (except for potentially speeding up the site) - only backend changes.

Now one single request is made to get the channel balances for every owner, instead of one per channel.

Significant modifications we're made to how we store and retrieve wallet information.  We no longer have an Eyeshade::Balance class as this abstraction forced a lot of redundancy (rates we're duplicated for each channel and for the owner) and because the Eyeshade::Balance was mostly there to handle conversion from BAT to probi, and we no longer deal in probi.

In addition this PR removes the JsonWalletBuilder class, instead option to simply return the JSON representation of the Eyeshade::Wallet.  The motivation for this change was to keep wallet representations as consistent as possible so changes in the future do not require.

These changes to the ruby representation of the wallet impact the js representation, thus changes were made to the wallet.js class and tests.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


nvonpentz:
Closing because this PR introduces larger changes to the wallet classes (ruby and js) that are not necessary at this time.

PR #1099, opened 2018-08-14
Use new eyeshade balance endpoint for wallet information 

### LINKED ISSUE #1095: Use new balances API for the Eyeshade wallet

nvonpentz:
Resolves #1095 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/services/publisher_balance_getter.rb
+++ b/app/services/publisher_balance_getter.rb
@@ -1,0 +1,55 @@
+# Used by the PublisherWalletGetter to retrieve balances
+class PublisherBalanceGetter < BaseApiClient
+  attr_reader :publisher
+
+  def initialize(publisher:)
+    @publisher = publisher
+  end
+
+  def perform
+    return [] if publisher.channels.verified.empty?
+    return perform_offline if Rails.application.secrets[:api_eyeshade_offline]
+
+    channel_balances_response = connection.get do |request|
+      request.headers["Authorization"] = api_authorization_header
+      request.url("/v1/balances?account=#{URI.escape(publisher.owner_identifier)}#{URI.escape(channels_query_string)}")
+    end
+
+    channel_balances = JSON.parse(channel_balances_response.body)
+
+    # Eyeshade returns an empty response for verified channels
+    # with no balance, so we must fill in 0.00 balances
+
+    # Find verified channel IDs with no balance
+    verified_channel_ids = []
+    publisher.channels.verified.each {|channel| verified_channel_ids.push(channel.details.channel_identifier)}
+
+    channel_balances.each do |channel_balance|
+      channel_identifier = channel_balance["account"]
+      verified_channel_ids.delete(channel_identifier) if verified_channel_ids.include?(channel_identifier)
+    end
+    # Set balances to 0.00
+    channel_balances += verified_channel_ids.map { |verified_channel_id| { "account" => "#{verified_channel_id}", "balance" => "0.00" } }
+  
+    channel_balances
+  end
+
+  def perform_offline
+    @publisher.channels.verified.map { |channel| { "account" => "#{channel.details.channel_identifier}", "balance" => "#{rand(0..3000)}"} }
+  end
+
+  private
+
+  def channels_query_string
+    return "" if publisher.channels.verified.count == 0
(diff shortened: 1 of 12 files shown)


app/services/publisher_wallet_getter.rb
```diff
@@ -16,26 +16,24 @@ def perform
...
-      end
-    end
-
     wallet_hash = JSON.parse(wallet_response.body)
 
-    channel_hash = {}
-    channel_responses.each do |identifier, response|
-      channel_hash[identifier] = JSON.parse(response.body)
+    rates = wallet_hash["rates"]
+    if wallet_hash && wallet_hash["contributions"] && wallet_hash["contributions"]["currency"]
```
yachtcaptain23: I think you can do wallet_hash.dig("contributions", "currency") for readability

https://stackoverflow.com/questions/19115838/how-do-i-use-the-fetch-method-for-nested-hash

For max readability: `currency = wallet_hash.dig("contributions, "currency") || publisher.default_currency`

app/services/publisher_balance_getter.rb
```diff
@@ -0,0 +1,78 @@
...
+    verified_channel_ids = []
+    publisher.channels.verified.each {|channel| verified_channel_ids.push(channel.details.channel_identifier)}
+
+    channel_balances.each do |channel_balance|
+      channel_identifier = channel_balance["account"]
+      verified_channel_ids.delete(channel_identifier) if verified_channel_ids.include?(channel_identifier)
+    end
+
+    # Set balances to 0.00
+    verified_channel_ids.each do |id|
```
yachtcaptain23: You could replace the block by doing

verified_channel_ids.map do |verified_channel_id| 'account' => '#{verified_channel_id}', 'balance' => "0.00" } end

app/services/publisher_balance_getter.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      channel_balances.push(channel_balance)
+    end
+
+    channel_balances
+  end
+
+  def perform_offline
+    channels_json = []
+    @publisher.channels.verified.each do |channel|
```
yachtcaptain23: And the same here

app/services/publisher_balance_getter.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    channels_json
+  end
+
+  private
+
+  def channels_query_string
+    return "" if publisher.channels.verified.count == 0
+
```
yachtcaptain23: You can simplify this block by doing:

publisher.channels.verified.select { |channel| "&account=#{channel_details.channel_identifier} }.reduce(:+)

nvonpentz: Hm, I don't think so since `publisher.channels.verified.select { |channel| "&account=#{channel_details.channel_identifier} }` returns an array of channels

app/services/publisher_balance_getter.rb
nvonpentz: was able to do so with
```
    @publisher.channels.verified.pluck(:id).map { |verified_channel_id| {"account" => "#{verified_channel_id}", "balance" => "#{rand(0..3000)}"} }
```

app/services/publisher_balance_getter.rb
nvonpentz: I'm not sure a `.map` makes sense here since we are pushing each result of the loop into `channel_balances`, which is an array that already has contents prior to the loop.

yachtcaptain23: `channel_balances += verified_channel_ids.map { |verified_channel_id| { 'account' => '#{verified_channel_id}', 'balance' => "0.00" } }`

app/services/publisher_balance_getter.rb
nvonpentz: Yeah, but channel isn't implicitly convertible to string.  The above gives this error:

```NoMethodError Exception: undefined method `+' for #<Channel:0x00007fac383456e8>```

We could do this instead
```publisher.channels.verified.map { |channel| "&account=#{channel.details.channel_identifier}" }.reduce(:+)```, though I'm not sure this is more readable than the original.

PR #1102, opened 2018-08-15
PublisherWalletGetter and PublisherBalanceGetter Improvements

### LINKED ISSUE #1095: Use new balances API for the Eyeshade wallet

nvonpentz:
Resolves #1095 

This PR cleans up the logic in the PublisherBalanceGetter and PublisherWalletGetter.  The PublisherBalanceGetter no longer sums the individual channel balances, and instead uses the owner account balance sent from eyeshade.  In addition, the PublisherBalanceGetter will also fill in empty balances for all accounts (owner and channel), so the PublisherWalletGetter can always expect a complete response regardless of what eyeshade returns.

The changes in PublisherWalletGetter are mostly superficial variable and method name changes. 

This PR also allows dev admins to toggle a `SHOULD_USE_TRANSACTION_TABLE` envar to easily switch from the old and new /balances endpoint.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/services/publisher_wallet_getter.rb
```diff
@@ -18,19 +18,37 @@ def perform
...
+        channel_responses[identifier] = connection.get do |request|
+          request.headers["Authorization"] = api_authorization_header
+          request.url("/v2/publishers/#{URI.escape(identifier)}/balance")
+        end
+      end
+
       channel_hash = {}
+      channel_responses.each do |identifier, response|
+        channel_hash[identifier] = JSON.parse(response.body)
+      end
```
nvonpentz:
```
    else
      channel_responses = {}
      publisher.channels.verified.each do |channel|
        identifier =  channel.details.channel_identifier
        channel_responses[identifier] = connection.get do |request|
          request.url("/v2/publishers/#{URI.escape(identifier)}/balance")
        end
      end

      channel_hash = {}
      channel_responses.each do |identifier, response|
        channel_hash[identifier] = JSON.parse(response.body)
      end
    end
```
Everything in this block was copy and pasted from the `PublisherWalletGetter` prior to any transaction table changes.

app/services/publisher_balance_getter.rb
```diff
@@ -56,6 +32,45 @@ def channels_query_string
...
+    # Find whether the owner account was included in the response
+    accounts_include_owner_account = false
+    accounts.each do |account|
+      if account["account_type"] == "owner"
+        accounts_include_owner_account = true
+      end
+    end
+
+    # Fill in missing channel accounts if needed
+    channel_ids_to_fill_in = (verified_channel_ids.map { |verified_channel_id| account_ids.include?(verified_channel_id) ? nil : verified_channel_id }).compact
```
yachtcaptain23: I think you can do `verified_channel_ids.select { |verified_channel_id| account_ids.include?(verified_channel_id) }`

PR #1103, opened 2018-08-17
Automatically generate monthly payout reports

nvonpentz:
On the 1st of the each month, the GeneratePayoutReportJob will run, query eyeshade for the balance for each verified channel.  If there is a balance, the channel id, amount, address and some other data is included in the report.  If a verified channel has a pending balance but is not connected to uphold, or needs to reconnected to uphold, the channel will not be included in the report, and the publisher will be sent an email.

During settlement week, a Brave admin will login to the admin dashboard and download the most recent report.  Brave admins also have the ability to generate one off reports.

image

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/controllers/admin/payout_reports_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    send_data payout_report.contents,
+      filename: payout_report_file_name(payout_report),
+      type: :json
+  end
+
+  private
+
+  def payout_report_file_name(payout_report)
+    created_at_str = payout_report.created_at.to_s # "2018-08-17 07:45:23 -0700"
```
CoryMcDonald: What if we refactored this to use [`strftime`](https://ruby-doc.org/core-2.2.0/Time.html#method-i-strftime)? I think this might be a little bit less code.

Something like?
`payout_report.created_at.strftime("%FT%H_%M_%S")`

nvonpentz: @CoryMcDonald Good idea, will update 👍

app/jobs/generate_payout_report_job.rb
```diff
@@ -0,0 +1,82 @@
...
+    total_amount_probi = 0
+    total_fees_probi = 0
+    num_payments = 0
+
+    publishers.find_each do |publisher|
+      publisher_has_unsettled_balance = false
+      wallet = publisher.wallet
+
+      publisher.channels.verified.each do |channel|
+        probi = wallet.channel_balances[channel.details.channel_identifier].probi
```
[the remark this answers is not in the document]
nvonpentz: Yes this should always be defined, see https://github.com/brave-intl/publishers/blob/0320aa7372ad3fde278d9e302f557dfc224d8b2b/app/services/publisher_balance_getter.rb#L37.  If not, there are larger problem problems and I'd rather payout generation fail.

app/views/admin/payout_reports/index.html.slim
```diff
@@ -0,0 +1,31 @@
+h1 Payout Reports
+table.display.table.table-bordered.table-striped.dynamic-table id="dynamic-table"
+  tr
+    th Created
```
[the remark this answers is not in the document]
nvonpentz: Typically yes, but these pages are for the internal (admin only) dashboard, so we don't need to.

app/views/admin/payout_reports/index.html.slim
```diff
@@ -0,0 +1,31 @@
...
+    th Download
+  tbody
+    - @payout_reports.each do |report|
+      tr.gradeX
+        td = report.created_at.strftime("%B %d, %Y")
+        td = link_to report.id,admin_payout_report_path(report.id)
+        td = report.num_payments
+        td = report.amount
+        td = report.fees
+        td = link_to "download", download_admin_payout_report_path(report.id)
```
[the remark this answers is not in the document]
nvonpentz: Same as above.

app/models/payout_report.rb
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  def encryption_key
+    Rails.application.secrets[:attr_encrypted_key]
+  end
+
+  def self.total_amount
+    amount = 0
+
+    PayoutReport.all.each do |report|
```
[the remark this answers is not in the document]
nvonpentz: Yep, will replace.  Thanks for your comments!

app/models/payout_report.rb
```diff
@@ -0,0 +1,26 @@
+class PayoutReport < ApplicationRecord
+  has_paper_trail
```
yachtcaptain23: I don't think we want to run paper trail on this

nvonpentz: @yachtcaptain23 why not?  I'm thinking we definitely want to track any changes made to a payout report after it's been generated.

app/controllers/admin/payout_reports_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  private
+
+  def assign_authority(report_contents)
+    report_contents = JSON.parse(report_contents)
+    
+    report_contents.each do |channel|
+      channel["authority"] = current_publisher.email # Assigns authority to admin email
+    end
```
nvonpentz: The payout_report.contents will have an authority of "GeneratePayoutReportJob".  However, whenever it is downloaded to an admin, we replace the "GeneratePayoutJob" with the admin's email.

app/mailers/publisher_mailer.rb
```diff
@@ -185,9 +185,9 @@ def unverified_domain_reached_threshold_internal(domain, email)
     )
   end
 
-  def verified_invalid_wallet(publisher, params)
```
nvonpentz: These `params` were being sent via the soon to be deprecated `PublisherNotifier`, which was used by eyeshade to send these emails when eyeshade was in charge of payout report generation.  The `params` don't ever seem to be used, so I'm removing them entirely.

app/models/publisher.rb
```diff
@@ -92,6 +92,10 @@ class Publisher < ApplicationRecord
     .where("uphold_updated_at < ?", UPHOLD_ACCESS_PARAMS_TIMEOUT.ago)
   }
 
+  scope :has_verified_channel, -> {
```
yachtcaptain23: would you rename to `verified_channels`? `has_verified_channel` implies a boolean return type as opposed to a collection

db/migrate/20180815211735_create_payout_report.rb
```diff
@@ -0,0 +1,14 @@
+class CreatePayoutReport < ActiveRecord::Migration[5.2]
+  def change
+    create_table :payout_reports, id: :uuid, default: -> { "uuid_generate_v4()" } do |t|
+      t.boolean  :final
+      t.decimal  :amount
```
yachtcaptain23: IIRC we should be defining the amount. The default might truncate the probi value. 

e.g. `:decimal, precision: 8, scale: 2` applies to USD (2 decimal points for string).

However, if it's in probi, shouldn't this be an integer?

app/models/publisher.rb
nvonpentz: how about `with_verified_channels`?

db/migrate/20180815211735_create_payout_report.rb
nvonpentz: Previously `amount` and `fees` were stored as BAT, not probi, but I changed them to be probi, and now they are stored as strings.

app/jobs/generate_payout_report_job.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      # Notify publishers that have money waiting, but will not will not receive funds
+      if publisher_has_unsettled_balance && should_send_notifications
+        if !publisher.uphold_verified? || wallet.address.blank?
+          PublisherMailer.wallet_not_connected(publisher).deliver_later
+        end
+      end
+    end
+
+    payout_report.update!(amount: total_amount_probi,
```
yachtcaptain23: Does this need to be `total_amount_probi.to_s`?

nvonpentz: @yachtcaptain23 I just stepped through it, and yes `total_amount_probi` is a FixNum right before the payout report is updated, but it doesn't seem to be a problem, and the FixNum is coerced to a string.

I can add `.to_s` to be clear if you'd like.

app/models/payout_report.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+    
+    total_amount_bat = total_amount_probi.to_d / 1E18
+    total_amount_bat
+  end
+
+  def self.total_fees
+    total_fees_probi = 0
+
+    PayoutReport.all.each do |report|
```
yachtcaptain23: One liner: `PayoutReport.sum { |payout_report| payout_report.fees.to_i } / 1E18`

app/models/payout_report.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  scope :final, -> {
+    where(final: true)
+  }
+
+  def encryption_key
+    Rails.application.secrets[:attr_encrypted_key]
+  end
+
+  def self.total_amount
```
yachtcaptain23: Would you use functional coding style here too?

PR #1105, opened 2018-08-20
Allow publishers to delete verified channels

nvonpentz:
Resolves #1104 
Blocked until #1103 lands

* Do not return early from the DeletePublisherChannelJob if the channel is verified

* Uncomment tests that were removed when verified channel removal was removed

* Rename `update_promo_server` -> `should_update_promo_server` for improved readability

* Refactor PromoChannelOwnerUpdater slightly

* Create PromoChannelOwnerUpdaterTests & new promo_enabled publisher fixture

Submitter Checklist:

- [ ] Submitted a [ticket](https://github.com/brave-intl/publishers/issues) for my issue if one did not already exist.
- [ ] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- [ ] Tagged reviewers.
- [ ] Integrated piwik/matomo (for code that adds new buttons).
- [ ] Addressed or ignored all brakeman warnings

Test Plan:

Reviewer Checklist:

Tests
- [ ] Adequate test coverage exists to prevent regressions

Security:
- [ ] No raw SQL -- Always prefer ActiveRecord query helpers ([more info on StackOverflow](https://stackoverflow.com/questions/41410752/rails-5-sql-injection#41452695))
- [ ] XSS is mitigated -- Avoid `html_safe` and `raw`; escape untrusted content from users and 3rd party APIs ([Rails XSS guide](https://brakemanpro.com/2017/09/08/cross-site-scripting-in-rails) and [OWASP XSS Prevention Cheat Sheet](https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet))

PR #1119, opened 2018-08-29
Allow publishers to challenge channel ownership

### LINKED ISSUE #376: Handle changes in domain registration
Related: https://github.com/brave-intl/bat-ledger/issues/42

When domain registrations change or expire, we should invalidate the Publisher record otherwise the old owner can continue to collect Brave Payments while they no longer own the domain.

To resume using Brave Payments again, the publisher needs to reverify their domain from scratch (#377).

### LINKED ISSUE #377: Allow publishers to reverify their domains and replace their old Publisher account
While discussing #364, we decided publishers with 2FA enabled who have lost their 2nd Factor can recover their accounts by reverifying their domains from scratch. Afterwards, we should invalidate and replace the old Publisher record and notify the previous Publisher email address.

Care should be taken to prevent active publishers from being harassed or otherwise DOS'd (e.g. anon shouldn't be able to disable Publisher accounts by spam retrying this flow). Adding to rack-attack rate limiter should be sufficient.

This reverify workflow can be reused in (#376); when domain registrations lapse or change Publishers will need to reverify as well.

Additional notes from #640:

1. A publisher requests to reverify the channels in their account (either for changes in ownership or for missing TOTP codes), and successfully demonstrates ownership
2. We send emails to the current owner alerting them of the activity
3. If there is no contest after a waiting period (3 days?), we allow the publisher to reset TOTP codes, or claim ownership of the channel(s)

* If there is a contest we manually mediate.
* If this results in an account without any channels or pending BAT, the account is removed.

### LINKED ISSUE #606: Channel removal service
service which can detach channels from Publishers (and Eyeshade) and delete them
to mimic the normal Remove Channel feature; for use by admins via console; and 

would be for fixing up YouTubers who did Google Login who were affected by multi-chan

please whitelist only YouTube Channel type. (should raise if running on SiteChannel, Twitch, etc)

### LINKED ISSUE #794: Channel Transfer - Dashboard UX
Channel dashboard UX for #377 is needed. This will look different for both sides of the transfer.

Here's the development placeholder for an owner attempting to transfer/claim an existing channel:

screen shot 2018-04-23 at 8 45 44 am

Here's the development placeholder the channel's current owner will see when someone else is attempting to transfer the channel:

screen shot 2018-04-23 at 9 38 13 am

We could also allow approval from the dashboard.

### LINKED ISSUE #795: Channel Transfer - Emails
To support #377 the emails the system sends are needed.

Possible emails might be:
  - To primary owner when channel is being claimed by another owner
  - To secondary owner after claiming channel
  - If rejected:
    - To secondary owner (claimer) that channel was rejected
    - To original owner that channel transfer was rejected.
  - If approved:
    - To secondary owner (claimer) that channel was approved
    - To original owner that channel transfer was approved and they no longer own this channel

### LINKED ISSUE #796: Channel Transfer - Rejection Process UX
How will the rejection process for channel transfers work?

There are two sources for rejecting a channel transfer:

- Emails will contain a link that needs to allow a channel transfer to be rejected without the recipient logging into the Publishers site.

- When the channel owner is logged into the publishers site channel transfers could be rejected from the dashboard.

nvonpentz:
This PR lets publishers attempt to verify any existing verified channel.  This is primarily useful for two use cases:

1. Changes in domain registration
2. Lost account credentials (email access or 2fa)

This recovery method is superior to others like answering recovery questions or using a recovery email address because it preserves the security properties that publishers expect when they use 2FA, i.e. restricting access to the account to only those who have the correct time based one time password.

The flow to recover a channel on when locked out of a publisher account:

1.  A publisher verifies their channel under account P1
2. The publisher looses access to their 2fa or email address to P1.
3. The publisher creates a new account P2, adds the channel, and completes the verification process which automatically creates a channel transfer request.  P1's is sent an email alerting them of the activity, and includes an email includes a reject transfer link which the user can click without logging in that cancels this process. No transfer approval link is sent out in to protect the honest user in the case that their emails has been compromised.  The rejection link is also accessible from the P1's publisher dashboard.
**Email sent to original owner**
image
**Dashboard of original owner**
image
**Dashboard of challenger**
image
4. Original owner clicks reject link in email or via dashboard:
**Email sent to challenger**
image
**Email sent to original owner**
image
5.  OR (instead of step 4), after 10 days without contest the channel removed from P1 and transferred to P2.
**Email sent to challenger**
image
**Email sent to original owner**
image

Resolves #376 
Resolves #377
Resolves #606 
Resolves #794 
Resolves #795 
Resolves #796 

Closes #1038 
Closes #1105 

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/controllers/channel_transfer_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  def reject_transfer
+    Channels::RejectChannelTransfer.new(channel: current_channel).perform
+    redirect_to home_publishers_path, notice: t("shared.channel_transfer_rejected")
+  end
+
+  private
+
+  def verify_token
+    @current_channel = Channel.find(params[:id])
+    if @current_channel.nil? || @current_channel.contest_token.empty? || @current_channel.contest_token != params[:token_id]
```
yachtcaptain23: I believe we should delete the token as a way of invalidating to prevent someone from trying to steal with multiple tokens.

app/mailers/publisher_mailer.rb
```diff
@@ -156,6 +156,114 @@ def unverified_domain_reached_threshold_internal(domain, email)
     )
   end
 
+  def channel_contested(channel)
+    @channel = channel
+    @email = channel.publisher.email
+
+    @transfer_url = reject_transfer_path(@channel)
```
yachtcaptain23: Would you use `token_reject_transfer_url(channel, channel.contest_token)` and keep the helper skinny?

app/services/channels/approve_channel_transfer.rb
```diff
@@ -0,0 +1,44 @@
...
+  class ApproveChannelTransfer < BaseService
+
+    def initialize(channel:, should_delete: true)
+      @channel = channel
+      @should_delete = should_delete
+    end
+
+    def perform
+      # Verify the contested_by channel
+      contested_by = @channel.contested_by_channel
```
yachtcaptain23: We should lock this in a https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html

app/models/base_channel_details.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  after_initialize :initialize_id, unless: -> { self.id }
+
+  def channel_identifier
+    raise "Override channel_identifier"
+  end
+
+  private
+
+  def initialize_id
```
yachtcaptain23: Would you explain what this is for?

app/models/site_channel_details.rb
```diff
@@ -107,9 +103,9 @@ def url
 
   private
 
-  def verified_publisher_id_exists?
-    self.class.joins(:channel).where(brave_publisher_id: brave_publisher_id, "channels.verified": true).any?
-  end
+  # def verified_publisher_id_exists?
```
yachtcaptain23: Would you delete if this is dead code?

app/services/channels/contest_channel.rb
```diff
@@ -0,0 +1,52 @@
+module Channels
+  class ContestChannel < BaseService
+    def initialize(channel:, contested_by:)
+      @channel = channel
+      @contested_by = contested_by
+      require_same_channel_id
+    end
+
+    def perform
+      @contested_by.verified = false
```
yachtcaptain23: Lock here too

app/services/channels/reject_channel_transfer.rb
```diff
@@ -0,0 +1,44 @@
...
+    def initialize(channel:, should_delete: true)
+      @channel = channel
+      @contested_by = @channel.contested_by_channel
+      @should_delete = should_delete
+
+      require_contested_channel
+    end
+
+    def perform
+      # Remove contention from original verified channel
```
yachtcaptain23: And here

app/services/site_channel_domain_setter.rb
```diff
@@ -40,8 +40,8 @@ def normalize_domain
     # Throw a Addressable::URI:InvalidURIError if it's an invalid URI
     Addressable::URI.parse("http://" + channel_details.brave_publisher_id)
 
-    if SiteChannelDetails.joins(:channel).where(brave_publisher_id: channel_details.brave_publisher_id, "channels.verified": true).any?
-      channel_details.brave_publisher_id_error_code = :taken
+    if site_already_verified?(channel_details)
+      # channel_details.brave_publisher_id_error_code = :taken
```
yachtcaptain23: dead code?

app/services/site_channel_domain_setter.rb
```diff
@@ -85,4 +85,10 @@ def inspect_host
 
   class DomainExclusionError < RuntimeError
   end
+
+  private
+
+  def site_already_verified?(channel_details)
+    SiteChannelDetails.joins(:channel).where(brave_publisher_id: channel_details.brave_publisher_id, "channels.verified": true).any?
```
yachtcaptain23: Would this be simplified to: `channel_details.channel.verified?`?

app/models/base_channel_details.rb
nvonpentz: This was left over from Larry's initial PR.  I couldn't figure out what he wanted it for, but forgot to remove it.  Will delete.

app/controllers/channel_transfer_controller.rb
nvonpentz: @yachtcaptain23 the token is set to `nil` in `Channels::RejectChannelTransfer`

app/services/site_channel_domain_setter.rb
nvonpentz: This method should return true if any verified site has the same brave publisher id, so not the same.  But this code will be removed anyway since we're allowing publishers to try and verify a site that has already been verified.

app/controllers/channel_transfer_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      respond_to do |format|
+        format.json {
+          head 404
```
CoryMcDonald: It might be easier if we added a constraint in the `routes.rb` file
`get :reject_transfer, to: 'channel_transfer#reject_transfer' constraints: lambda { |req| req.format == :json }`

Then we wouldn't have to specify the format and have the duplicated logic

app/controllers/publishers/omniauth_callbacks_controller.rb
```diff
@@ -21,17 +21,12 @@ def register_youtube_channel
...
 
-      if existing_channel
-        if existing_channel.publisher == current_publisher
-          redirect_to home_publishers_path, notice: t(".channel_already_registered")
-          return
-        else
-          redirect_to home_publishers_path, flash: { taken_channel_id: existing_channel.id }
-          return
-        end
+      if existing_channel && existing_channel.publisher == current_publisher
```
CoryMcDonald: We could use the [nil-safe navigation operator](http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/) here which would look like
`if existing_channel&.publisher == current_publisher`

app/helpers/channels_helper.rb
```diff
@@ -113,4 +114,16 @@ def should_display_verification_token?(channel)
     return false if channel.verified? || channel.details_type != "SiteChannelDetails"
     ["no_txt_records", "token_incorrect_dns", "token_not_found_dns"].include?(channel.verification_details)
   end
+
+  def time_until_transfer(channel)
+    return unless channel.verification_pending? || channel.contest_token.present?
+    contest_timesout_at = channel.contest_timesout_at || channel.contesting_channel.contest_timesout_at
+    contest_already_timed_out = (contest_timesout_at - Time.now) < 0
```
CoryMcDonald: Are we concerned about the user's timezone here? Seems like this could be a potential for locale issues.

nvonpentz: @CoryMcDonald I don't think the users time zone matters in this situation.  The app sets a `contest_timesout_at` which we are counting down to.  The time until the timeout shouldn't change regardless of what time the user is in unless I'm misunderstanding something.

app/services/site_channel_verifier.rb
```diff
@@ -15,7 +15,7 @@ def initialize(has_admin_approval: false, channel:)
 
   def perform
     return true if channel.verified?
-    return false if channel.verification_awaiting_admin_approval? && !has_admin_approval
+    return false if channel.verification_awaiting_admin_approval? && !has_admin_approval || channel.verification_pending?
```
yachtcaptain23: Would you put closing parentheses at the appropriate places?

app/services/publisher_channel_deleter.rb
```diff
@@ -1,37 +1,30 @@
...
-    Rails.logger.info("PublisherChannelDeleter eyeshade offline; not deleting channel")
-    true
-  end
-
-  private
-
-  def api_base_uri
-    Rails.application.secrets[:api_eyeshade_base_uri]
-  end
+    success = @channel.destroy
```
yachtcaptain23: We should delete the channel after it's confirmed with other services that it's been properly disposed of. I'm worried that if the heroku dyno resets here, Publishers will have a missing channel and elsewhere will have the channel.

yachtcaptain23: It seems we could move delete to `DeletePublisherChannelJob` as the last step.

nvonpentz: @yachtcaptain23 Hm I'm not sure that would make a difference.  If we made the change, wouldn't it be equally possible that the heroku dyno restarts after it receives the responses back from promo and eyeshade? In that case, it'd be deleted on promo and eyeshade but not publishers, right?

If that's right, I think I prefer this method because it allows us to make the deletion requests to eyeshade and promo async.

yachtcaptain23: By moving to a job, it retries during failure (which I believe includes power failure). If you run on the main web thread, it doesn't retry the job.

PR #1132, opened 2018-09-05
Feature/hide uphold widget

yachtcaptain23:
Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/views/admin/publishers/edit.html.slim
+++ b/app/views/admin/publishers/edit.html.slim
@@ -12,4 +12,4 @@
         = f.label "Uphold widget disabled"
         = f.check_box :excluded_from_payout, {}, checked: @publisher.excluded_from_payout?
       .form-group
-        = f.submit
+        = f.submit class: "btn btn-primary"
--- a/app/views/admin/publishers/edit.html.slim
+++ b/app/views/admin/publishers/edit.html.slim
@@ -1,0 +1,15 @@
+#publisher
+  #publisher-section
+    - if flash[:notice]
+      .alert.alert-success role="alert"
+        = flash[:notice]
+    - if flash[:alert]
+      .alert.alert-danger role="alert"
+        = flash[:alert]
+    h2 = @publisher.name
+    = form_for(@publisher, url: admin_publisher_path, method: :patch) do |f|
+      .form-group
+        = f.label "Uphold widget disabled"
+        = f.check_box :uphold_disabled, {}, checked: @publisher.uphold_disabled?
+      .form-group
+        = f.submit
--- a/app/views/admin/publishers/edit.html.slim
+++ b/app/views/admin/publishers/edit.html.slim
@@ -10,6 +10,6 @@
     = form_for(@publisher, url: admin_publisher_path, method: :patch) do |f|
       .form-group
         = f.label "Uphold widget disabled"
-        = f.check_box :uphold_disabled, {}, checked: @publisher.uphold_disabled?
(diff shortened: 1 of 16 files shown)


db/schema.rb
```diff
@@ -246,6 +247,7 @@
     t.text "role", default: "publisher"
     t.datetime "javascript_last_detected_at"
     t.datetime "default_currency_confirmed_at"
+    t.boolean "uphold_disabled", default: false, null: false
```
nvonpentz: Would `exclude_from_payout` or something similar be better here?  We want to prevent payouts regardless of the wallet provider

config/secrets.yml
```diff
@@ -134,7 +134,7 @@ test:
   uphold_client_id: test_client_id
   uphold_client_secret: test_client_secret
   uphold_scope: cards:read,cards:write,user:read
-  uphold_api_uri: "https://uphold-api.example.com"
+  uphold_api_uri: "https://api-sandbox.uphold.com"
```
nvonpentz: I don't think we should be making any external requests for tests.

If you replace the failing travis test with this it should work:
```ruby
  test "when online returns true when the wallet can be set" do
    prev_offline = Rails.application.secrets[:api_eyeshade_offline]
    begin
      Rails.application.secrets[:api_eyeshade_offline] = false

      publisher = publishers(:verified)
      publisher.uphold_access_parameters = "{\"foo\":\"bar\"}"
      wallet = "{\"wallet\":\"abc123\"}"

      stub_request(:put, /v1\/owners\/#{URI.escape(publisher.owner_identifier)}\/wallet/).
          with(body: "{\n  \"provider\": \"uphold\",\n  \"parameters\": {\"foo\":\"bar\",\"server\":\"https://uphold-api.example.com\"}\n}\n").
          to_return(status: 200, body: wallet, headers: {})

      result = PublisherWalletSetter.new(publisher: publisher).perform
      assert_equal 200, result.status
    ensure
      Rails.application.secrets[:api_eyeshade_offline] = prev_offline
    end
  end
```

The problem was the the stub request did not match the request made, and thus returned `<Faraday::ConnectionFailed>` and `nil` as the result

nvonpentz:
image

Edit contact button is floating off to the right and doesn't work

app/views/admin/publishers/edit.html.slim
```diff
@@ -0,0 +1,15 @@
...
+    - if flash[:alert]
+      .alert.alert-danger role="alert"
+        = flash[:alert]
+    h2 = @publisher.name
+    = form_for(@publisher, url: admin_publisher_path, method: :patch) do |f|
+      .form-group
+        = f.label "Uphold widget disabled"
+        = f.check_box :excluded_from_payout, {}, checked: @publisher.excluded_from_payout?
+      .form-group
+        = f.submit
```
nvonpentz: Can we make this a button for consistency?
image

yachtcaptain23: `f.submit` is the standard convention for working with forms.

yachtcaptain23:
My long term vision is for other options and interactions to occur on the same row as `Edit Settings`.

app/views/admin/publishers/edit.html.slim
nvonpentz: sorry, I was referring to a bootstrap button

PR #1149, opened 2018-09-06
Update criteria for EnqueueSiteChannelVerifications

### LINKED ISSUE #1148: "Started" is not a valid verification status
and thus should not be used to determine whether a site channel is enqueued for verification:

https://github.com/brave-intl/publishers/blob/5a284356c004675ca6249989cf711de2ac0d1e70/app/models/site_channel_details.rb#L40

nvonpentz:
Resolves #1148 

Only enqueue site channels for verification that are:
* Not verified
* Have a verification method
* Were updated as recently

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/models/site_channel_details.rb
+++ b/app/models/site_channel_details.rb
@@ -37,8 +37,16 @@ class SiteChannelDetails < ApplicationRecord
   }
 
   scope :recent_ready_to_verify_site_channels, -> (max_age: 6.weeks) {
-    recent_unverified_site_channels(max_age: max_age).where("channels.verification_status": "started").
-        or(recent_unverified_site_channels(max_age: max_age).where("channels.verification_status": "failed"))
+    joins(:channel)
+        .where.not(
+          brave_publisher_id: SiteChannelDetails
+                              .joins(:channel)
+                              .select(:brave_publisher_id)
+                              .distinct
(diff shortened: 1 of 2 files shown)


app/models/site_channel_details.rb
```diff
@@ -37,8 +37,10 @@ def validate_each(record, attribute, value)
   }
 
   scope :recent_ready_to_verify_site_channels, -> (max_age: 6.weeks) {
-    recent_unverified_site_channels(max_age: max_age).where("channels.verification_status": "started").
-        or(recent_unverified_site_channels(max_age: max_age).where("channels.verification_status": "failed"))
+    joins(:channel)
+        .where.not(brave_publisher_id: SiteChannelDetails.joins(:channel).select(:brave_publisher_id).distinct.where("channels.verified": true))
```
yachtcaptain23: Nit: Would you fix the indentation alignment?

yachtcaptain23: E.g.
```
.where.not(
  brave_publisher_id: SiteChannelDetails
                                    .joins(:channel)
                                    .select(:brave_publisher_id)
                                    .distinct
                                    .where("channels.verified": true)
  )
```

PR #1172, opened 2018-09-13
V1 API for stats

### LINKED ISSUE #896: expose statistics about channels
currently stats scrapes an eyeshade report to build it's own local copy of information about channels. as publishers is the source of truth for this info we should decide on a way to expose info here instead.

some of the fields that stats is currently collecting are: `publisher, name, url, created_at, verified, authorized, alexa_rank, platform, audience, provider`

also, it has been noted that having dates for signup and verification would be ideal

dlipeles:
#### API for stats, v1

* `api/v1/stats/channels`
`returns list of channel uuids`

* `api/v1/stats/channels/:uuid`
`returns the following for a given channel :`
```
{
  "uuid": 
  "channel_id": 
  "channel_type": 
  "name": 
  "stats": 
  "url": 
  "owner_id": 
  "created_at": 
  "verified": 
}
```
* `api/v1/stats/channels/:uuid` returns `404` if channel not found
* `api/v1/stats/` + `signups_per_day`, `email_verified_signups_per_day`, `etc...`
`returns results of the query`

##### Tests 

Found at `/test/controllers/api/v1/stats/channels_controller_test.rb`

Closes #896

### DIFF
--- a/app/controllers/api/v1/stats/channels_controller.rb
+++ b/app/controllers/api/v1/stats/channels_controller.rb
@@ -3,45 +3,53 @@ class Api::V1::Stats::ChannelsController < Api::BaseController
   def channels
 
     if params[:uuid] == nil
+
       channels = Channel.all.map { |channel| channel.id }
       data = JSON.pretty_generate(channels)
       render(json: data)
+
     else
+
       channel = Channel.find_by_id(params[:uuid])
 
-      case channel.details_type
-        when "SiteChannelDetails"
-          channel_details = SiteChannelDetails.find_by_id(channel.details_id)
-          channel_type = "website"
-          channel_name = channel_details.url
-        when "YoutubeChannelDetails"
-          channel_details = YoutubeChannelDetails.find_by_id(channel.details_id)
-          channel_type = "youtube"
-          channel_name = channel_details.name
-        when "TwitterChannelDetails"
-          channel_details = TwitterChannelDetails.find_by_id(channel.details_id)
-          channel_type = "twitter"
-          channel_name = channel_details.name
-        when "TwitchChannelDetails"
-          channel_details = TwitchChannelDetails.find_by_id(channel.details_id)
-          channel_type = "twitch"
-          channel_name = channel_details.name
-        end
+      if(channel == nil)
+        redirect_to "/404"
+      else
 
-        data = JSON.pretty_generate({
-          uuid: channel.id,
-          channel_id: channel_details.channel_identifier,
-          channel_type: channel_type,
-          name: channel_name,
-          stats: channel_details.stats,
-          url: channel_details.url,
-          owner_id: channel.publisher.owner_identifier,
-          created_at: channel.created_at,
-          verified: channel.verified
-          })
+        case channel.details_type
+          when "SiteChannelDetails"
+            channel_details = SiteChannelDetails.find_by_id(channel.details_id)
+            channel_type = "website"
+            channel_name = channel_details.url
+          when "YoutubeChannelDetails"
+            channel_details = YoutubeChannelDetails.find_by_id(channel.details_id)
+            channel_type = "youtube"
+            channel_name = channel_details.name
+          when "TwitterChannelDetails"
+            channel_details = TwitterChannelDetails.find_by_id(channel.details_id)
+            channel_type = "twitter"
+            channel_name = channel_details.name
+          when "TwitchChannelDetails"
+            channel_details = TwitchChannelDetails.find_by_id(channel.details_id)
+            channel_type = "twitch"
+            channel_name = channel_details.name
+          end
 
-        render(json: data)
+          data = JSON.pretty_generate({
+            uuid: channel.id,
+            channel_id: channel_details.channel_identifier,
+            channel_type: channel_type,
+            name: channel_name,
+            stats: channel_details.stats,
+            url: channel_details.url,
+            owner_id: channel.publisher.owner_identifier,
+            created_at: channel.created_at,
+            verified: channel.verified
+            })
 
+          render(json: data)
+
+      end
     end
   end
 end
--- a/app/controllers/api/v1/stats/channels_controller.rb
+++ b/app/controllers/api/v1/stats/channels_controller.rb
@@ -1,0 +1,48 @@
+class Api::V1::Stats::ChannelsController < Api::BaseController
+
+def channels
+  channels = Channel.all.map { |channel| channel.id }
+  data = JSON.pretty_generate(channels)
+
+  render(json: data)
+
+end
+
+def channels_uuid
+
+  channel = Channel.find_by_id(params[:uuid])
+
+  case channel.details_type
+  when "SiteChannelDetails"
+    channel_details = SiteChannelDetails.find_by_id(channel.details_id)
+    platform = "Site"
+    channel_id = channel_details.brave_publisher_id
+  when "YoutubeChannelDetails"
+    channel_details = YoutubeChannelDetails.find_by_id(channel.details_id)
+    platform = "Youtube"
+    channel_id = channel_details.youtube_channel_id
+  when "TwitterChannelDetails"
+    channel_details = TwitterChannelDetails.find_by_id(channel.details_id)
+    platform = "Twitter"
+    channel_id = channel_details.twitter_channel_id
+  when "TwitchChannelDetails"
+    channel_details = TwitchChannelDetails.find_by_id(channel.details_id)
+    platform = "Twitch"
+    channel_id = channel_details.twitch_channel_id
+  end
+
+  data = JSON.pretty_generate({
+    uuid: channel.id,
+    platform: platform,
+    name: channel_details.name,
+    channel_id: channel_id,
+    stats: channel_details.stats,
+    publisher_id: channel.publisher_id,
+    created_at: channel.created_at,
+    verified: channel.verified
+  })
+
+  render(json: data)
+
+end
+end
--- a/app/controllers/api/v1/stats/channels_controller.rb
+++ b/app/controllers/api/v1/stats/channels_controller.rb
@@ -2,7 +2,7 @@ class Api::V1::Stats::ChannelsController < Api::BaseController
 
   def channels
 
-    if(params[:uuid] == nil)
+    if params[:uuid].nil?
 
       channels = Channel.all.map { |channel| channel.id }
       data = JSON.pretty_generate(channels)
@@ -10,20 +10,20 @@ class Api::V1::Stats::ChannelsController < Api::BaseController
 
     else
 
-      channel = Channel.find_by_id(params[:uuid])
+      begin
 
-      if(channel == nil)
+        channel = Channel.find(params[:uuid])
 
-        error_response = JSON.pretty_generate({
-          errors: [{
-            status: "404",
-            title: "Not Found",
-            detail: "Channel with uuid " + params[:uuid] + " not found"
-          }]
-        })
-        render(json: error_response)
-
-      else
+        rescue ActiveRecord::RecordNotFound
+          error_response = JSON.pretty_generate({
+            errors: [{
+              status: "404",
+              title: "Not Found",
+              detail: "Channel with uuid #{params[:uuid]} not found"
+              }]
+            })
+          render(json: error_response) and return
+        end
 
         case channel.details_type
 
@@ -59,7 +59,6 @@ class Api::V1::Stats::ChannelsController < Api::BaseController
 
           render(json: data)
 
-      end
     end
   end
 end
--- a/app/controllers/api/v1/stats/channels_controller.rb
+++ b/app/controllers/api/v1/stats/channels_controller.rb
@@ -13,10 +13,20 @@ class Api::V1::Stats::ChannelsController < Api::BaseController
       channel = Channel.find_by_id(params[:uuid])
 
       if(channel == nil)
-        redirect_to "/404"
+
+        error_response = JSON.pretty_generate({
+          errors: [{
+            status: "404",
+            title: "Not Found",
+            detail: "Channel with uuid " + params[:uuid] + " not found"
+          }]
+        })
+        render(json: error_response)
+
       else
 
         case channel.details_type
+
           when "SiteChannelDetails"
             channel_details = SiteChannelDetails.find_by_id(channel.details_id)
             channel_type = "website"
--- a/app/controllers/api/v1/stats/channels_controller.rb
+++ b/app/controllers/api/v1/stats/channels_controller.rb
@@ -2,7 +2,7 @@ class Api::V1::Stats::ChannelsController < Api::BaseController
 
   def channels
 
-    if params[:uuid] == nil
+    if(params[:uuid] == nil)
 
       channels = Channel.all.map { |channel| channel.id }
       data = JSON.pretty_generate(channels)
--- a/app/controllers/api/v1/stats/channels_controller.rb
+++ b/app/controllers/api/v1/stats/channels_controller.rb
@@ -1,48 +1,47 @@
 class Api::V1::Stats::ChannelsController < Api::BaseController
 
-def channels
-  channels = Channel.all.map { |channel| channel.id }
-  data = JSON.pretty_generate(channels)
+  def channels
 
-  render(json: data)
+    if params[:uuid] == nil
+      channels = Channel.all.map { |channel| channel.id }
+      data = JSON.pretty_generate(channels)
+      render(json: data)
+    else
+      channel = Channel.find_by_id(params[:uuid])
 
-end
+      case channel.details_type
+        when "SiteChannelDetails"
+          channel_details = SiteChannelDetails.find_by_id(channel.details_id)
+          channel_type = "website"
+          channel_name = channel_details.url
+        when "YoutubeChannelDetails"
+          channel_details = YoutubeChannelDetails.find_by_id(channel.details_id)
+          channel_type = "youtube"
+          channel_name = channel_details.name
+        when "TwitterChannelDetails"
+          channel_details = TwitterChannelDetails.find_by_id(channel.details_id)
+          channel_type = "twitter"
+          channel_name = channel_details.name
+        when "TwitchChannelDetails"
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/stats/stats_controller.rb
```diff
@@ -1,4 +1,5 @@
-class Api::StatsController < Api::BaseController
+class Api::V1::Stats::StatsController < Api::BaseController
```
yachtcaptain23: IMO the controller could be better named.

Perhaps EngagementController? @nvonpentz do you have any thoughts?

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,48 @@
+class Api::V1::Stats::ChannelsController < Api::BaseController
+
+def channels
```
yachtcaptain23: Would you fix indentation? 

Also do we need this function?

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    platform = "Youtube"
+    channel_id = channel_details.youtube_channel_id
+  when "TwitterChannelDetails"
+    channel_details = TwitterChannelDetails.find_by_id(channel.details_id)
+    platform = "Twitter"
+    channel_id = channel_details.twitter_channel_id
+  when "TwitchChannelDetails"
+    channel_details = TwitchChannelDetails.find_by_id(channel.details_id)
+    platform = "Twitch"
+    channel_id = channel_details.twitch_channel_id
```
nvonpentz: I think we want `channel_details.channel_identifier` here and above:

```
irb(main):117:0> channel_details.channel_identifier
=> "youtube#channel:UCtsfHRe2WQnkNH5WYJWL-Yw"
irb(main):118:0> channel_details.youtube_channel_id
=> "UCtsfHRe2WQnkNH5WYJWL-Yw"
```

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  data = JSON.pretty_generate({
+    uuid: channel.id,
+    platform: platform,
+    name: channel_details.name,
+    channel_id: channel_id,
+    stats: channel_details.stats,
+    publisher_id: channel.publisher_id,
```
nvonpentz: I think we want `channel.publisher.owner_identifier` here:
```
irb(main):121:0> channel.publisher_id
=> "08df5822-557c-4944-a38e-ef8c7b557f96"
irb(main):122:0> channel.publisher.owner_identifier
=> "publishers#uuid:08df5822-557c-4944-a38e-ef8c7b557f96"
```

nvonpentz: also prefer `owner_id` for consistency and other past naming convention problems https://github.com/brave-intl/publishers/blob/8b6093cb8fc8487534e1e898d0a224ef00a0166c/app/services/promo_registrar.rb#L116

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+end
+
+def channels_uuid
+
+  channel = Channel.find_by_id(params[:uuid])
+
+  case channel.details_type
+  when "SiteChannelDetails"
+    channel_details = SiteChannelDetails.find_by_id(channel.details_id)
+    platform = "Site"
```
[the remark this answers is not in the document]
nvonpentz: promo uses "website"
https://github.com/brave-intl/publishers/blob/8b6093cb8fc8487534e1e898d0a224ef00a0166c/app/services/promo_registrar.rb#L120
 and also "channel_type" instead of platform

app/controllers/api/v1/stats/stats_controller.rb
nvonpentz: I think stats fits since this info will be sent to the stats service and includes information that may not be related to engagement 

maybe DataController?

app/controllers/api/v1/stats/channels_controller.rb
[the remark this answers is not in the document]
dlipeles: Will change to lowercase and `channel_type`
so:

`channel_type` : `website`
`channel_type` : `youtube`
`channel_type` : `twitter`
`channel_type` : `twitch`

please confirm

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Fixed, consolidated this to one function

app/controllers/api/v1/stats/stats_controller.rb
dlipeles: Any consensus on this? It might make more sense to leave it 'as is' right now, wait for more stats to be built, then see if any patterns emerge.

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Fixed

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Fixed, I also added the `url` property you mentioned in our call

question: I don't see `channel_identifier`, or `url` in `schema.rb` how did you know about those properties?

app/controllers/api/v1/stats/channels_controller.rb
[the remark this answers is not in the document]
dlipeles: Great, fixed

app/controllers/api/v1/stats/channels_controller.rb
nvonpentz: @dlipeles they are methods on the models, not fields in the database

https://github.com/brave-intl/publishers/blob/8b6093cb8fc8487534e1e898d0a224ef00a0166c/app/models/youtube_channel_details.rb#L12-L14

dlipeles: Ah very cool, thanks Nick

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,47 @@
+class Api::V1::Stats::ChannelsController < Api::BaseController
+
+  def channels
+
+    if params[:uuid] == nil
```
yachtcaptain23: IMO we should enforce a `uuid` as a RESTful convention

dlipeles: Sure, I'd like to hear more of what you had in mind

Do we want to expose a list of channel_id's at `/some_route` ?

dlipeles: Looked into a few API's, I think this is okay as is, it seems common practice to return all of `resource` at `/resource`, and a specific `resource` at `/resource/:id`

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,65 @@
+class Api::V1::Stats::ChannelsController < Api::BaseController
+
+  def channels
+
+    if(params[:uuid] == nil)
```
CoryMcDonald: I'd prefer to use `.blank?` or `.nil?` it's more ruby-esque. You can view the [differences here](https://stackoverflow.com/questions/885414/a-concise-explanation-of-nil-v-empty-v-blank-in-ruby-on-rails)

config/routes.rb
```diff
@@ -86,25 +86,32 @@
...
-      get :twitch_channels_by_view_count
-      get :javascript_enabled_usage
-    end
+
+    # /api/v1/
     namespace :v1, defaults: { format: :json } do
+      # /api/v1/stats/
+      namespace :stats, defaults: { format: :json } do
+        get "/channels", to: "channels#channels"
+        get "/channels/:uuid", to: "channels#channels"
```
CoryMcDonald: I don't really like this design. We should follow a more isolated approach, where this actually routes to a different method in the controller.

This could be more rails-esque. If you name your function `index` and another one `show` in the channels controller, you can simply have this be 
`resource :channels, only: [:index, :show]`.

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+          when "TwitchChannelDetails"
+            channel_details = TwitchChannelDetails.find_by_id(channel.details_id)
+            channel_type = "twitch"
+            channel_name = channel_details.name
+          end
+
+          data = JSON.pretty_generate({
+            uuid: channel.id,
+            channel_id: channel_details.channel_identifier,
+            channel_type: channel_type,
```
CoryMcDonald: Can we use this helper that has already been created instead of defining our own `channel_type` variable in each one of these?
https://github.com/brave-intl/publishers/blob/666ea3ea39a7aaeb9face944413d3685bf62a605/app/helpers/publishers_helper.rb#L367

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      channel = Channel.find_by_id(params[:uuid])
+
+      if(channel == nil)
+
+        error_response = JSON.pretty_generate({
+          errors: [{
+            status: "404",
+            title: "Not Found",
+            detail: "Channel with uuid " + params[:uuid] + " not found"
```
CoryMcDonald: String interpolation would probably make more sense here
`"Channel with uuid #{params[:uuid]} not found"`

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      channels = Channel.all.map { |channel| channel.id }
+      data = JSON.pretty_generate(channels)
+      render(json: data)
+
+    else
+
+      channel = Channel.find_by_id(params[:uuid])
+
+      if(channel == nil)
```
CoryMcDonald: This is personal preference so it's up to you, but i'd rather have the `.find` raise an exception and then we catch it in this method instead of checking for nil and having code be more nested.

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,65 @@
+class Api::V1::Stats::ChannelsController < Api::BaseController
+
+  def channels
+
+    if(params[:uuid] == nil)
+
+      channels = Channel.all.map { |channel| channel.id }
+      data = JSON.pretty_generate(channels)
+      render(json: data)
```
CoryMcDonald: nit: parenthesis not needed `render json: data`

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Agree it looks nicer to use `.nil?`, Fixed.

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Normally I would agree, but since this is an api for stats I wouldn't want `I18n.t` to rename values in unpredictable ways. Certainly will want to use this if we ever render this data to the browser

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Agree, it's cleaner, fixed.

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Agree, was debating this before, since it does add some more code, but in the future we'll probably want to check for more errors so it's good to lay a foundation for handling, Fixed.

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Seems that we like to use the parenthesis here so I'll keep it as a Publishers convention

config/routes.rb
dlipeles: There's a high chance we'll rename the routes & controllers once more statistics are added, 
I will leave this as-is for now, but I agree with your sentiment for a cleaner route

CoryMcDonald: Could we still split out the specific UUID lookup into a separate method?

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,64 @@
+class Api::V1::Stats::ChannelsController < Api::BaseController
+
+  def channels
+
+    if params[:uuid].nil?
+
+      channels = Channel.all.map { |channel| channel.id }
+      data = JSON.pretty_generate(channels)
```
yachtcaptain23: `pretty_generate` is not what you're looking for when you sample the response. I'd recommend `channels.to_json`.

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,64 @@
+class Api::V1::Stats::ChannelsController < Api::BaseController
+
+  def channels
+
+    if params[:uuid].nil?
```
yachtcaptain23: Let's make this RESTful by doing:

```
def index
   render(json: channels.pluck(:id).to_json, status: :ok) and return
end
```

and 

```
def show
  # your other code
end
```

config/routes.rb
yachtcaptain23: I agree with @CoryMcDonald. `ChannelsController` should follow a RESTful convention

test/controllers/api/v1/stats/channels_controller_test.rb
```diff
@@ -0,0 +1,35 @@
...
+        assert(result.key?("name"))
+        assert(result.key?("stats"))
+        assert(result.key?("url"))
+        assert(result.key?("owner_id"))
+        assert(result.key?("created_at"))
+        assert(result.key?("verified"))
+
+    else get "/api/v1/stats/channels/invalid_uuid", headers: { "HTTP_AUTHORIZATION" => "Token token=fake_api_auth_token" }
+      result = JSON.parse(response.body)
+      assert_equal(404, Integer(result["errors"][0]["status"]))
```
yachtcaptain23: I believe `assert_equal 404, response.status` is more conventional

test/controllers/api/v1/stats/channels_controller_test.rb
```diff
@@ -0,0 +1,35 @@
+require "test_helper"
+require "shared/mailer_test_helper"
```
yachtcaptain23: deadcode

test/controllers/api/v1/stats/channels_controller_test.rb
```diff
@@ -0,0 +1,35 @@
...
+    result = JSON.parse(response.body)
+
+    assert result.is_a?(Array)
+    assert_equal(Channel.count, result.length)
+
+  end
+
+  test "/api/v1/stats/channels/:uuid returns json representation of channel, or 404 if not found" do
+
+    if(Channel.last != nil)
```
yachtcaptain23: You should be creating a `Channel` to test against

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Fixed

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+        rescue ActiveRecord::RecordNotFound
+          error_response = JSON.pretty_generate({
+            errors: [{
+              status: "404",
+              title: "Not Found",
+              detail: "Channel with uuid #{params[:uuid]} not found"
+              }]
+            })
+          render(json: error_response) and return
```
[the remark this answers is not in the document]
dlipeles: Agree, the status code should've been set as 404, it's not a bad idea to set it in the error json as well

see http://jsonapi.org/format/#errors

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Fixed

config/routes.rb
dlipeles: Fixed

test/controllers/api/v1/stats/channels_controller_test.rb
dlipeles: Fixed, but `require "test_helper"` is needed .

test/controllers/api/v1/stats/channels_controller_test.rb
dlipeles: Fixed

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+
+      channels = Channel.all.map { |channel| channel.id }
+      data = channels.to_json
+      render(status: 200, json: data)
+
+  end
+
+  def show
+
+      begin
```
CoryMcDonald: We don't have to have this begin here, we can just rescue down below like this

```ruby
def some_method
  danger_danger
  true # good response
rescue Error
  false # error response
end
```

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,63 @@
+class Api::V1::Stats::ChannelsController < Api::BaseController
+
+  def index
+
+      channels = Channel.all.map { |channel| channel.id }
```
yachtcaptain23: If I understand this correctly, this selects all the columns and then does a parse. `Channel.pluck(:id)` should tell the database to only get the `id` column.

dlipeles: Added ✔️

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Added ✔️

test/controllers/api/v1/stats/channels_controller_test.rb
dlipeles: Added ✔️

dlipeles:
Closing this branch, see continuation and improvement in #1296

PR #1180, opened 2018-09-18
Admin dashboard data

### LINKED ISSUE #1173: Conversion and piwik metrics for brave.com/publishers and publishers.basicattentiontoken.org
I want to know how many visitors we get to brave.com/publishers and visitors to publishers.basicattentiontoken.org

Additionally, if we can get Piwik numbers on the `Become a Verified Publisher` button.

dlipeles:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/admin/dashboard_controller.rb
+++ b/app/controllers/admin/dashboard_controller.rb
@@ -2,6 +2,7 @@ class Admin::DashboardController < AdminController
   def index
     @site = Piwik::Site.load(6)
     @seo_info = @site.seo_info
+    @visits = getVisits()
   end
 
   def fetch(type)
@@ -17,4 +18,10 @@ class Admin::DashboardController < AdminController
       :devices_detection]
     )
   end
+
+  def getVisits()
+    query = Piwik::VisitsSummary.getVisits(:idSite => 6, :period => :month, :date => :last12)
+    return query.result.to_json
+  end
+
 end
--- a/app/controllers/admin/dashboard_controller.rb
+++ b/app/controllers/admin/dashboard_controller.rb
@@ -2,7 +2,6 @@ class Admin::DashboardController < AdminController
   def index
     @site = Piwik::Site.load(6)
     @seo_info = @site.seo_info
-    @visits = getVisits()
   end
 
   def fetch(type)
@@ -15,13 +14,9 @@ class Admin::DashboardController < AdminController
   def valid_type?(type)
     type.to_sym.in?([
       :events,
+      :visits,
       :devices_detection]
     )
-  end
-
-  def getVisits()
-    query = Piwik::VisitsSummary.getVisits(:idSite => 6, :period => :month, :date => :last12)
-    return query.result.to_json
   end
 
 end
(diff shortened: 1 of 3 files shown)


app/controllers/admin/dashboard_controller.rb
```diff
@@ -17,4 +18,10 @@ def valid_type?(type)
       :devices_detection]
     )
   end
+
+  def getVisits()
```
CoryMcDonald: Does the linter not pick this up? Can we add this to the rubocop config?

app/controllers/admin/dashboard_controller.rb
```diff
@@ -17,4 +18,10 @@ def valid_type?(type)
...
+    query = Piwik::VisitsSummary.getVisits(:idSite => 6, :period => :month, :date => :last12)
+    return query.result.to_json
```
CoryMcDonald: Ruby automatically returns what's at the end of methods so we don't necessarily have to have a `return` statement here

app/controllers/admin/dashboard_controller.rb
```diff
@@ -2,6 +2,7 @@ class Admin::DashboardController < AdminController
   def index
     @site = Piwik::Site.load(6)
     @seo_info = @site.seo_info
+    @visits = getVisits()
```
CoryMcDonald: No need to have parenthesis for a method that doesn't have params

app/javascript/admin/dashboard/index.js
```diff
@@ -1,5 +1,68 @@
 import Chart from 'chart.js';
 
+function createVisitsChart(id, title, dataSource) {
```
yachtcaptain23: When you create a chart, would you continue to use `createChart` to keep things DRY? Please edit the `backgroundColor` and `borderColor` given you have more colors.

app/controllers/admin/dashboard_controller.rb
yachtcaptain23: Would you move the logic to the view?

In the future, I want things to be done asynchronously. (e.g. the page makes calls to the server and asks for content)

yachtcaptain23: Ah yeah they use `visits` instead of `visit_summary` as spec'd here: `https://github.com/matomo-org/piwik-ruby-api/blob/5fd3a74db5feafdc37089b8d2c3af475448f509c/lib/piwik/site.rb`

app/controllers/admin/dashboard_controller.rb
dlipeles: Fixed, deleted this function

app/controllers/admin/dashboard_controller.rb
dlipeles: Fixed, deleted this function

app/controllers/admin/dashboard_controller.rb
dlipeles: Thanks, that worked

app/javascript/admin/dashboard/index.js
dlipeles: Yes, Fixed

PR #1187, opened 2018-09-24
Use new endpoints for publisher statements and improve statement format

### LINKED ISSUE #1181: Use eyeshade transaction table APIs for publisher statement generation
Currently the PublisherStatementGenerator uses the legacy eyeshade endpoint that returns a `source_url` where the statement can be downloaded from.  However, `source_url` will 404 until the statement has been properly generated by eyeshade, which can take around a minute.  As a result, Publishers must query this endpoint until there is a successful response, at which point Publishers can process the statement and present it to the user.

We should replace the existing statement generation code and use the new eyeshade API `GET /accounts/{owner_identifier}/transactions` which queries the transaction table (the proper source of truth) for settlement history.  This will return all the settlements for a given owner in JSON; there is no need to wait for a `source_url` to exist.

nvonpentz:
Resolves #1181 

This PR removes and replaces the existing statement infrastructure.  Instead of requesting a source URL from eyeshade, and repeatedly querying the source url until the statement is ready, Publishers now makes a single API call to eyeshade requesting all the transaction for a given owner and parses and returns the response as a PDF.

This is an improvement in a number of ways:
* Since statement information is returned quickly, this process can be done synchronously, and Pubishers does not need to use AJAX to get the statement.
* Publishers no longer needs store the publisher statements in the database so the publisher_statements table and PublisherStatement model have been deleted.
* We are able to remove a lot of code PublisherStatementSyncer, PublisherStatementGenerator, SyncPublisherStatementJob, statements.js, etc.
* Publishers (not eyeshade) has fine grain control over the statement that is presented to publishers.  We have improved the statement format:

image

This is not the final format of the statement, but an intermediate step until we can revisit this later.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/controllers/publishers_controller.rb
```diff
@@ -520,6 +499,10 @@ def publisher_create_auth_token_params
     params.require(:publisher).permit(:email, :brave_publisher_id)
   end
 
+  # def publisher_statement_params
```
yachtcaptain23: Commented code

app/helpers/channels_helper.rb
```diff
@@ -126,4 +126,43 @@ def time_until_transfer(channel)
       distance_of_time_in_words(Time.now, channel.contesting_channel.contest_timesout_at)
     end
   end
+
+  def get_channel_from_channel_id(channel_id)
```
yachtcaptain23: Would you add a comment above function? E.g. inputs and expected output

app/views/admin/publishers/show.html.slim
```diff
@@ -36,28 +37,34 @@
...
-                    span= 'Ready'
-                    = link_to(t("shared.download"), statement_publishers_url(id: s.id), class: 'download')
-                  - else
-                    = t ".statements.delayed"
+          = link_to("this month", statement_admin_publishers_path(publisher_id: @publisher.id, statement_period: "this_month"))
+        .statement
+          = link_to("last month", statement_admin_publishers_path(publisher_id: @publisher.id, statement_period: "last_month"))
+        .statement
+          = link_to("all", statement_admin_publishers_path(publisher_id: @publisher.id, statement_period: "all"))
+        / .statement
```
yachtcaptain23: Commented code

app/assets/stylesheets/statement.scss
```diff
@@ -0,0 +1,54 @@
+.statement {
```
yachtcaptain23: I prefer if you re-use the path to `app/assets/stylesheets/pages/statements.scss`. That way you don't need to add as part of the pre-compilation process.

app/controllers/admin/publishers_controller.rb
```diff
@@ -30,30 +30,16 @@ def update
...
-  def statement_ready
-    statement = PublisherStatement.find(params[:id])
-    if statement && statement.contents
-      head 204
-    else
-      head 404
-    end
+    @transactions = PublisherStatementGetter.new(publisher: @publisher, statement_period: statement_period).perform
+    @statement_period = publisher_statement_period(@transactions)
+    statement_file_name = "BravePaymentsStatemet-#{@statement_period}"
```
yachtcaptain23: Misspelling.

Also, is there a better way of handling users who have 0 transactions? I prefer displaying an error via `flash[:alert]` or `flash[:error]` if it's empty rather than accepting an empty string.

app/controllers/publishers_controller.rb
```diff
@@ -396,37 +394,18 @@ def log_out
...
-    else
-      head 404
-    end
-  end
-
   def statement
-    statement = PublisherStatement.find(params[:id])
+    statement_period = params[:statement_period]
+    @transactions = PublisherStatementGetter.new(publisher: current_publisher, statement_period: statement_period).perform
+    @statement_period = publisher_statement_period(@transactions)
```
yachtcaptain23: I prefer a `flash[:notice]` rather than opening a blank pdf based on the response from `PublisherStatementGetter.new`

app/helpers/channels_helper.rb
nvonpentz: Yeah I agree the helper is not the right place for this.

app/helpers/publishers_helper.rb
```diff
@@ -260,66 +260,14 @@ def publisher_verification_dns_record(publisher)
...
-    "#{publisher_id}-#{date}-#{period}.csv"
-  end
-
-  def link_to_publisher_statement(publisher_statement)
-    link_to(publisher_statement_filename(publisher_statement), statement_publishers_url(id: publisher_statement.id))
-  end
-
+  def publisher_statement_period(transactions)
+    return "" if transactions.empty?
+    statement_begin_date = "#{transactions.first["created_at"].to_time.strftime("%d/%m/%y")}"
```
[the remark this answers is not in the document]
nvonpentz: No

app/helpers/channels_helper.rb
nvonpentz: Though I don't think the channel model is the right place for this since it takes a string not a channel.

```
get_channel_from_channel_id(Channel.first.details.channel_identifier)
=> #<Channel id: "55ecd577-0425-420f-8796-78598b06c8a0", publisher_id: "6cdcc152-90f1-4892-8e1a-cf356847594a",..., contest_timesout_at: nil>
```

app/controllers/admin/publishers_controller.rb
nvonpentz: I think it's best to just tell them they have no statements on the statements page and don't give them the option to download

nvonpentz: Though, if they have previous transactions in "all" but none from "this_month" then we'd still be giving them the option to download "this_month".  In that case it'd be good to flash an alert.

This is what it'd look like if they haven't ever had a transaction:
image

app/views/publishers/statement.html.slim
```diff
@@ -0,0 +1,18 @@
...
+      th Date
+      th Description
+      th Channel
+      th Amount (BAT)
+      th Deposit
+    tbody
+      - deposit_sum = BigDecimal.new('0.0e18')
+      - @transactions.each do |transaction|
+        tr
+          td = transaction["created_at"].to_time.strftime("%m/%d/%y")
```
yachtcaptain23: Nitpicking: "%Y-%m-%d" (month/day/year is an American phenomenon.)

app/models/channel.rb
```diff
@@ -129,6 +129,50 @@ def channel_id
...
 
+  # Returns the verified channel with the given channel identifier
+  # e.g.
+  # Channel.find_by_channel_identifier(channel.details.channel_identifier)
+  # => <Channel id: "55ecd577-0425-420f-8796-78598b06c8a0",...,>
+  def self.find_by_channel_identifier(channel_identifier)
+    channel_id_split_on_prefix = channel_identifier.split(":", 2)
+    channel_is_site_channel = channel_id_split_on_prefix.length == 1 # hack to identify site channels
+
+    if channel_is_site_channel
```
yachtcaptain23:
```
if channel_is_site_channel
  channel_details_type_identifier = channel_id_split_on_prefix.first
  return SiteChannelDetails.where(brave_publisher_id: channel_details_type_identifier).
    joins(:channel).
    where('channels.verified = true').first.channel
end

prefix = channel_id_split_on_prefix.first
channel_details_type_identifier = channel_id_split_on_prefix.second
case prefix
...
```

PR #1188, opened 2018-09-26
Do not apply fee twice to last settlement balance

nvonpentz:
The last settlement balance returned by eyeshade in the /wallet response is the amount that was settled, so it already has taken fees into account.  However, when we display a publishers last settlement balance we are applying the fee twice.

This happens in two places:

1. When we calculate the last settlement balance with ruby in PublishersHelper#publisher_humanize_last_settlement and PublishersHelperpublisher_converted_last_settlement

2.  When the default currency is changed, we us JS to get the latest balance and update the DOM. 
since we use `probi` instead of `probi_before_fees` in the wallet JSON that's sent to the dashboard.

https://github.com/brave-intl/publishers/blob/08928c4c12cda5cd0d0d5228ef44ec4f27056fed/app/models/json_builders/wallet_json_builder.rb#L47-L49

This might lead publishers who didn't view their statement to believe they have received less than they actually did.

Thanks to @evq for finding the bug.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/services/publisher_wallet_getter.rb
+++ b/app/services/publisher_wallet_getter.rb
@@ -79,6 +79,12 @@ class PublisherWalletGetter < BaseApiClient
           "EUR" => 0.20187818378874756,
           "GBP" => 0.1799810085548496
         },
+        "lastSettlement"=>
+          {"altcurrency"=>"BAT",
+           "currency"=>"USD",
+           "probi"=>"405520562799219044167",
+           "amount"=>"69.78",
+           "timestamp"=>1536361540000},
         "wallet" => {
             "provider" => "uphold",
             "authorized" => true,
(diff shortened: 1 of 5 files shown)


app/services/publisher_wallet_getter.rb
```diff
@@ -79,14 +79,21 @@ def perform_offline
           "EUR" => 0.20187818378874756,
           "GBP" => 0.1799810085548496
         },
+        "lastSettlement"=>
```
yachtcaptain23: Nit: Placement of the `{`

evq:
lgtm as well, thanks for the fix @nvonpentz !

PR #1189, opened 2018-09-26
add referrals to combined payout report, fix docker test instructions

evq:
This PR closes brave-intl/creators-private-issues#1540 as both contribution and referral payout reports will be generated.

It also includes two fixes that were needed in order for me to run updated tests inside a docker container:
1. Yarn install needed to be run after building the docker image (since we mount the repo at runtime it doesn't make sense to do it in the Dockerfile)
2. The gem we use to autoclean our db refuses to run in our container since it the DATABASE_URL inside uses the container name, not localhost - I've added an env var to override. Ideally we would use the whitelist functionality present in the master version of that gem, however there has not yet been a release since it was added.

Submitter Checklist:

- [x] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests
- [x] Adequate test coverage exists to prevent regressions

Security:
- [x] No raw SQL -- Always prefer ActiveRecord query helpers ([more info on StackOverflow](https://stackoverflow.com/questions/41410752/rails-5-sql-injection#41452695))
- [x] XSS is mitigated -- Avoid `html_safe` and `raw`; escape untrusted content from users and 3rd party APIs ([Rails XSS guide](https://brakemanpro.com/2017/09/08/cross-site-scripting-in-rails) and [OWASP XSS Prevention Cheat Sheet](https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet))

### DIFF
--- a/app/jobs/generate_payout_report_job.rb
+++ b/app/jobs/generate_payout_report_job.rb
@@ -19,6 +19,28 @@ class GeneratePayoutReportJob < ApplicationJob
       publisher_has_unsettled_balance = false
       wallet = publisher.wallet
 
+      probi = wallet.channel_balances[publisher.owner_identifier].probi_before_fees # probi = balance
+      if probi > 0
+        publisher_has_unsettled_balance = true
+
+        if publisher.uphold_verified? && !wallet.address.blank?
+          num_payments += 1
+          total_amount_probi += probi
+
+          payouts.push({
+            "name" => "#{publisher.name}",
+            "altcurrency" => "BAT",
+            "probi" => "#{probi}",
+            "fees" => "0",
+            "authority" => "GeneratePayoutReportJob",
+            "transactionId" => "#{payout_report.id}",
+            "owner" => "#{publisher.owner_identifier}",
+            "type" => "referral",
+            "address" => "#{wallet.address}",
+          })
+        end
+      end
+
       publisher.channels.verified.each do |channel|
         probi = wallet.channel_balances[channel_identifier(channel)].probi # probi = balance - fee
         next if probi <= 0
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ docker-compose up
 
 ### Create the databases
 ```sh
-docker-compose run app rake db:setup; docker-compose run eyeshade-worker sh -c "cd eyeshade && ./bin/migrate-up.sh"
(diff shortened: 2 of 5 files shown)


app/jobs/generate_payout_report_job.rb
```diff
@@ -19,6 +19,28 @@ def perform(should_send_notifications: true, final: true)
       publisher_has_unsettled_balance = false
       wallet = publisher.wallet
 
+      probi = wallet.channel_balances[publisher.owner_identifier].probi_before_fees # probi = balance
+      if probi > 0
```
CoryMcDonald: `probi.positive?` is more ruby-esque but it's not that big of deal if we leave it how it is 😄

app/jobs/generate_payout_report_job.rb
```diff
@@ -19,6 +19,28 @@ def perform(should_send_notifications: true, final: true)
       publisher_has_unsettled_balance = false
       wallet = publisher.wallet
 
+      probi = wallet.channel_balances[publisher.owner_identifier].probi_before_fees # probi = balance
```
CoryMcDonald: We're calling this function down below and there seems to be some duplicate logic, is there a way to simplify this and combine these two branches of logic to make it more DRY?

app/jobs/generate_payout_report_job.rb
evq: good to know! this is my first ruby PR so if there's other ruby conventions I should know I'm all ears :)

CoryMcDonald: Sure thing! Only thing else I see is `!wallet.address.blank?` could alternatively be written as `wallet.address.present?` that way there's not negative logic mixed into the if statement

app/jobs/generate_payout_report_job.rb
nvonpentz: Ideally we would do as you describe and organize the loop such that we iterate through all the `wallet.channel_balances` (which is really account balances, i.e. channel balances + owner balance) instead of once for the owner, and then once for each channel.  However there are some other problems which make that difficult.

I think the logic here is different enough that it justifies the repetition.

evq:
@nvonpentz re: `yarn.lock` I was pretty surprised that the lockfile did not already have checksums. The only thing that comes to mind is that maybe some developers are using outdated versions of yarn?

I think `yarn install` should be used nearly always. Update checksums should only be run if there is a mismatch - but that would be a red flag since it would indicate that content of a released version has changed.

In such cases I think an audit of the changes would be called for to ensure it hasn't been replaced by malicious content.

PR #1193, opened 2018-09-27
[WIP] Remove calls to deprecated eyeshade APIs

### LINKED ISSUE #1014: Change twitch channel identifier to use twitch#author:<CHANNEL_ID> format

### LINKED ISSUE #1184: remove use of deprecated eyeshade APIs
The following APIs can be deprecated now that channel information no longer needs to be synchronized with eyeshade:

`POST /v2/owners`
`DELETE /v1/owners/{owner}/{publisher}`

nvonpentz:
This PR removes two services, the PublisherChannelSetter and the PublisherEyeshadeChannelDeleter.  These services were previously used to keep eyeshade up to date with the mappings between channels and owners.  Since publishers is the now the source of truth for this information, eyeshade no longer needs to store it, so these endpoints will be deprecated.

Since we are no longer using the legacy system, this PR switches to using the `twitch#author:{uuid}` globally.  Also, since the transaction table transition has been smooth, this PR removes the escape hatch that allows us to switch back to the legacy system via an environment variable.

Closes #1184 
Closes #1014 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


nvonpentz:
Still need to add some changes to such that if the request to get the wallet 404s (which will happen to any non uphold verified publisher) we can still display the balance rather than 'Unavailable'.

app/services/publisher_wallet_getter.rb
```diff
@@ -11,22 +11,31 @@ def initialize(publisher:)
...
+      wallet_response = connection.get do |request|
+        request.headers["Authorization"] = api_authorization_header
+        request.url("/v1/owners/#{URI.escape(publisher.owner_identifier)}/wallet")
+      end
 
-    wallet_hash = JSON.parse(wallet_response.body)
+      wallet_hash = JSON.parse(wallet_response.body)
+    rescue Faraday::ResourceNotFound
+      wallet_hash = {}
+    end
```
nvonpentz: Instead of making the request and rescuing upon a 404, I think we could also just set `wallet_hash = {}` if the publisher is not uphold verified and skip the request all together.  Not sure what's better in this case.  We could also do both.

evq: I think skipping the request if not uphold verified sounds slightly better to me

app/jobs/delete_publisher_channel_job.rb
```diff
@@ -1,23 +1,26 @@
...
+    end
+
+    channel_is_verified = @channel.verified?
+
+    if should_update_promo_server
+      referral_code = @channel.promo_registration.referral_code
     end
 
-    success = @channel.destroy
+    success = @channel.destroy!
```
yachtcaptain23: When would destroy fail?

app/jobs/delete_publisher_channel_job.rb
```diff
@@ -1,23 +1,26 @@
...
     if @channel.contested_by_channel.present?
-      Channels::ApproveChannelTransfer.new(channel: @channel, should_delete: false).perform
-    elsif @channel.verification_pending?
-      raise "Can't remove a channel that is contesting another a channel."
+      return Channels::ApproveChannelTransfer.new(channel: @channel).perform
+    end
+
+    channel_is_verified = @channel.verified?
+
+    if should_update_promo_server
```
yachtcaptain23: You can remove this block and call 

```
if success && channel_is_verified && should_update_promo_server
  PromoChannelOwnerUpdater.new(@channel.promo_registration.referral_code).perform
end
```

app/models/channel.rb
```diff
@@ -283,18 +283,11 @@ def verified_duplicate_channels_must_be_contested
         errors.add(:base, "can only contest one channel")
       end
 
-      contesting_channel = duplicate_verified_channels.first
-      if contesting_channel.contested_by_channel_id != self.id
+      contesting_channel = duplicate_verified_channels.first 
+      if (contesting_channel.contested_by_channel_id != self.id) && (contesting_channel.contested_by_channel_id != nil)
+        debugger
```
yachtcaptain23: debugger line

app/services/channels/approve_channel_transfer.rb
```diff
@@ -1,35 +1,28 @@
...
         @channel.save!
-        
+
         original_owner_email = @channel.publisher.email
         original_owner_name = @channel.publisher.name
         channel_name = @channel.publication_title
 
         # Delete the channel from eyeshade and clean up the promo registration
-        DeletePublisherChannelJob.perform_now(channel_id: @channel.id) if @should_delete
+        DeletePublisherChannelJob.perform_now(channel_id: @channel.id)
```
yachtcaptain23: My gut says we should move this line above the ActiveRecord::Transaction. Thoughts?

app/jobs/delete_publisher_channel_job.rb
nvonpentz: @yachtcaptain23 I don't think it will but if theres a problem I'd like to fail hard

PR #1241, opened 2018-10-18
Sanitize HTML tags

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/models/site_banner.rb
+++ b/app/models/site_banner.rb
@@ -15,6 +15,27 @@ class SiteBanner < ApplicationRecord
   BACKGROUND_DIMENSIONS = [900,176]
   BACKGROUND_UNIVERSAL_FILE_SIZE = 70_000 # In bytes
 
+  NUMBER_OF_DONATION_AMOUNTS = 3
+  MAX_DONATION_AMOUNT = 20
+
+  validates_presence_of :title, :description, :donation_amounts, :default_donation, :publisher
+  validate :donation_amounts_in_scope
+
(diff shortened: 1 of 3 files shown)


app/models/site_banner.rb
```diff
@@ -15,6 +15,38 @@ class SiteBanner < ApplicationRecord
...
+
+  def social_links_validation
+    return if errors.present? || social_links.nil?
+    if (social_links&.keys - ["twitch", "youtube", "twitter"]).present?
+      errors.add(:base, "Unacceptable additional social links")
+    end
+    errors.add(:base, "Invalid twitch link") unless social_links["twitch"].blank? || /^http(|s):\/\/twitch.tv\/[A-Za-z]*$/.match(social_links["twitch"])
+    errors.add(:base, "Invalid youtube user or channel link") unless social_links["youtube"].blank? || /^http(|s):\/\/youtube\.com\/(channel|user)\/[A-Za-z]*$/.match(social_links["youtube"])
+    errors.add(:base, "Invalid twitter user link") unless social_links["twitter"].blank? || /^http(|s):\/\/twitter\.com\/[A-Za-z]*$/.match(social_links["twitter"])
+  end
```
nvonpentz: Do we allow users to enter `www.`?  We may want to to use the the domain normalization gem(s) that we use for site channel verification instead of regular expressions where possible.

Issue #1113, opened 2018-08-24
Allow admins to generate infinite referral codes


yachtcaptain23:
### Motivation
Currently, the referral promo implementation does not allow publishers to generate codes without channels.  This makes it difficult to A/B test the effectiveness of sharing the codes across multiple mediums or time periods.

Furthermore, publishers are only presented with two statistics "Downloads" and "Confirmed" which do not represent a complete life cycle of a referral code.

In order to give publishers and Brave employees the tools they need to most effectively use/judge our referral promo, we will allow of infinite referral code generation, and present all the insights available.  First we will support this from the admin dashboard only, and later will support this for all publishers.

### Proposed admin dashboard UI
![image](https://user-images.githubusercontent.com/12549658/46040878-79e39c80-c0df-11e8-9ddc-a0f6cd0e72cd.png)
There are a few cases the design does not cover.  Namely there is no way to differentiate a pause for a campaign and a pause for an individual code.  We need to have be able to filter by campaign and select all referral codes.  No operations will be done on campaigns (except campaign creation), instead operations will be done on all the codes within a campaign.

Additionally, the design does not discuss statement generation.  Statement generation will be very similar to pausing in that it can happen per code or per campaign via select by campaign feature.  Ideally admins can specify price per download and price per install and this will be reflected in the report.

### Backend Changes
* **DONE** Add a `type` column to the promo_registration table that is one of `channel`, `owner`,  or `unattached` 
* **DONE** Create rake task to update all existing PromoRegistrations to have kind "channel"
* **DONE** Relax constraints on the PromoRegistration model.
    * **DONE** A PromoRegistration should only validate `channel_id` or `publisher_id` if the type of the code is `channel` or `owner` respectively.
    * **DONE** Add a migration to relax the database level constraint requiring a channel_id for a a promo registration.
* **WILL BE COMPLETED LATER** Add foreign key for a publisher_id to the PromoRegistration.  This will allow PromoRegistrations to be owned by publishers. (this is not required for unattached codes)
* **DONE** Add PromoCampaigns
  * **DONE** Create `promo_campaigns` table and model
  * **DONE** Add foreign_key to `promo_campaign_id` to promo_registrations table, update model
* **DONE** Create a new the UnattachedPromoRegistrar service to hit the new endpoint for unattached codes.  This endpoint will accept an argument for how many codes to generate.
* **DONE** Create a AdminPromoStatusUpdater service  which updates the status of a list of referral codes
[...]

PR #1265, opened 2018-10-22
Add admin interface for referral code creation / management / reporting

nvonpentz:
Resolves https://github.com/brave-intl/publishers/issues/1113

The set_default_promo_registration_kind rake task should be run as soon as the code is merged.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/admin/unattached_promo_registrations_controller.rb
+++ b/app/controllers/admin/unattached_promo_registrations_controller.rb
@@ -44,7 +44,7 @@ class Admin::UnattachedPromoRegistrationsController < AdminController
     @report_contents = report_info["contents"]
 
     report_string = render_to_string :layout => false
-    send_data report_string, filename: "BraveReferralPromoStatement.html", type: "application/html"
+    send_data report_string, filename: "BraveReferralPromoReport.html", type: "application/html"
   end
 
   def update_statuses
--- a/app/controllers/admin/unattached_promo_registrations_controller.rb
+++ b/app/controllers/admin/unattached_promo_registrations_controller.rb
@@ -24,7 +24,7 @@ class Admin::UnattachedPromoRegistrationsController < AdminController
     redirect_to admin_unattached_promo_registrations_path, notice: "#{number} codes created."
   end
 
-  def statement
+  def report
     referral_codes = params[:referral_codes]
     @reporting_interval = params[:reporting_interval]
     @event_types = params[:event_types]
@@ -33,18 +33,18 @@ class Admin::UnattachedPromoRegistrationsController < AdminController
                         alert: "Please check at least one of downloads, installs, or confirmations."
     end
     
-    statement_start_and_end_date = parse_statement_dates(params[:referral_code_statement_period], @reporting_interval)
-    statement_info = PromoStatementGenerator.new(referral_codes: referral_codes,
-                                                  start_date: statement_start_and_end_date[:start_date],
-                                                  end_date: statement_start_and_end_date[:end_date],
+    report_start_and_end_date = parse_report_dates(params[:referral_code_report_period], @reporting_interval)
+    report_info = PromoReportGenerator.new(referral_codes: referral_codes,
+                                                  start_date: report_start_and_end_date[:start_date],
+                                                  end_date: report_start_and_end_date[:end_date],
                                                   reporting_interval: @reporting_interval).perform
 
-    @start_date = statement_info["start_date"]
-    @end_date = statement_info["end_date"]
-    @statement_contents = statement_info["contents"]
+    @start_date = report_info["start_date"]
+    @end_date = report_info["end_date"]
+    @report_contents = report_info["contents"]
 
-    statement_string = render_to_string :layout => false
-    send_data statement_string, filename: "BraveReferralPromoStatement.html", type: "application/html"
+    report_string = render_to_string :layout => false
+    send_data report_string, filename: "BraveReferralPromoStatement.html", type: "application/html"
   end
 
   def update_statuses
@@ -67,13 +67,13 @@ class Admin::UnattachedPromoRegistrationsController < AdminController
 
   private
 
-  def parse_statement_dates(statement_period, reporting_interval)
-    start_date = Date.new(statement_period["start(1i)"].to_i,
-                          statement_period["start(2i)"].to_i,
-                          statement_period["start(3i)"].to_i)
-    end_date = Date.new(statement_period["end(1i)"].to_i,
-                        statement_period["end(2i)"].to_i,
-                        statement_period["end(3i)"].to_i)
+  def parse_report_dates(report_period, reporting_interval)
+    start_date = Date.new(report_period["start(1i)"].to_i,
+                          report_period["start(2i)"].to_i,
+                          report_period["start(3i)"].to_i)
+    end_date = Date.new(report_period["end(1i)"].to_i,
+                        report_period["end(2i)"].to_i,
+                        report_period["end(3i)"].to_i)
     {
       start_date: start_date,
       end_date: end_date
--- a/app/controllers/admin/unattached_promo_registrations_controller.rb
+++ b/app/controllers/admin/unattached_promo_registrations_controller.rb
@@ -5,12 +5,12 @@ class Admin::UnattachedPromoRegistrationsController < AdminController
     filter = params[:filter]
     case filter
     when "All codes", nil, ""
-      @promo_registrations = PromoRegistration.where(kind: "unattached").order("created_at DESC")
+      @promo_registrations = PromoRegistration.unattached.order("created_at DESC")
     when "Not assigned"
-      @promo_registrations = PromoRegistration.where(kind: "unattached").where(promo_campaign_id: nil).order("created_at DESC")
+      @promo_registrations = PromoRegistration.unattached.where(promo_campaign_id: nil).order("created_at DESC")
     else
       @promo_registrations = PromoRegistration.joins(:promo_campaign).
-                                               where(kind: "unattached").
+                                               unattached.
                                                where(promo_campaigns: {name: filter}).
                                                order("created_at DESC")
     end
--- a/app/controllers/admin/unattached_promo_registrations_controller.rb
+++ b/app/controllers/admin/unattached_promo_registrations_controller.rb
@@ -1,0 +1,86 @@
+class Admin::UnattachedPromoRegistrationsController < AdminController
+  include PromosHelper
+
+  def index
+    filter = params[:filter]
+    case filter
+    when "All codes", nil, ""
+      @promo_registrations = PromoRegistration.where(kind: "unattached").order("created_at DESC")
+    when "Not assigned"
+      @promo_registrations = PromoRegistration.where(kind: "unattached").where(promo_campaign_id: nil).order("created_at DESC")
+    else
+      @promo_registrations = PromoRegistration.joins(:promo_campaign).
+                                               where(kind: "unattached").
+                                               where(promo_campaigns: {name: filter}).
+                                               order("created_at DESC")
+    end
+    @current_campaign = params[:filter] || "All codes"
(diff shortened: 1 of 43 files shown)


db/migrate/20181015023430_create_promo_campaigns.rb
```diff
@@ -0,0 +1,8 @@
+class CreatePromoCampaigns < ActiveRecord::Migration[5.2]
+  def change
+    create_table :promo_campaigns, id: :uuid,default: -> { "uuid_generate_v4()" }, force: :cascade do |t|
+      t.text :name
```
yachtcaptain23: You'll want to add a uniqueness index, https://stackoverflow.com/questions/45826033/how-do-i-create-a-unique-index-within-a-rails-create-table-migration

app/models/promo_campaign.rb
```diff
@@ -0,0 +1,4 @@
+class PromoCampaign < ApplicationRecord
+  has_many :promo_registrations
+  validates :name, uniqueness: true
```
yachtcaptain23: Would you add another validation that makes sure you clear out whitespace from the front and back and maybe middle? In your controller you're passing directly what admins are typing.

app/models/promo_registration.rb
```diff
@@ -1,11 +1,38 @@
 class PromoRegistration < ApplicationRecord
   has_paper_trail
+  # self.per_page = 15
+
+  KINDS = ["channel", "owner", "unattached"].freeze
```
yachtcaptain23: nit:

```
CHANNEL = "channel".freeze
OWNER = "owner".freeze
UNATTACHED = "unattached".freeze
KINDS = [CHANNEL, OWNER, UNATTACHED].freeze
```

app/models/promo_registration.rb
```diff
@@ -1,11 +1,38 @@
...
   
   belongs_to :channel, validate: true, autosave: true
+  belongs_to :promo_campaign
 
-  validates :channel_id, presence: true
-
+  validates :channel_id, presence: true, if: -> { kind == "channel"}
```
yachtcaptain23: ?

app/models/promo_registration.rb
yachtcaptain23: Would you also add comments to explain what these mean (from a perspective of us looking back at this code 1 year from now)

app/models/promo_registration.rb
```diff
@@ -1,11 +1,38 @@
...
+  validates :channel_id, presence: false, if: -> { kind == "channel"}
+  
   validates :promo_id, presence: true
 
+  validates :kind, presence: true
```
yachtcaptain23: You can track multiple `presence` fields with the following example:
`validates_presence_of :title, :description, :donation_amounts, :default_donation, :publisher`

app/services/promo_statement_generator.rb
```diff
@@ -0,0 +1,114 @@
...
+    promo_registrations.each do |promo_registration|
+      # Pull all statistics associated with a code
+      events = JSON.parse(promo_registration.stats)
+
+      # Select only those within date range supplied
+      events_within_statement_period = events.select { |event|
+        (event["ymd"].to_date >= @start_date) && (event["ymd"].to_date <= @end_date)
+      }
+
+      if @reporting_interval == "culmulative"
```
yachtcaptain23: Possible mispelling?

app/services/promo_statement_generator.rb
```diff
@@ -0,0 +1,114 @@
...
+        base_case = {"retrievals" => 0, "first_runs" => 0, "finalized" => 0 }
+        culmative_statement_contents = events_within_statement_period.reduce(base_case) { |aggregate_stats, event|
```
yachtcaptain23: `cumulative` IIRC

app/models/promo_registration.rb
```diff
@@ -1,11 +1,38 @@
...
+      aggregate_stats["finalized"] += event_date["finalized"]
+      aggregate_stats.slice("retrievals", "first_runs", "finalized")
+    }
+  end
+
+  private
+
+  def kind_is_valid
+    if kind.present? && KINDS.exclude?(kind)
+      errors.add(:kind, "Invalid promo registration kind #{kind}")
```
yachtcaptain23: Use https://guides.rubyonrails.org/active_record_validations.html#inclusion instead please

db/migrate/20181015023430_create_promo_campaigns.rb
nvonpentz: I thought I added this.  Thanks

app/models/promo_campaign.rb
nvonpentz: Yes

app/models/promo_registration.rb
nvonpentz: re nit:  What's the advantage of setting the constants individually?

I will add comments.

app/models/promo_registration.rb
nvonpentz: Good catch, will remove

app/models/promo_registration.rb
yachtcaptain23: It's just that when we use strings like "unattached" elsewhere in the code, it's not 100% certain that it's right (e.g. what if someone says "detached" or "dangling" etc.). Using misnamed constants will appropriately throw an exception in runtime due to the constant not being found.

app/views/admin/shared/_sidebar.html.slim
```diff
@@ -14,4 +14,7 @@ aside
               span FAQs
           li 
             = link_to admin_payout_reports_path
-              span Payout Reports
\ No newline at end of file
+              span Payout Reports
+          li 
+            = link_to admin_unattached_promo_registrations_path
+              span Referral Promo
```
yachtcaptain23: Would you call it `Referral Promos`? It's an index of all of them

app/views/admin/unattached_promo_registrations/index.html.slim
```diff
@@ -0,0 +1,77 @@
+.unattached-promo-registrations
+  h1 Referral Promo
```
yachtcaptain23: Ditto here to change to `Referral Promos`

app/views/admin/shared/_sidebar.html.slim
nvonpentz: Hm not sure `Referral Promos` is right here.  We're running one promotion, in which users and admins can generate multiple referral codes.

yachtcaptain23: Perhaps call it Referral Campaigns then?

yachtcaptain23: (I'm nit-picking lol. not too big of a deal)

app/controllers/admin/unattached_promo_registrations_controller.rb
```diff
@@ -0,0 +1,86 @@
+class Admin::UnattachedPromoRegistrationsController < AdminController
+  include PromosHelper
+
+  def index
+    filter = params[:filter]
+    case filter
+    when "All codes", nil, ""
+      @promo_registrations = PromoRegistration.where(kind: "unattached").order("created_at DESC")
```
yachtcaptain23: preferably you create a scope called `unattached`

app/views/admin/unattached_promo_registrations/index.html.slim
```diff
@@ -0,0 +1,100 @@
...
+      = text_field_tag "campaign_name", nil, placeholder: "Campaign name"
+      = submit_tag "+ Campaign", class: "btn btn-info"
+
+  hr
+
+  = form_tag admin_unattached_promo_registrations_path, method: :get, id: "filter-by-campaign"do
+    = select_tag :filter, options_for_select(@campaigns + ["Not assigned", "All codes"], @current_campaign)
+    = submit_tag "Filter by campaign", class: "btn btn-success"
+  = form_tag admin_unattached_promo_registrations_path, method: :patch, id: "unattached-referral-code-form" do
+    table.table.table-bordered.table-striped
```
yachtcaptain23: Would you add a comment somewhere of when the stats were last synced? You should store that timestamp value in Rails.cache (aka Redis)

yachtcaptain23: And also add a comment saying that stats are adjusted every 12 hours (ideally one can read from the config file but I don't know how to do this.)

app/views/admin/unattached_promo_registrations/index.html.slim
```diff
@@ -0,0 +1,100 @@
...
+              span.tf-tooltip-content-content== "A confirmation is counted when a user downloads the browser and opens, then opens it again at least 30 days later."
+      tbody
+        - @promo_registrations.each do |promo_registration| 
+          - promo_registration_aggregate_stats = promo_registration.aggregate_stats
+          tr
+            td = check_box_tag "referral_codes[]", "#{promo_registration.referral_code}", 1
+            td = promo_registration.referral_code
+            td = promo_registration.promo_campaign&.name
+            td = promo_registration.active ? "active" : "paused"
+            td = promo_registration_aggregate_stats["retrievals"] || 0
```
yachtcaptain23: Would you make all the magic strings that are essentially states into constants in a model? I.e. in the `promo_registrations` model, please use FIRST_RUNS = "first_runs" so we can use `PromoRegistration::FIRST_RUNS` instead of another dev needing to painstakingly figure out if it's `first_run` or `first_runs`

https://softwareengineering.stackexchange.com/questions/365339/what-is-wrong-with-magic-strings

app/views/admin/unattached_promo_registrations/index.html.slim
nvonpentz: Yes the stats are sync'd every 12 hours for all the unattached codes, but they are also sync'd when an admin generates a statement for the codes included in that statement.  

https://github.com/brave-intl/publishers/pull/1265/files#diff-a548229142426d69e713710d88d3fe0bR13

That admins are guaranteed that the stats in the statements are up to date.

I'll look into storing and displaying when codes were last sync'd for the index view.  We may want to do this for the publisher dashboard as as well since I know we get a lot of questions about that.

app/helpers/promos_helper.rb
```diff
@@ -80,4 +80,45 @@ def on_channel_type(channel)
       raise
     end
   end
+
+  def reporting_interval_column_header(reporting_interval)
+    case reporting_interval
+    when "by_day"
```
yachtcaptain23: I'd prefer `by_day`, `by_week`, `by_month`, `cumulative` to be stored as constants. It's easy to misspell and get wrong (i.e. the convention is daily, weekly, monthly, running_total)

app/services/promo_statement_generator.rb
```diff
@@ -0,0 +1,114 @@
...
+        base_case = {PromoRegistration::RETRIEVALS => 0, PromoRegistration::FIRST_RUNS => 0, "finalized" => 0 }
+        culmative_statement_contents = events_within_statement_period.reduce(base_case) { |aggregate_stats, event|
+          aggregate_stats[PromoRegistration::RETRIEVALS] += event[PromoRegistration::RETRIEVALS]
+          aggregate_stats[PromoRegistration::FIRST_RUNS] += event[PromoRegistration::FIRST_RUNS]
+          aggregate_stats[PromoRegistration::FINALIZED] += event[PromoRegistration::FINALIZED]
+          aggregate_stats.slice(PromoRegistration::RETRIEVALS, PromoRegistration::FIRST_RUNS, PromoRegistration::FINALIZED)
+        }
+
+        statement_contents_for_referral_code = {
+          @start_date => culmative_statement_contents
```
yachtcaptain23: misspelling here too

app/helpers/promos_helper.rb
nvonpentz: Yeah, BD asked for 'by week' etc, but I will change since I think that's better.

PR #1271, opened 2018-10-25
Prepare infinity code dashboard for release

nvonpentz:
This PR allows the dashboard to load in the case when the last stats updated was not stored in redis.  This value is only set by the AdminSyncPromoStats job, which can only complete after a code has been created.  This change will simplify deploying the code because we will not need to create a code via rails console, then run the AdminSyncPromoStats job via rails console before we can use the feature.

Also this PR restricts the number of codes that an admin can create to 50 per click.  Currently it is unbounded and thus admins have the ability to accidentally crash the system and enter huge amounts of data directly into two databases.

Lastly, this PR also iterates on the infinity codes admin dashboard frontend.  After this PR it will look like:

image

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/views/admin/unattached_promo_registrations/index.html.slim
+++ b/app/views/admin/unattached_promo_registrations/index.html.slim
@@ -9,14 +9,14 @@
         hr
         .
           = number_field_tag "number_of_codes_to_create", nil, placeholder: "# of codes"
-          = submit_tag "Create", class: "btn btn-info"
+          = submit_tag "Create", class: "btn btn-info", style: "float:right;"
     .panel.panel--double
       = form_tag admin_promo_campaigns_path, id: "create-promo-campaign" do
         .panel--header = "Create campaigns"
         hr
         .
           = text_field_tag "campaign_name", nil, placeholder: "Campaign name"
-          = submit_tag "Create", class: "btn btn-info"
+          = submit_tag "Create", class: "btn btn-info", style: "float: right"
   .row
   .panel.panel--white.panel--table
     .panel--header = "Manage"
--- a/app/views/admin/unattached_promo_registrations/index.html.slim
+++ b/app/views/admin/unattached_promo_registrations/index.html.slim
@@ -20,7 +20,11 @@
   .row
   .panel.panel--white.panel--table
     .panel--header = "Manage"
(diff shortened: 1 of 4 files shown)


app/views/admin/unattached_promo_registrations/index.html.slim
```diff
@@ -1,110 +1,113 @@
...
+        = "Stats last refreshed #{distance_of_time_in_words(Time.now - Rails.cache.fetch('unattached_promo_registration_stats_last_synced_at'))} ago. Stats in reports are always up-to-date."
+    = form_tag admin_unattached_promo_registrations_path, method: :get, id: "filter-by-campaign" do
+      = select_tag :filter, options_for_select(@campaigns + ["Not assigned", "All codes"], @current_campaign)
+      = submit_tag "Filter by campaign", class: "btn btn-success"
+    = form_tag admin_unattached_promo_registrations_path, method: :patch, id: "unattached-referral-code-form" do
+      table.table
+        tr
+          th 
+          th = "Code"
+          th data-toggle="tooltip" data-placement="top" title="Tooltip"= "Campaign"
```
yachtcaptain23: What does `title="Tooltip"= "Campaign"` do?

nvonpentz: leftover from an attempt to use bootstrap tooltips. can remove

app/views/admin/unattached_promo_registrations/index.html.slim
```diff
@@ -1,110 +1,113 @@
...
+                = check_box_tag "event_types[]", PromoRegistration::RETRIEVALS
+                = label_tag = " " + event_type_column_header(PromoRegistration::RETRIEVALS)
+              .
+                = check_box_tag "event_types[]", PromoRegistration::FIRST_RUNS
+                = label_tag = " " + event_type_column_header(PromoRegistration::FIRST_RUNS)
+              .
+                = check_box_tag "event_types[]", PromoRegistration::FINALIZED
+                = label_tag = " " + event_type_column_header(PromoRegistration::FINALIZED)
+            .flex-one.align-self-flex-end
+              = submit_tag "Download", id: "download-referral-reports", class: "btn btn-primary", data: { disable_with: false }
```
yachtcaptain23: https://api.rubyonrails.org/v5.1/classes/ActionView/Helpers/FormTagHelper.html

Preferably we'd replace `disable_with: 'Please wait'`

yachtcaptain23: Actually nvm, it's fine

PR #1294, opened 2018-10-30
Update promo registration flow for channels that have previously registered

nvonpentz:
Resolves https://github.com/brave-intl/publishers/issues/1291

Blocked by https://github.com/brave-intl/vault-promo-services/pull/56
More information found in https://github.com/brave-intl/vault-promo-services/pull/56.

* Update the url used by the PromoRegistrationGetter - this ensures
  the referral code owned by the channel_id supplied is returned

* No longer update the referral server when a channel has changed
  owners via the PromoRegistrationsGetter

* Allow publishers to remove verified channels

* Remove dead tests

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/services/promo_registrar.rb
+++ b/app/services/promo_registrar.rb
@@ -33,8 +33,6 @@ class PromoRegistrar < BaseApiClient
       request.body = request_body(channel)
       request.url("/api/1/promo/publishers")
     end
-    
-    referral_code = JSON.parse(response.body)["referral_code"]
 
     {
       referral_code: JSON.parse(response.body)["referral_code"],
@@ -43,17 +41,13 @@ class PromoRegistrar < BaseApiClient
   rescue Faraday::Error => e
     if e.response[:status] == 409
       Rails.logger.warn("PromoRegistrar #register_channel returned 409, channel already registered.  Using PromoRegistrationGetter to get the referral_code.")
-      Rails.logger.info("Attempted to register channel #{channel.details.channel_identifier}, but it already registered.  Saving referral code #{referral_code}.")
       
       # Get the referral code
       registration = PromoRegistrationGetter.new(publisher: @publisher, channel: channel).perform
-      
-      # Look at the owner_id that Promo has assigned to the referral, and determine whether to update
-      should_update_promo_server = registration["owner_id"] != @publisher.id ? true : false
 
       {
         referral_code: registration["referral_code"],
-        should_update_promo_server: should_update_promo_server
+        should_update_promo_server: registration["owner_id"] != @publisher.id ? true : false
       }
     else
       require "sentry-raven"
--- a/app/services/promo_registrar.rb
+++ b/app/services/promo_registrar.rb
@@ -5,17 +5,21 @@ class PromoRegistrar < BaseApiClient
   def initialize(publisher:, promo_id: active_promo_id)
     @publisher = publisher
     @promo_id = promo_id
+    @should_update_promo_server = false
   end
 
   def perform
     channels = @publisher.channels.where(verified: true)
-
     channels.each do |channel|
       if should_register_channel?(channel)
         referral_code = register_channel(channel)
         if referral_code.present?
           promo_registration = PromoRegistration.new(channel_id: channel.id, promo_id: @promo_id, kind: "channel", referral_code: referral_code)
-          promo_registration.save!
+          success = promo_registration.save!
+          if success && @should_update_promo_server
+            PromoChannelOwnerUpdater.new(publisher_id: @publisher.id, referral_code: referral_code).perform
+          end
+          @should_update_promo_server = false
         end
       end
     end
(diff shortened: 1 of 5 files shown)


test/services/promo_registrar_test.rb
```diff
@@ -63,4 +57,7 @@ class PromoRegistrarTest < ActiveJob::TestCase
       PromoRegistrar.new(publisher: publisher).perform
     end
   end
+
+  test "" do
```
yachtcaptain23: Deadcode?

yachtcaptain23:
Does the following need to be updated to reflect the new url?

```
class PromoChannelOwnerUpdater < BaseApiClient
  include PromosHelper

  def initialize(publisher_id: "removed", referral_code:)
    @publisher_id = publisher_id
    @referral_code = referral_code # The brave_publisher_id or youtube channel id, not uuid
  end

  def perform
    return perform_offline if perform_promo_offline?
    return nil if @referral_code.nil?
    response = connection.put do |request|
      request.url("/api/1/promo/publishers/#{@referral_code}")
      request.body = { "owner_id": @publisher_id }.to_json
    end
  end
```

app/services/promo_registrar.rb
```diff
@@ -35,8 +39,14 @@ def register_channel(channel)
     if e.response[:status] == 409
       Rails.logger.warn("PromoRegistrar #register_channel returned 409, channel already registered.  Using PromoRegistrationGetter to get the referral_code.")
       Rails.logger.info("Attempted to register channel #{channel.details.channel_identifier}, but it already registered.  Saving referral code #{referral_code}.")
-      referral_code = PromoRegistrationGetter.new(publisher: @publisher, channel: channel).perform
-      referral_code
+      # Get the referral code
+      registration = PromoRegistrationGetter.new(publisher: @publisher, channel: channel).perform
+
+      if registration["owner_id"] != @publisher.id
```
yachtcaptain23: Nitpicking as this is a best practice: `register_channel(channel)`'s currently returning `referral_code` as part of the return type and `@should_update_promo_server` as an instance_variable.  This can be improved where the return type is a hash, that way we don't have code that passes a private variable as an instance variable. 

e.g. `{referral_code: referral_code, should_update_promo_server: false}`

test/services/promo_registrar_test.rb
nvonpentz: Hm, I should probably write some more tests.

app/services/promo_registrar.rb
nvonpentz: I think this is a good idea, will change.

nvonpentz:
> Does the following need to be updated to reflect the new url?

@yachtcaptain23 No it does not.  The url for the PromoRegistrationGetter has changed, the new endpoint will return the referral code so it can be saved in publishers (otherwise it doesn't know the combination of letters and numbers that make up the referral code for this channel).

The only changes to the endpoint used by the PromoChannelOwnerUpdater is that the new `owner_id` supplied will be recorded in both the promo server's referral_codes table and the promo server's publishers table https://github.com/brave-intl/vault-promo-services/pull/56/files#diff-63026f276bb9547ee75687884ce8035fR48. The url stays the same.

nvonpentz:
Rebased with staging to remove the gem vuln to get ci to pass.

app/services/promo_registrar.rb
```diff
@@ -29,14 +33,28 @@ def register_channel(channel)
...
+      
+      # Get the referral code
+      registration = PromoRegistrationGetter.new(publisher: @publisher, channel: channel).perform
+      
+      # Look at the owner_id that Promo has assigned to the referral, and determine whether to update
+      should_update_promo_server = registration["owner_id"] != @publisher.id ? true : false
+
+      {
+        referral_code: registration["referral_code"],
+        should_update_promo_server: should_update_promo_server
```
yachtcaptain23: Preferably have fewer lines by doing `should_update_promo_server: registration["owner_id"] != @publisher.id`

app/services/promo_registrar.rb
```diff
@@ -29,14 +33,28 @@ def register_channel(channel)
...
       Rails.logger.warn("PromoRegistrar #register_channel returned 409, channel already registered.  Using PromoRegistrationGetter to get the referral_code.")
       Rails.logger.info("Attempted to register channel #{channel.details.channel_identifier}, but it already registered.  Saving referral code #{referral_code}.")
-      referral_code = PromoRegistrationGetter.new(publisher: @publisher, channel: channel).perform
-      referral_code
+      
+      # Get the referral code
+      registration = PromoRegistrationGetter.new(publisher: @publisher, channel: channel).perform
+      
+      # Look at the owner_id that Promo has assigned to the referral, and determine whether to update
+      should_update_promo_server = registration["owner_id"] != @publisher.id ? true : false
```
yachtcaptain23: *mind expansion meme*

PR #1296, opened 2018-10-30
Feature/stats-api-channels

### LINKED ISSUE #896: expose statistics about channels

dlipeles:
#### API for stats, v1

* `api/v1/stats/channels`
`returns list of channel id's`

* `api/v1/stats/channels/:channel_id`
`returns the following for a given channel :`
```
{
  "channel_id": 
  "channel_identifier": 
  "channel_type": 
  "name": 
  "stats": 
  "url": 
  "owner_id": 
  "created_at": 
  "verified": 
}
```
* `api/v1/stats/channels/:uuid` returns `404` if channel not found
* `api/v1/stats/` + `signups_per_day`, `email_verified_signups_per_day`, `etc...`
`returns results of the query`

##### Tests 

Found at `/test/controllers/api/v1/stats/channels_controller_test.rb`

Closes #896


app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+        channel_type: channel_type,
+        name: channel_name,
+        stats: channel_details.stats,
+        url: channel_details.url,
+        owner_id: channel.publisher.owner_identifier,
+        created_at: channel.created_at,
+        verified: channel.verified
+      }
+
+      render(status: 200, json: data)
```
yachtcaptain23: You need `render(...) and return`. `render` doesn't return out of a function

app/controllers/api/v1/stats/statistics_controller.rb
```diff
@@ -1,4 +1,5 @@
-class Api::StatsController < Api::BaseController
+class Api::V1::Stats::StatisticsController < Api::BaseController
```
yachtcaptain23: Would you rename to `Api::V1::Stats::PublishersController`?

app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -0,0 +1,62 @@
+class Api::V1::Stats::ChannelsController < Api::BaseController
+
+  def index
+
```
corymcdonald: Typically Rails won't have these controllers with new lines in the beginning and end.  I think we should introduce some linting rules to prevent this? That's probably something that i'll add when I join the team 😄

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Added ✔️

app/controllers/api/v1/stats/statistics_controller.rb
dlipeles: We should keep this as statistics for now as there are sections that use only `channels` and not `publishers`. 

ie, `youtube_channels_by_view_count`, `twitch_channels_by_view_count`

I would be willing to do a refactor later on when we add more stats and some patterns emerge

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Added ✔️ , that sounds good.

app/controllers/api/v1/stats/statistics_controller.rb
yachtcaptain23: I'm okay with leaving `youtube_channels_by_view_count` and moving it to `channels_controller`. `Stats::StatisticsController` sounds redundant, rather `Stats::MiscellaneousController` is more descriptive.

PR #1310, opened 2018-11-05
Source lastSettlement information from the /transactions endpoint

nvonpentz:
Instead of sourcing the lastSettlement information from the /wallet endpoint (which relies on legacy mongo documents), determine the date and amount of the lastSettlement from the list of all transactions supplied in the /transactions endpoint (which relies on the postgres transactions table).

The logic that determines the last settlement amount and last settlement date can be found in PublisherWalletGetter#last_settlement_balance.  The settlement date is determined by looking through all the transactions and choosing the date of the most recent transaction where "settlement_amount" is present.  Then we determine the last settlement amount by taking the sum of all the settlement_amounts for all the transactions that occur in the month of the last settlement settlement date.

We do not subtract a fee from the last settlement amount.

Since the we now hit the /transactions endpoint when we run the PublisherWalletGetter, I had to stub the /transactions response in all existing tests that use the wallet.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


lib/eyeshade/wallet.rb
```diff
@@ -38,9 +38,9 @@ def initialize(wallet_json:, channel_json:)
         @channel_balances[identifier] = Eyeshade::Balance.new(balance_json: json)
       end
 
-      if wallet_json["lastSettlement"]
+      if wallet_json["lastSettlement"].present?
         @last_settlement_balance = Eyeshade::Balance.new(balance_json: wallet_json["lastSettlement"].merge({'rates' => @rates}), apply_fee: false)
-        @last_settlement_date = Time.at(wallet_json["lastSettlement"]["timestamp"]/1000).to_datetime
+        @last_settlement_date = Time.at(wallet_json["lastSettlement"]["timestamp"]).to_datetime
```
nvonpentz: I'm not sure why we dividing by 1000 previously

app/services/publisher_wallet_getter.rb
```diff
@@ -140,6 +147,36 @@ def total_balance_bat(accounts)
     accounts.map {|account| account["balance"].to_d }.reduce(0, :+)
   end
 
+  def last_settlement_balance(transactions)
```
nvonpentz: I'm not sure if it's better to have this method in the PublisherWalletGetter like it is now,or if I should add a `last_settlement_balance` boolean parameter to the PublisherTransactionsGetter, and have this be a private method in the PublisherTransactionsGetter instead.

lib/eyeshade/wallet.rb
evq: nodejs timestamps are usually in milliseconds - that probably got exposed over the api

app/services/publisher_wallet_getter.rb
```diff
@@ -140,6 +147,36 @@ def total_balance_bat(accounts)
...
+    return {} if transactions == []
+    last_settlement_date = transactions.reduce(Date.new(0)) { |last_settlement_date, transaction|
```
yachtcaptain23: This is convoluted from a readability standpoint. Would you simplify it?

nvonpentz: I could do this this imperatively with a loop
```ruby
last_settlement_date = Date.new(0)
transactions.each do |transaction|
  if transaction["created_at"].to_date > last_settlement_date && transaction["settlement_amount"].present?
    last_settlement_date = transaction["created_at"].to_date
  end
end
```

or I could separate the logic that selects only settlement transactions

```ruby
last_settlement_date = transactions.select { |transactions|
  transaction["settlement_amount"].present?
}.reduce(Date.new(0)) { |last_settlement_date, transaction|
  if transaction["created_at"].to_date > last_settlement_date
    last_settlement_date = transaction["created_at"].to_date
  end
  last_settlement_date
}
```

or I could do a little bit of both

```ruby
transactions = transactions.select { |transactions|
  transaction["settlement_amount"].present?
}

last_settlement_date = Date.new(0)
transactions.each do |transaction|
  if transaction["created_at"].to_date > last_settlement_date
    last_settlement_date = transaction["created_at"].to_date
  end
end

```

Let me know what you'd prefer @yachtcaptain23

evq: you might be able to cheat a bit since transactions is ordered by `created_at` when returned from eyeshade

maybe something like this?
```
last_settlement = transactions.lazy.reverse_each().select{|t| 
  t["settlement_amount"].present?
}.map{|t| t["created_at"]}.first() || Date.new(0)
```

app/services/publisher_wallet_getter.rb
yachtcaptain23: Oh I nitpicking and meant to say we could merge the two blocks below since you're iterating over the same array twice.

app/services/publisher_transactions_getter.rb
```diff
@@ -87,7 +89,7 @@ def perform_offline
           "channel" => "#{channel.details.channel_identifier}",
           "amount" => "#{contribution_settlement_amount}",
           "settlement_currency" => publisher.default_currency,
-          "settlement_amount" => "56.81",
+          "settlement_amount" => OFFLINE_CONTRIBUTION_SETTLEMENT_AMOUNT,
```
corymcdonald: Did you mix these variables up? I think this is supposed to be the `OFFLINE_REFERRAL_SETTLEMENT_AMOUNT` and the other one is supposed to be the `18.18` value

app/services/publisher_wallet_getter.rb
```diff
@@ -43,6 +45,12 @@ def perform
       channel_hash = {}
     end
 
+    # Get last settlement balance from transactions endpoint
+    last_settlement_balance = last_settlement_balance(PublisherTransactionsGetter.new(publisher: @publisher).perform)
+    unless last_settlement_balance.empty? || (last_settlement_balance["amount"].to_i == 0)
```
corymcdonald: nit:` last_settlement_balance["amount"].to_i.zero?`

app/services/publisher_transactions_getter.rb
nvonpentz: Good catch, I did switch them.  I don't think it makes a difference since the totals still add up, but I will revert just in case.

PR #1316, opened 2018-11-06
Patch/banner-improvements

dlipeles:
Several updates to the banner including

* Code refactor
* Improved image handling
* `401/500` error fixes
* Improved UX
* Visual updates


app/assets/stylesheets/pages/banner-intro.scss
```diff
@@ -7,13 +7,16 @@
...
   font-size: 17px;
 }
 
 .banner-intro-tagline {
   padding-right: 30px;
+  padding-top: 30px;
   padding-left: 30px;
-  margin-bottom: 60px;
+  margin-bottom: 50px;
+  font-size: 1.23rem;
```
corymcdonald: It seems weird we're mixing `rem` and `px`. I'd honestly prefer rem for the entire project but this can be a discussion/task that we can do later

corymcdonald:
Should we be using `tsx` as opposed to `jsx`?

app/assets/stylesheets/pages/banner-intro.scss
dlipeles: Agreed, these are under the umbrella of low-priority refactors I'd still like to do

nvonpentz:
@dlipeles would you please rebase with staging?  Travis is failing since the bundler audit test fails which I think was resolved with this https://github.com/brave-intl/publishers/pull/1298

PR #1331, opened 2018-11-09
Show referral stats. Closes #1326

### LINKED ISSUE #1326: Referral Total Stats
Include referral totals from eyeshade's `v1/accounts/earnings/referrals/total` as part of the admin dashboard

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/admin/stats/contributions_controller.rb
+++ b/app/controllers/admin/stats/contributions_controller.rb
@@ -1,5 +1,11 @@
 class Admin::Stats::ContributionsController < AdminController
   def index
     @result = Rails.cache.fetch(Cache::EyeshadeStatsJob::EYESHADE_CONTRIBUTION_TOTALS) || []
+    @publishers = Publisher.where(id: @result.map { |entry| entry['account_id'].split(":")[1] }).load
+    @youtube_channel_details = YoutubeChannelDetails
+      .where(youtube_channel_id:
+             @result.select { |entry| entry['channel'].starts_with?(YoutubeChannelDetails::YOUTUBE_PREFIX) }
+                    .map{ |entry| entry['channel'].remove(YoutubeChannelDetails::YOUTUBE_PREFIX) }
+      ).load
   end
 end
--- a/app/services/eyeshade/referral_totals.rb
+++ b/app/services/eyeshade/referral_totals.rb
@@ -1,0 +1,27 @@
+class Eyeshade::ReferralTotals < Eyeshade::BaseApiClient
+  def perform
+    if Rails.application.secrets[:api_eyeshade_offline]
+      result = perform_offline
+    else
+      begin
+        response = connection.get do |request|
+          request.headers["Authorization"] = api_authorization_header
+          request.url("/v1/accounts/earnings/referrals/total")
+        end
+        result = JSON.parse(response.body)
+      end
+    end
+    result
+  end
+
+  def perform_offline
+    [
+      {
+        "channel"=>"youtube#channel:UC0jeQ0Y8y6n-FWZNDIRO2Ug",
+        "earnings"=>"40.000000000000000000",
+        "account_id"=>"publishers#uuid:a064ed75-b185-4841-aefb-7a0ed61731b2"
+      }
+    ]
+  end
+end
+
--- a/app/services/eyeshade/referral_totals.rb
+++ b/app/services/eyeshade/referral_totals.rb
@@ -1,17 +1,13 @@
 class Eyeshade::ReferralTotals < Eyeshade::BaseApiClient
   def perform
-    if Rails.application.secrets[:api_eyeshade_offline]
-      result = perform_offline
-    else
-      begin
-        response = connection.get do |request|
-          request.headers["Authorization"] = api_authorization_header
-          request.url("/v1/accounts/earnings/referrals/total")
-        end
-        result = JSON.parse(response.body)
-      end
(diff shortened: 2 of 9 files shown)


app/services/eyeshade/referral_totals.rb
```diff
@@ -0,0 +1,27 @@
+class Eyeshade::ReferralTotals < Eyeshade::BaseApiClient
+  def perform
+    if Rails.application.secrets[:api_eyeshade_offline]
```
nvonpentz: minor: could you do `return perform_offline if Rails.application.secrets[:api_eyeshade_offline]` for consistency?

app/services/eyeshade/referral_totals.rb
```diff
@@ -0,0 +1,27 @@
...
+      result = perform_offline
+    else
+      begin
```
nvonpentz: what is your reasoning for the `begin` block?

app/views/admin/stats/referrals/index.html.slim
```diff
@@ -0,0 +1,14 @@
+div.panel-body
+  div.adv-table
+    table.display.table.table-bordered.table-striped.dynamic-table id="dynamic-table"
+      tr
+        th Channel
+        th Earnings
+        th Publisher
+      tbody
+        - @result.each do |entry|
+          - publisher_id = entry['account_id'].split(":")[1]
```
nvonpentz: You can use this method to find the publisher given the id with the prefix in order to display name, a link to their admin page, as well as grab other information you might want to display https://github.com/brave-intl/publishers/blob/1e80e15b001374ab8e7b2697c68bebce182ca095/app/models/publisher.rb#L328-L330

yachtcaptain23: I'll add this.

app/services/eyeshade/referral_totals.rb
yachtcaptain23: Probably just copy pasted. I'll add a catch block

app/views/admin/stats/referrals/index.html.slim
```diff
@@ -0,0 +1,14 @@
+div.panel-body
+  div.adv-table
+    table.display.table.table-bordered.table-striped.dynamic-table id="dynamic-table"
+      tr
+        th Channel
+        th Earnings
+        th Publisher
+      tbody
+        - @result.each do |entry|
```
corymcdonald: Should we display a "No results found" message if there's nothing in the result array?

app/views/admin/stats/referrals/index.html.slim
nvonpentz: We can also use the *_channel_details.url method to display the url for any channel type.  https://github.com/brave-intl/publishers/blob/1e80e15b001374ab8e7b2697c68bebce182ca095/app/models/youtube_channel_details.rb#L32-L34

nvonpentz: you can probably avoid a n+1 query with eager loading in the controller.

app/controllers/admin/stats/contributions_controller.rb
```diff
@@ -1,5 +1,11 @@
 class Admin::Stats::ContributionsController < AdminController
   def index
     @result = Rails.cache.fetch(Cache::EyeshadeStatsJob::EYESHADE_CONTRIBUTION_TOTALS) || []
+    @publishers = Publisher.where(id: @result.map { |entry| entry['account_id'].split(":")[1] }).load
```
corymcdonald: i'm a sucker for user Rails-esque style
`entry['account_id'].split(":").second`

app/services/eyeshade/referral_totals.rb
```diff
@@ -0,0 +1,23 @@
+class Eyeshade::ReferralTotals < Eyeshade::BaseApiClient
+  def perform
+    return perform_offline if Rails.application.secrets[:api_eyeshade_offline]
+
+    response = connection.get do |request|
+      request.headers["Authorization"] = api_authorization_header
+      request.url("/v1/accounts/earnings/referrals/total")
+    end
+
+    JSON.parse(response.body)
```
corymcdonald: I see a lot of parsing on this data from the eyeshade API. I wonder if it would make more sense to create an ActiveModel in which we initialize it with data from the Eyeshade API and it handles are the parsing. Seems like it would make the code easier to read and more simple.

app/controllers/admin/stats/contributions_controller.rb
corymcdonald: eh, on second thought nvm. the `[1]` is pretty prevalent throughout this codebase

app/services/eyeshade/referral_totals.rb
yachtcaptain23: I don't think it's a bad idea, but it's just 3 attributes.

yachtcaptain23: Preferably we do this right moving forward by using protocol buffers, https://developers.google.com/protocol-buffers/docs/reference/ruby-generated, but protobuf will give slightly more overhead in development.

corymcdonald: Yeah, I don't think it's necessary just an idea that we could potentially refactor later

Issue #1337, opened 2018-11-12
Save all referral stats to the stats column in the promo_registrations table


nvonpentz:
## Motivation
Currently publisher promo stats are stored in the publishers table.  A request is made to the referral server with the publisher's owner identifier and the referral server returns aggregate information for each publisher.  The problem is that we also store and fetch referral code statistics by the referral code for unattached referral codes.  Ideally, we fetch and store statistics only by referral code, since we can use a single method for all three types of codes: unattached, channel-owned, and publisher-owned.  This also will allow publisher-partners to get in depth statistics for their channel owned referral codes, which was not possible before.

## Action items
* Consolidate the AdminPromoStatsFetcher and the PublisherPromoStatsFetcher into a single service.  This effectively means removing the PublisherPromoStatsFetcher and renaming the AdminPromoStatsFetcher to PromoRegistrationStatsFetcher.

* Consolidate the SyncPublisherPromoStatsJob and SyncAdminPromoStatsJob into a single job.  This effectively means removing the SyncPublisherPromoStatsJob, renaming the SyncAdminPromoStatsJob to SyncPromoRegistrationStatsJob, and allowing it to take an optional parameter to determine which types for promo registrations (channel, owner, unattached) to sync the stats for.

* Remove publisher_promo_stats_2018q1 from the publishers table

* Update PromosHelper#total_referral_downloads and PromosHelper#confirmed_referral_downloads to source the publisher stats not from the publishers table, but from the stats saved to the referral code

This issue should be solved in two phases in order to ensure a smooth transition in which there is no downtime for publisher promo stats.  We should only remove the existing publisher promo stats infrastructure after the stats for channel referral codes have been synced and stored to the promo registration table.  Once that information is synced, we can remove the legacy promo stats services/jobs and the publisher_promo_stats_2018q1 column, and update the PromosHelper to source stats from stats column in the promo_registrations table.

PR #1338, opened 2018-11-12
Save referral stats for channel owned codes to the promo registrations table

nvonpentz:
Implements the first stage of #1337.

This PR adds a background job to save all channel referral stats to the promo registrations table.  Previously stats were saved to the publisher.  This does not change from which table the referral stats are sourced to display to the user, they will still be sourced from the `promo_stats_2018q1` column in the publishers table.  But when we do switch to the stats column in the promo_registrations table, this will ensure that the data will already be there.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:
Resolves 

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/jobs/sync_channel_promo_registrations_stats_job.rb
+++ b/app/jobs/sync_channel_promo_registrations_stats_job.rb
@@ -1,0 +1,10 @@
+# Fetches and saves the referral stats for channel owned codes
+class SyncChannelPromoRegistrationsStatsJob < ApplicationJob
+  include PromosHelper
+  queue_as :scheduler
+
+  def perform(promo_id: active_promo_id)
+    promo_registrations = PromoRegistration.channel
+    success = PromoRegistrationsStatsFetcher.new(promo_registrations: promo_registrations).perform
+  end
+end
--- a/app/jobs/sync_channel_promo_registrations_stats_job.rb
+++ b/app/jobs/sync_channel_promo_registrations_stats_job.rb
@@ -3,8 +3,8 @@ class SyncChannelPromoRegistrationsStatsJob < ApplicationJob
   include PromosHelper
   queue_as :scheduler
 
-  def perform(promo_id: active_promo_id)
+  def perform
     promo_registrations = PromoRegistration.channel
-    success = PromoRegistrationsStatsFetcher.new(promo_registrations: promo_registrations).perform
+    PromoRegistrationsStatsFetcher.new(promo_registrations: promo_registrations).perform
   end
 end
--- a/app/jobs/sync_channel_promo_registrations_stats_job.rb
+++ b/app/jobs/sync_channel_promo_registrations_stats_job.rb
@@ -4,7 +4,7 @@ class SyncChannelPromoRegistrationsStatsJob < ApplicationJob
   queue_as :scheduler
 
   def perform
-    promo_registrations = PromoRegistration.channel
+    promo_registrations = PromoRegistration.channels_only
     PromoRegistrationsStatsFetcher.new(promo_registrations: promo_registrations).perform
   end
 end
--- a/app/jobs/sync_unattached_promo_registrations_stats_job.rb
+++ b/app/jobs/sync_unattached_promo_registrations_stats_job.rb
@@ -1,11 +1,11 @@
-# Fetches and saves the referral stats for admins
-class SyncAdminPromoStatsJob < ApplicationJob
+# Fetches and saves the referral stats for unattached codes
+class SyncUnattachedPromoRegistrationsStatsJob < ApplicationJob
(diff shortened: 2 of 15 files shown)


app/models/promo_registration.rb
```diff
@@ -30,6 +30,7 @@ class PromoRegistration < ApplicationRecord
   validates :referral_code, presence: true, uniqueness: { scope: :promo_id }
 
   scope :unattached, -> { where(kind: UNATTACHED) }
+  scope :channel, -> { where(kind: CHANNEL) }
```
nvonpentz: I'm open to a different name to this scope.  Maybe `channel_owned` might be more intuitive

app/jobs/sync_channel_promo_registrations_stats_job.rb
```diff
@@ -0,0 +1,10 @@
+# Fetches and saves the referral stats for channel owned codes
+class SyncChannelPromoRegistrationsStatsJob < ApplicationJob
+  include PromosHelper
+  queue_as :scheduler
+
+  def perform(promo_id: active_promo_id)
+    promo_registrations = PromoRegistration.channel
```
yachtcaptain23: Did you want to lookup `active_promo_id`?

app/models/promo_registration.rb
yachtcaptain23: How about `channel_promo`?

app/jobs/sync_channel_promo_registrations_stats_job.rb
nvonpentz: Hm, the `active_promo_id` was there to be able to distinguish between different promotions.  We originally we're going to run multiple referral promos, but in practice we're only running one long one, so the `active_promo_id` should be the same for all codes.

I think @evq may be using `active_promo_id` to distinguish between unattached and channel owned referral codes, but I don't think that matters here.

app/models/promo_registration.rb
nvonpentz: I think `PromoRegistration.channel_registration` would probably be more correct if we want to go that direction

yachtcaptain23: Or `channels_only`? `channel_registration` seems verbose.

app/jobs/sync_channel_promo_registrations_stats_job.rb
nvonpentz: I did not and it is removed now, thanks.

app/models/promo_registration.rb
nvonpentz: I like `channels_only`, will change

nvonpentz: should I change `unattached` to `unattached_only` for consistency?

app/jobs/sync_unattached_promo_registrations_stats_job.rb
```diff
@@ -0,0 +1,14 @@
...
+class SyncUnattachedPromoRegistrationsStatsJob < ApplicationJob
+  include PromosHelper
+  queue_as :low
+
+  def perform(promo_id: active_promo_id)
+    promo_registrations = PromoRegistration.unattached_only
+    success = PromoRegistrationsStatsFetcher.new(promo_registrations: promo_registrations).perform
+
+    if success
+      Rails.cache.write("unattached_promo_registration_stats_last_synced_at", Time.now)
```
corymcdonald: Curious about where we fetch this, and if we're using this

nvonpentz: We fetch this here https://github.com/brave-intl/publishers/blob/1e80e15b001374ab8e7b2697c68bebce182ca095/app/views/admin/unattached_promo_registrations/index.html.slim#L27

That view is only accessible by admins.  I do have plans to display when the stats were last refreshed for publishers in the second part to resolve #1337

app/jobs/sync_unattached_promo_registrations_stats_job.rb
```diff
@@ -0,0 +1,14 @@
+# Fetches and saves the referral stats for unattached codes
+class SyncUnattachedPromoRegistrationsStatsJob < ApplicationJob
+  include PromosHelper
+  queue_as :low
+
+  def perform(promo_id: active_promo_id)
```
yachtcaptain23: Unused parameters?

yachtcaptain23:
Nit picking: We should create the tasks under `app/jobs/sync/promo_server`.

app/jobs/sync_unattached_promo_registrations_stats_job.rb
nvonpentz: removed

PR #1339, opened 2018-11-12
Partner Role

### LINKED ISSUE #1327: Create Publisher Partner role
Action Items
- [ ] Separate model, inherit from Publishers?
- [ ] If necessary, update abilities to not need uphold.

Useful reference documents:
https://github.com/CanCanCommunity/cancancan

dlipeles:
Basis for the forthcoming `Partner` role as mentioned in @mandar-brave's v1.1 

Using the console `Publisher` can be set to `Partner` like so:

```ruby
record.role = "partner"
record.save
```

`partner` will not have access to admin privileges.

Closes #1327

### DIFF
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -6,8 +6,9 @@ class Publisher < ApplicationRecord
   UPHOLD_ACCESS_PARAMS_TIMEOUT = 2.hours
   PROMO_STATS_UPDATE_DELAY = 10.minutes
   ADMIN = "admin"
+  PARTNER = "partner"
   PUBLISHER = "publisher"
-  ROLES = [ADMIN, PUBLISHER]
+  ROLES = [ADMIN, PARTNER, PUBLISHER]
   JAVASCRIPT_DETECTED_RELEASE_TIME = "2018-06-19 22:51:51".freeze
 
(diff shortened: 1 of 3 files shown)


app/models/publisher.rb
```diff
@@ -71,6 +72,7 @@ class Publisher < ApplicationRecord
 
   scope :email_verified, -> { where.not(email: nil) }
   scope :not_admin, -> { where.not(role: ADMIN) }
+  scope :not_partner, -> { where.not(role: PARTNER) }
```
nvonpentz: how do you see this scope being used? thinking that we should add `publisher` `partner` and `admin` scopes and tests for the scopes.

dlipeles: Thanks for the review @nvonpentz , you're cool with calling it `partner` right? 

For uphold, one idea might be to prevent them from reaching the dashboard, but I'm not sure if this completely prevents them, as they might be able to connect through APIs. Lmk if you have any other ideas

I added that scope because it seemed like a good idea we might want to use later, but good thinking I'll add some tests for these scopes.

yachtcaptain23: We could chain `Publisher.not_admin.not_partner` or use `Publisher.publisher`

PR #1345, opened 2018-11-13
Catalog HTTPS configuration errors

### LINKED ISSUE #253: Catalog HTTPS configuration errors, generate appropriate guidance for users in trusted file flow.
There are a reasonable number of HTTPS configuration errors which impeded trusted file flows.

We need a catalog of common errors and then guidance copy created to help users overcome their configuration issue.

Edit (Albert Wang): Examples are found on https://badssl.com/dashboard/

corymcdonald:
## Overview
When the user [adds a new channel website](https://127.0.0.1:3000/site_channels/new) that contains an invalid HTTPS certificate it will provide the user with additional details.

screen shot 2018-11-13 at 4 16 21 pm

Here are the results of what `fetch` returns as invalid 

| domain | working as intended |
| ------- | ------- |
|expired.badssl.com | ✅ |
|wrong.host.badssl.com | ✅ |
|self-signed.badssl.com | ✅ |
|untrusted-root.badssl.com | ✅ |
| sha1-intermediate.badssl.com | ❌ |
| rc4.badssl.com | ❌ |
|rc4-md5.badssl.com | ✅ |
|dh480.badssl.com | ✅ |
|dh512.badssl.com | ✅ |
| dh1024.badssl.com | ❌ |
|superfish.badssl.com | ✅ |
|edellroot.badssl.com | ✅ |
|dsdtestprovider.badssl.com | ✅ |
|preact-cli.badssl.com | ✅ |
|webpack-dev-server.badssl.com | ✅ |
|null.badssl.com | ✅ |
| revoked.badssl.com | ❌ |
| pinning-test.badssl.com/ | ❌ |
| invalid-expected-sct.badssl.com | ❌ |

I added `badssl.com` to the list of exceptions for testing purposes.

## Caveats 
The following are detected as valid in publishers even though they are marked as invalid on badssl.com

rc4.badssl.com - this is actually deprecated in the latest version of openssl, I believe uplifting to a newer version of ruby will address this kind of certificate error

So I think this will eventually get covered when we upgrade our ruby version, until then we’ll have to consume 2.3.8’s openssl library instead of the latest version

sha1-intermediate.badssl.com 🚫
revoked.badssl.com 🚫 (Revoked is an interesting use case because we’d have to make an HTTP request to the root server to verify) 

Chrome only security exception, we’re probably fine to accept it.
invalid-expected-sct.badssl.com 🚫 

Chrome detects this as being invalid but Firefox does not
dh1024.badssl.com 🚫

Closes #253


db/schema.rb
```diff
@@ -13,7 +13,6 @@
 ActiveRecord::Schema.define(version: 2018_10_16_134245) do
 
   # These are extensions that must be enabled in order to support this database
-  enable_extension "pg_stat_statements"
```
corymcdonald: This was changed when I ran a `db:migrate`. I'm curious if this is required because we don't have a migration which enables it.

In addition it's enabled by default on all Heroku instances - https://devcenter.heroku.com/changelog-items/425

test/features/site_channel_verification_test.rb
```diff
@@ -31,6 +31,32 @@ class SiteChannelVerificationTest < Capybara::Rails::TestCase
     refute_content "Cancel"
   end
 
+  test "When bad ssl happens" do
+    Rails.application.secrets[:host_inspector_offline] = false
```
corymcdonald: Not sure if it's okay or not to make an actual HTTP request, but since this is technically "integration" I viewed it as fine. 

Thoughts?

db/schema.rb
yachtcaptain23: Which postgres version are you running?

app/services/site_channel_host_inspector.rb
```diff
@@ -30,13 +30,11 @@ def inspect_uri(uri)
...
                  elsif response.body.include?("/wp-content/")
                    "wordpress"
-                 else
-                   nil
                  end
     end
 
     { response: response, web_host: web_host }
-  rescue Publishers::Fetch::RedirectError, Publishers::Fetch::ConnectionFailedError => e
+  rescue => e
```
yachtcaptain23: Wouldn't line 24 throw an exception and prevent us from knowing some SSL errors?

yachtcaptain23:
IMO: We should clear the HTTPS error if the site verification works.

test/features/site_channel_verification_test.rb
nvonpentz: Our test suite thus far does not make any external http requests, I'd prefer if we could stub these.  With the external services within Brave (e.g. eyeshade, promo) we've just been using the `stub_request` method, but have also set up vcr cassettes to stub sendgrid requests, and stubbed dns responses for site verification in test/stubs/dnsruby/.

test/features/site_channel_verification_test.rb
```diff
@@ -31,6 +31,32 @@ class SiteChannelVerificationTest < Capybara::Rails::TestCase
...
+    assert_content "Cancel"
+
+    click_link "I'll use a trusted file"
+    assert_current_path verification_public_file_site_channel_path(channel)
+    assert_content "Choose Different Verification Method"
+    refute_content "Cancel"
+
+    assert_content "The following error was encountered: SSL_connect returned=1 errno=0 state=error: certificate verify failed"
+
+    Rails.application.secrets[:host_inspector_offline] = true
```
nvonpentz: (minor) This won't apply if we stub the responses, but setting the `Rails.application.secrets[:host_inspector_offline] = true` assumes that it was previously set to true, which may not be the case (now or in the future).  We've been using this convention to change envars for specific tests:

https://github.com/brave-intl/publishers/blob/d6f02765c389d99e62026edf9865c94b0aad7680/test/controllers/admin/payout_reports_controller_test.rb#L8-L14

Then you just declare what value you want the envar to be at the top of the test.

db/schema.rb
corymcdonald: postgres (PostgreSQL) 10.5 

When you run your migrations does it update this value? Perhaps my version isn't correct

app/services/site_channel_host_inspector.rb
corymcdonald: Yep, the `fetch` will raise an exception but since this is caught the exception is returned in the response it's able to be set in the `response_result`

test/features/site_channel_verification_test.rb
corymcdonald: Ah! Nice catch. I wouldn't have thought about that since I'm used to RSpec isolating all my tests. I'm going to switch over to mocking the request so we don't have to worry about this.

corymcdonald: Actually, I did need to set this so the request is actually mocked - https://github.com/brave-intl/publishers/pull/1345/commits/19075a783167f44b10d0b0d74477b2098a82f08d

test/features/site_channel_verification_test.rb
corymcdonald: Mocked it out - https://github.com/brave-intl/publishers/pull/1345/commits/19075a783167f44b10d0b0d74477b2098a82f08d :)

corymcdonald:
@yachtcaptain23 

Looks like it's doing that 🎉 

This is what happens when there is the original bad domain. 
```
irb(main):019:0> pp p.site_channel_details
[#<SiteChannelDetails:0x00007fb6b28bb648
  id: "1c0496da-eb1d-4c36-840b-3e9a8b44a403",
  brave_publisher_id: "expired.badssl.com",
  brave_publisher_id_unnormalized: nil,
  brave_publisher_id_error_code: nil,
  verification_token: nil,
  verification_method: nil,
  supports_https: false,
  host_connection_verified: false,
  detected_web_host: nil,
  created_at: Wed, 14 Nov 2018 13:52:39 PST -08:00,
  updated_at: Wed, 14 Nov 2018 13:52:39 PST -08:00,
  stats: "{}",
  https_error:
   "SSL_connect returned=1 errno=0 state=error: certificate verify failed">]
```
Then when the user clicks "Try Again" and it's valid

```
irb(main):026:0> pp p.site_channel_details
[#<SiteChannelDetails:0x00007fb6a6d9fbe8
  id: "1c0496da-eb1d-4c36-840b-3e9a8b44a403",
  brave_publisher_id: "expired.badssl.com",
  brave_publisher_id_unnormalized: nil,
  brave_publisher_id_error_code: nil,
  verification_token:
   "8cdecb48fc65dd579a15a6805cb376cf7119e67fba340e994ae1c948ecf19a07",
  verification_method: nil,
  supports_https: true,
  host_connection_verified: true,
  detected_web_host: nil,
  created_at: Wed, 14 Nov 2018 13:52:39 PST -08:00,
  updated_at: Wed, 14 Nov 2018 13:54:00 PST -08:00,
  stats: "{}",
  https_error: nil>]
```

db/schema.rb
corymcdonald: Reverted that change 👍 
https://github.com/brave-intl/publishers/pull/1345/commits/cc84aeefd4661eced8eb89faa5eebe048eb71bae

PR #1346, opened 2018-11-13
Additional scopes and tests

dlipeles:


yachtcaptain23:
Also part of https://github.com/brave-intl/publishers/issues/1327

PR #1349, opened 2018-11-15
[WIP] Feature/user groups

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/models/membership.rb
```diff
@@ -0,0 +1,10 @@
+class Membership < ActiveRecord::Base
+  belongs_to :organization, class_name: "Organization"
+
+  # Publisher model will be responsible for casting to the right sub-role
+  # E.g. casting from Publisher to Partner
```
corymcdonald: I'm a little uncertain about the architecture of how we're going to cast from a Publisher to a Partner, do you have any ideas on the matter?

app/models/organization.rb
```diff
@@ -0,0 +1,10 @@
+class Organization < Publisher
+  default_scope { where(role: Publisher::ORGANIZATION) }
+
+  validate :no_email
+  has_many :memberships
+
+  def no_email
```
corymcdonald: Do we intend to find these by name? since there are no emails?

Curious what the workflow is for creating these organizations

yachtcaptain23:
They can be added on an existing partner. Also can find by channel if
needed.

app/models/membership.rb
yachtcaptain23: https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-becomes

yachtcaptain23:
In my mind, the permissions within the organization can be specified in the `publishers` table or `memberships` table.

app/controllers/admin/partners_controller.rb
```diff
@@ -52,8 +58,16 @@ def partner
       existing_publisher || Partner.new(email: email_params)
     end
 
+    def organization
+      Organization.find_by(name: params[:organization_name]) || Organization.new(name: params[:organization_name])
```
corymcdonald: Let's do this instead
```suggestion
      Organization.find_or_initialize_by(name: params[:organization_name])
```

app/controllers/admin/partners_controller.rb
```diff
@@ -24,15 +24,21 @@ def new
     def create
       # Find any existing publishers so we don't create duplicate entries
       @partner = partner
+      @organization = organization
 
       if @partner.persisted? && (@partner.partner? || @partner.admin?)
         flash.now[:alert] = "Email is already a partner"
         render :new
+      elsif @organization.persisted?
+        flash.now[:alert] = "The organization '#{params[:organization_name]}' already exists. Please have a partner of the organization add the user you want or ask Engineering team for assistance"
```
corymcdonald: I think i'd prefer getting the name from the object we've created
```suggestion
        flash.now[:alert] = "The organization '#{@organization.name}' already exists. Please have a partner of the organization add the user you want or ask Engineering team for assistance"
```

app/models/publisher.rb
```diff
@@ -5,9 +5,9 @@ class Publisher < ApplicationRecord
   UPHOLD_CODE_TIMEOUT = 5.minutes
   UPHOLD_ACCESS_PARAMS_TIMEOUT = 2.hours
   PROMO_STATS_UPDATE_DELAY = 10.minutes
-  ADMIN = "admin"
-  PARTNER = "partner"
-  PUBLISHER = "publisher"
+  ADMIN = "admin".freeze
```
corymcdonald: could be easier to add this to the top of the file

```ruby
# frozen_string_literal: true
```

db/schema.rb
```diff
@@ -48,6 +48,8 @@
     t.datetime "updated_at", null: false
     t.string "verification_status"
     t.string "verification_details"
+    t.boolean "shown_verification_failed_modal", default: false
```
corymcdonald: Are these legacy? Can you recreating the db so we have the most up to date schema?

corymcdonald:
We have a button on the User page that says "Make Partner" should we enforce that they are associated to an organization there as well?

db/schema.rb
yachtcaptain23: I must have added wrong hunk when rebasing. I'll fix.

yachtcaptain23:
> We have a button on the User page that says "Make Partner" should we enforce that they are associated to an organization there as well?

Yes.

app/models/publisher.rb
yachtcaptain23: I'm hesitant to take this on unless I'm certain that all strings within the model are immutable.

PR #1350, opened 2018-11-15
Allow publishers to own referral codes, do not display publisher#promo_stats_2018q1

nvonpentz:
Part two of #1337.  Blocked until https://github.com/brave-intl/publishers/pull/1338 is in production for a couple days to propagate stats to PromoRegistration#stats.

This makes promo registrations polymorphic.  So a channel promo_registration now belongs to one channel and one publisher. A publisher has many promo registrations.

Now we can do things like `publisher.promo_registrations` to get all the promo registrations associated with the publisher. This allows us to easily sum the stats for all promo registrations belonging to a publisher and display them to the publisher.

Part 3 of #1337 will include removing `promo_stats_2018q1q1` from the publishers table and other clean up.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/models/promo_registration.rb
+++ b/app/models/promo_registration.rb
@@ -4,17 +4,17 @@ class PromoRegistration < ApplicationRecord
   # A promo registration can belong to a channel,
   # a publisher, or be unattached.  Unattached codes
   # are created by admins.
-
-  # Constants
   CHANNEL = "channel".freeze
   OWNER = "owner".freeze
   UNATTACHED ="unattached".freeze
   KINDS = [CHANNEL, OWNER, UNATTACHED].freeze
 
+  # Event types
   RETRIEVALS = "retrievals" # Aliased as 'Downloads'
   FIRST_RUNS = "first_runs" # Aliased as 'Installs'
   FINALIZED = "finalized" # Aliased as 'Confirmed'
 
+  # Reporting intervals
   DAILY = "daily"
   WEEKLY = "weekly"
   MONTHLY = "monthly"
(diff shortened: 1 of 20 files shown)


app/models/promo_registration.rb
```diff
@@ -4,26 +4,29 @@ class PromoRegistration < ApplicationRecord
...
   MONTHLY = "monthly"
   RUNNING_TOTAL = "running_total"
   
   belongs_to :channel, validate: true, autosave: true
   belongs_to :promo_campaign
+  belongs_to :publisher, dependent: :destroy
 
-  validates :channel_id, presence: true, if: -> { kind == CHANNEL}
+  validates :channel_id, presence: true, if: -> { kind == CHANNEL }
+  validates :publisher_id, presence: true, if: -> { kind != UNATTACHED }
```
corymcdonald: Might read a little better if we used `unless`
```suggestion
  validates :publisher_id, presence: true, unless: -> { kind == UNATTACHED }
```

app/services/promo_registrar.rb
```diff
@@ -15,7 +15,11 @@ def perform
         referral_code = result[:referral_code]
         should_update_promo_server = result[:should_update_promo_server]
         if referral_code.present?
-          promo_registration = PromoRegistration.new(channel_id: channel.id, promo_id: @promo_id, kind: "channel", referral_code: referral_code)
+          promo_registration = PromoRegistration.new(channel_id: channel.id,
+                                                     promo_id: @promo_id,
+                                                     kind: "channel",
```
corymcdonald: Can we use the constant here?
```suggestion
                                                     kind: PromoRegistration::CHANNEL,
```

nvonpentz: gonna try this commit suggestion feature

nvonpentz:
So we update the old stats here when a publisher logs in:

https://github.com/brave-intl/publishers/blob/2f79dd5651d8f989c420b5c9e74dad0d40bee527/app/controllers/publishers_controller.rb#L367-L369

But since this launches a job, the dashboard won't update until they refresh, so I'm not really sure how effective this is.  (This was previously synchronous and thus guaranteed the stats would be refreshed before the dashboard loads, but we made it async for performance reasons).

I'll need to either remove these lines entirely or replace them with logic that does the same thing with the new system.  What should we do?  I'm tending toward just removing these all together and relying on the Sync::ChannelPromoRegistrationsStatsJob to update the stats in the background.  We could up the frequency of Sync::ChannelPromoRegistrationsStatsJob in sidekiq since it's really only one request per job (where as the previous system was 1 request per publisher within the job).

Issue #1328, opened 2018-11-08
Assign a Publisher to Publisher Partner role


yachtcaptain23:
- [ ] Allow assigning of a publisher to a `publisher_partner` in the admin dashboard.

yachtcaptain23:
Closed due to merging of #1354

PR #1354, opened 2018-11-16
Allow admins to create partners

corymcdonald:
Addresses issue #1328 

🖼 Here's some screenshots 🖼 

image

image

image

image

image


yachtcaptain23:
Would you create an index page to list all partners? And move the button to create a partner to there.

nvonpentz:
The admin will need to be able to set the permissions for a partner and update those permissions later on.

We also need to keep track of which admin created which partner and only display the partners that the admin created to the admin.

corymcdonald:
@nvonpentz @yachtcaptain23 This is ready for re-review, I've addressed your comments 😄

app/controllers/admin/publishers_controller.rb
```diff
@@ -1,19 +1,19 @@
...
+  include Search
 
   def index
     @publishers = Publisher
 
     if params[:q].present?
       # Returns an ActiveRecord::Relation of publishers for pagination
-      @publishers = Publisher.where("publishers.id IN (#{sql(params[:q])})").distinct
+      search_query = "%#{remove_prefix_if_necessary(params[:q])}%"
+      @publishers = Publisher.where(search_sql, *[search_query] * 7).distinct
```
yachtcaptain23: What does `*[search_query]* 7` mean?

app/controllers/admin/partners_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      @partner = Partner.new
+    end
+
+    def create
+      # Find any existing publishers so we don't create duplicate entries
+      @partner = partner
+
+      if @partner.persisted? && (@partner.partner? || @partner.admin?)
+        flash.now[:alert] = "Email is already a partner"
+        render "new"
```
yachtcaptain23: Would you use the symbol?

app/controllers/admin/publishers_controller.rb
corymcdonald: Before this there was a SQL injection on the admin controller. I removed that by refactoring the search but didn't realize we could use [placeholder conditions](https://guides.rubyonrails.org/active_record_querying.html#placeholder-conditions) as opposed to `?`. I refactored it so that now it's more clear what's going on 
https://github.com/brave-intl/publishers/pull/1354/commits/9f2060a32ca89c8ea4c8dfaa5c7a8a9e7a39c60b

app/controllers/admin/partners_controller.rb
corymcdonald: Yup sure thing - https://github.com/brave-intl/publishers/pull/1354/commits/9f2060a32ca89c8ea4c8dfaa5c7a8a9e7a39c60b

PR #1355, opened 2018-11-16
Don't recreate the same channel due to https errors. Closes #1348

### LINKED ISSUE #1348: New channels disappear during "Complete Verification Later" workflow
When the user goes through the workflow and adds a new channel they can effectively "lose" their channel by clicking on the "Complete Verification later" button. I initially discovered this when testing the SSL workflows and although the links were grayed out they were still clickable and therefore lost

Reproduce:
1. Sign in as a publisher
2. Click "+ Add Channel"
3. Choose "Website"
4. Enter any domain you wish
5. Select either "I'll use a trusted file" or "I'll edit my DNS record"
6. Click "Complete Verification Later"
7. Redirected to home page with the created channel not in the channel list.

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


nvonpentz:
Not sure if this is completed yet, but I'm still able to reproduce #1348 on this branch

corymcdonald:
I was investigating what to do in that case because really what happens is that the domain normalizer strips everything and goes to `username.github.io` which calls the `fetch` call to return 404. 

So we'd need to write some code to handle that in this method 
https://github.com/brave-intl/publishers/blob/staging/app/services/site_channel_host_inspector.rb#L46

yachtcaptain23:
Spoke with Nick

Action items:
(1) Get rid of the `abandoned` scope on `SiteChannelDetails`
(2) Update the code in `app/jobs/clean_abandoned_site_channels_job.rb`
(3) Re-insert back original controller logic.

app/jobs/clean_abandoned_site_channels_job.rb
```diff
@@ -6,15 +6,13 @@ def perform
...
 
-    channel_details = SiteChannelDetails.abandoned
+    channels = Channel.not_visible_site_channels
     n = 0
-    channel_details.each do |details|
-      raise unless details.verification_token.nil?
-
-      details.channel.destroy
+    channels.joins(:site_channel_details).each do |channel|
+      raise unless channel.details.verification_method.nil?
```
corymcdonald: can we change this to 
`raise if channel.details.verification_method.present?` 
i think that's slightly more readable

corymcdonald:
Hey Albert, I don't think this actually resolves the issue, here's a gif to reproduce it 
nov-21-2018 11-07-05

yachtcaptain23:
@corymcdonald I agree that this solution doesn't explain the whole story. The issue's wording should be restructured, as the issue you're seeing might not be an issue, and I thought it had tied into another issue in Sentry. I think we should continue to allow users to make channels that haven't fulfilled the flow (it's a tradeoff between refactoring a bunch of logic and inheriting more risk or maintain a confusing mental model of visible and non-visible/abandoned channels 😞 )

I originally intended to solve https://sentry.io/brave-software/publishers-production-t2/issues/761612367/?query=is:unresolved and #1348 in this PR. However, they seem to be separate problems: the sentry error was 1 simple bad channel in which I fixed. This PR solves some weird state called `abandoned` when it should have just been `non-visible`

To elaborate further on the Sentry error: We have this constraint that prevents a publisher from creating duplicates (and somehow, someone got around it. I don't think it's a race condition though):

```
  def site_channel_details_brave_publisher_id_unique_for_publisher
    duplicate_unverified_channels = publisher.channels.visible_site_channels
                                                     .where(site_channel_details: { brave_publisher_id: self.details.brave_publisher_id })
                                                     .where.not(id: id)

    if duplicate_unverified_channels.any?
      errors.add(:brave_publisher_id, "must be unique")
    end
  end
```

Anyways, an elegant solution to solve #1348 for a dropped funnel, is to allow the user to continue from where they left off. However, there's several channels that are abandoned and not meant to be visible. Overall the state transitions for adding a channel are confusing, but they haven't caused any major issues, yet (except for us developers).

nvonpentz:
should be resolved now, the verification method wasn't being set which is needed to make a channel visible

PR #1357, opened 2018-11-18
Multi-Channel mode for tipping banners

dlipeles:


app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
 
   MAX_IMAGE_SIZE = 10_000_000
 
-  def new
-    @site_banner = current_publisher.site_banner || SiteBanner.new
+  #Fetch Banner by channel_id, if it does not exist create a template banner.
+  def fetch
+    site_banner = current_publisher.site_banners.where(channel_id: params[:channel_id]).first
```
corymcdonald: Alternatively we could 
```suggestion
    site_banner = current_publisher.site_banners.find_by(channel_id: params[:channel_id])
```

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
...
+    if site_banner.present?
+      data = site_banner.read_only_react_property
+      data[:backgroundImage] = data[:backgroundUrl]
+      data[:logoImage] = data[:logoUrl]
+      render(json: data.to_json)
+    else
+      site_banner = SiteBanner.new
```
corymcdonald: This is a pretty long method, can we extract this into a new method?

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
...
+      site_banner.update(
+        publisher_id: current_publisher.id,
+        channel_id: params[:channel_id],
+        title: 'Brave Rewards',
```
corymcdonald: Can we internationalize `Brave Rewards` and the `description` below?

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
...
+        description:
+'Thanks for stopping by. We joined Brave\'s vision of protecting your privacy because we believe that fans like you would support us in our effort to keep the web a clean and safe place to be.
+
+Your tip is much appreciated and it encourages us to continue to improve our content.',
+        donation_amounts: [1, 5, 10],
+        default_donation: 5,
+        social_links: {'youtube': '', 'twitter': '', 'twitch': ''}
```
corymcdonald: Although I prefer single quotes our linter is configured to do double quotes (even if it doesn't run on this file yet) we should probably be consistent.

@yachtcaptain23 what are your thoughts?

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
 
   MAX_IMAGE_SIZE = 10_000_000
 
-  def new
-    @site_banner = current_publisher.site_banner || SiteBanner.new
+  #Fetch Banner by channel_id, if it does not exist create a template banner.
```
corymcdonald: super nit: let's put a space after the comment

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
...
+      render(json: data.to_json)
+    end
   end
 
-  def create
-    site_banner = current_publisher.site_banner || SiteBanner.new
+  def channels
+
+    # Handle edge case where user hasn't created any channels yet
+    if current_publisher.channels.length == 0
```
corymcdonald: nit: let's do `.zero?` instead of comparing with `== 0` :innocent:

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
...
+      data = {}
+      data[:channels] = [{'id' => '00000000-0000-0000-0000-000000000000', 'name' => ' ', 'type' => 'website'}]
```
corymcdonald: I'm a little bit confused about why we're creating a dummy channel here to return to the user? Shouldn't we just return an empty array and the UI take care of it?

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
...
+      data[:channel_mode] = false
+      data[:channel_index] = 0
+      render(json: data.to_json) and return
```
corymcdonald: nit:  let's not do and return, just `return render`

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
...
+    end
+
+    if current_publisher.site_banners.where(channel_id: '00000000-0000-0000-0000-000000000000').exists?
+      current_publisher.site_banners.where(channel_id: '00000000-0000-0000-0000-000000000000').first.delete
```
corymcdonald: Maybe we don't need this if we're not going to be creating a dummy channel?

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
...
+    end
+
+    channels = []
+
+    current_publisher.channels.each do |c|
+
+    case c.details_type
+
+      when "SiteChannelDetails"
```
corymcdonald: So we have the same code in the `channels_controller` is there a reason why this auth_name is different? Should we extract this into a helper function so that the code doesn't get out of sync and there's a bug around different channel_names being displayed differently across our application?

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
...
+        channel_type = "youtube"
+        channel_name = channel_details.auth_name
+      when "TwitchChannelDetails"
+        channel_details = TwitchChannelDetails.find_by_id(c.details_id)
+        channel_type = "twitch"
+        channel_name = channel_details.name
+
+    end
+
+    channel = {
```
corymcdonald: It might make more sense if the result of what this loop is doing was a method in the channels model. If that was extracted that might make this method a lot more readable

Then this entire loop could be reduced down to
```ruby
channels = current_publisher.channels.map { |c| c.site_banner_properties }
```
(up for suggestions on a better method name too) 😄

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
...
+              'id' => c.id,
+              'name' => channel_name,
+              'type' => channel_type
+              }
+
+    channels.push(channel)
+    end
+
+    channel_mode = true
```
corymcdonald: I was thinking this  block could potentially be more ruby-esque so we don't have define these temp variables

```ruby
    default_banner = current_publisher.site_banners.find_by(default: true)
    if default_banner.present?
      default_id = default_banner.channel_id
      channel_index = channels.index { |c| c["id"] == default_id }
    end

    data = {}
    data[:channels] = channels
    data[:channel_mode] = default_banner.empty?
    data[:channel_index] = channel_index || 0
    render json: data.to_json
```

app/javascript/packs/banner_editor.jsx
```diff
@@ -490,6 +512,35 @@ export default class BannerEditor extends React.Component {
       });
   }
 
+  incrementChannelIndex(){
+    let temp = this.state.channelIndex;
+    temp++;
+    this.setState({channelIndex: temp, loading: true},
```
corymcdonald: Not sure if we're linting javascript with eslint but I've seen this as lint before. Since `setState` is asynchronous we should refer to the previous state like so

```suggestion
    this.setState((prevState) => { channelIndex: (prevState.channelIndex += 1), loading: true },
```

app/javascript/packs/banner_editor.jsx
```diff
@@ -490,6 +512,35 @@ export default class BannerEditor extends React.Component {
...
+  toggleChannelMode(){
+    if(this.state.channelMode === true){
+      this.setState({channelMode: !this.state.channelMode, state: 'same'},
+      () => { this.save() }
+      )
+    }
+    else{
+      this.setState({channelMode: !this.state.channelMode},
+      () => { this.save() }
+      )
```
corymcdonald: Can we format this a little differently? The parenthesis at the end looks really strange

config/secrets.yml
```diff
@@ -61,8 +61,8 @@ default: &default
   terms_of_service_url: <%= ENV["TERMS_OF_SERVICE_URL"] %>
   google_client_id: <%= ENV["GOOGLE_CLIENT_ID"] %>
   google_client_secret: <%= ENV["GOOGLE_CLIENT_SECRET"] %>
-  twitch_client_id: <%= ENV["TWITCH_CLIENT_ID"] %>
-  twitch_client_secret: <%= ENV["TWITCH_CLIENT_SECRET"] %>
+  twitch_client_id: t7adodv1mqh9ujramlwes2ccnxx05z
+  twitch_client_secret: t6xub0swz5gzktlmy1kff9v9c7owt5
```
corymcdonald: Probably didn't mean to commit this?

config/routes.rb
```diff
@@ -38,9 +38,11 @@
       resources :totp_authentications, only: %i(create)
       resources :promo_registrations, only: %i(index create)
     end
-    resources :site_banners, only: %i(new create), controller: 'publishers/site_banners' do
+    resources :site_banners, controller: 'publishers/site_banners' do
       collection do
         get :fetch
+        get :channels
```
corymcdonald: So I think our controller method could be alternatively named `index` because that is by default included as a resource on `site_banner` currently. Then we don't have to define this `get :channels`

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -23,24 +113,27 @@ def create
...
+   site_banner.update(
+     publisher_id: current_publisher.id,
+     title: params[:title],
+     donation_amounts: donation_amounts,
+     default_donation: donation_amounts.second,
+     social_links: params[:social_links].present? ? JSON.parse(params[:social_links]) : {},
+     description: params[:description]
+   )
+   head :ok
+ end
```
nvonpentz: `site_banner = current_publisher.site_banners || SiteBanner.new`

What happens if this evaluates to `current_publisher.site_banners`?

app/controllers/publishers/site_banners_controller.rb
nvonpentz: best practice is to use `"` for ruby for this project

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,104 @@ class Publishers::SiteBannersController < ApplicationController
...
+              'type' => channel_type
+              }
+
+    channels.push(channel)
+    end
+
+    channel_mode = true
+    channel_index = 0
+
+    if(current_publisher.site_banners.where(default: true).exists?)
```
nvonpentz: can we make a `default_site_banner` method on the Publisher model?

app/controllers/publishers/site_banners_controller.rb
dlipeles: Done ✔️

app/javascript/packs/banner_editor.jsx
dlipeles: Done ✔️

app/controllers/publishers/site_banners_controller.rb
dlipeles: Done ✔️

app/controllers/publishers/site_banners_controller.rb
dlipeles: Done ✔️

app/controllers/publishers/site_banners_controller.rb
dlipeles: Done ✔️

app/controllers/publishers/site_banners_controller.rb
dlipeles: Done ✔️

app/controllers/publishers/site_banners_controller.rb
dlipeles: Done ✔️

app/controllers/publishers/site_banners_controller.rb
dlipeles: Done ✔️

db/migrate/20181118005740_add_default_to_site_banner.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultToSiteBanner < ActiveRecord::Migration[5.2]
+  def change
+    add_column :site_banners, :default, :boolean
```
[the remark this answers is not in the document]
yachtcaptain23: Don't add the uniqueness contraint here given the uniqueness constaint in database only allows 1 true one false. You can add a validation in the model

app/controllers/publishers/site_banners_controller.rb
yachtcaptain23: Fetch should never save a new site banner (i.e. if we write a GET, it shouldn't save anything)

config/secrets.yml
dlipeles: Done ✔️ (😅)

app/models/json_builders/channels_json_builder.rb
```diff
@@ -26,15 +26,15 @@ def initialize
...
+  [Channel.left_joins(publisher: :site_banners).verified.site_channels, Channel.left_joins(publisher: :site_banners).youtube_channels, Channel.left_joins(publisher: :site_banners).twitch_channels, Channel.left_joins(publisher: :site_banners).twitter_channels].each do |verified_channels|
       verified_channels.find_each do |verified_channel|
         if @excluded_channel_ids.include?(verified_channel.details.channel_identifier)
           excluded = true
           @excluded_verified_channel_ids.push(verified_channel.details.channel_identifier)
         else
           excluded = false
         end
-        channels.push([verified_channel.details.channel_identifier, true, excluded, verified_channel.publisher&.site_banner&.read_only_react_property || {}])
+        channels.push([verified_channel.details.channel_identifier, true, excluded, verified_channel.publisher&.site_banners || {}])
```
yachtcaptain23: It should only load the site banner for that channel. We can walk through this together.

db/migrate/20181118005740_add_default_to_site_banner.rb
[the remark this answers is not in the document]
nvonpentz: ++ for model validations.

> Setting a default value can negate the default for the other site banners.

how should we achieve this, model callback?

app/controllers/publishers/site_banners_controller.rb
dlipeles: Done ✔️

app/controllers/publishers/site_banners_controller.rb
dlipeles: There's an edge case where a user hasn't created a channel yet, the easiest way would be to hide the editor from users who havn't created a channel. 

I still think we want to give them access even if they havn't added a channel yet, so I provide them with this 'sandbox' banner they can edit. 

It gets deleted once they add their first channel, but we could even more gracefully handle this by copying the sandbox over to their channel banners once they add the channel(s). 

The alternative as you mentioned is to handle this in the UI, but that requires even more logic than here, (we would have to add edge case logic to the render, fetch, and save functions).

app/controllers/publishers/site_banners_controller.rb
dlipeles: See above.

app/controllers/publishers/site_banners_controller.rb
dlipeles: I agree with your suggestion, but there's a bug here we need to address this in a separate issue, I noticed the `.name` function didn't work on youtube channels, but it does for twitch, once that's resolved I'll refactor this using one method. (And update the `channels_controller` if necessary)

app/controllers/publishers/site_banners_controller.rb
dlipeles: Agree, see above

app/javascript/packs/banner_editor.jsx
dlipeles: Done ✔️

config/routes.rb
dlipeles: That's true, I think I'm more of a fan of explicit definitions though and `channels` gives a better idea of what the function does than `index` IMO

app/controllers/publishers/site_banners_controller.rb
dlipeles: See #1367

db/migrate/20181116173430_add_channel_id_to_site_banner.rb
```diff
@@ -0,0 +1,5 @@
+class AddChannelIdToSiteBanner < ActiveRecord::Migration[5.2]
+  def change
+    add_column :site_banners, :channel_id, :uuid
```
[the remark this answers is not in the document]
dlipeles: For our purposes it should be enough to keep `null: false` , unless you see some benefit to making it an index

db/migrate/20181118005740_add_default_to_site_banner.rb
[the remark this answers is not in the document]
dlipeles: Validation for the default attribute 

```ruby 
  validate :donation_amounts_in_scope, :at_most_one_default

  # Validate user has at most 1 default banner
  def at_most_one_default
    default_banners = SiteBanner.where(publisher_id: publisher_id, default: true).pluck(:default)
    errors.add(:base, "Must have at most one default banner") if default_banners.length > 1
  end
```

app/controllers/publishers/site_banners_controller.rb
dlipeles: Done ✔️ (Removed this function)

app/models/json_builders/channels_json_builder.rb
yachtcaptain23: `verified_channel.publisher&.site_banners` is an ActiveRecord relation. The method `read_only_react_property` is meant to return the value stored for that site banner. Given that we'll have defaults, this should return the default version if there's not already a site banner for that channel.

corymcdonald:
When I run the migration I get the following error
```sh
-- change_column(:site_banners, :default, :boolean, {:null=>false, :default=>false})
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

No such column: site_banners.default
/Users/cory/repos/publishers/db/migrate/20181120193230_add_default_to_site_banner.rb:3:in `change'
bin/rails:4:in `require'
bin/rails:4:in `<main>'

Caused by:
ActiveRecord::ActiveRecordError: No such column: site_banners.default
```

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,75 @@ class Publishers::SiteBannersController < ApplicationController
...
+
+    site_banner = current_publisher.site_banners.find_by(channel_id: params[:channel_id])
+
+    if site_banner
+      data = site_banner.read_only_react_property
+      data[:backgroundImage] = data[:backgroundUrl]
+      data[:logoImage] = data[:logoUrl]
+      render(json: data.to_json)
+    else
+      site_banner = new(params[:channel_id])
```
corymcdonald: This isn't idempotent, a GET shouldn't create a new entry in the database similar to what Albert said. I'm going to write something up on a different way to structure this 👍

app/javascript/packs/banner_editor.jsx
```diff
@@ -181,11 +188,12 @@ export default class BannerEditor extends React.Component {
 
     let that = this
 
-    let url = '/publishers/' + document.getElementById("publisher_id").value + "/site_banners/fetch";
+    let url = '/publishers/' + document.getElementById("publisher_id").value + "/site_banners/fetch?channel_id=" + this.props.channels[this.state.channelIndex].id;
```
corymcdonald: Can we move the publisher_id to a prop that gets passed in?

corymcdonald:
I think we’ll want to modify the approach in this pull request a bit. 

I don’t think we should be creating these mock channels with fake IDs, because then we can potentially have duplicate identifiers in our database. 

I’m curious about why we need to return a fake entry ("00000000-0000-0000-0000-000000000000”) as a response for the javascript code as opposed to having a default value if the response in empty, or have ([defaultProps](https://reactjs.org/docs/typechecking-with-proptypes.html)) handle it. It seems like we know what we want the response to be and so we can move the logic into the front end. 

If there no ID associated with the banner on the front end then we know to route to `/create` as opposed to `/update` 

When we fetch a site banner right now we make a GET call to fetch the banner the current implementation updates a default in the database. Instead of doing that why don’t we instead have the default template that is returned. We don’t need a channel_id as it’s not technically required in the database so perhaps if there’s not a channel_id you can simply say “this is going to be a default” and then set it as such.

app/javascript/packs/banner_editor.jsx
```diff
@@ -425,13 +446,14 @@ export default class BannerEditor extends React.Component {
     // Three POST requests make up the save: Content, Logo, and Cover.
     this.setState({state: 'save', saving: 'true'}, () => Spinner.show('save-spinner', 'save-container'))
 
-    let url = '/publishers/' + document.getElementById("publisher_id").value + "/site_banners";
+    let url = '/publishers/' + document.getElementById("publisher_id").value + "/site_banners/save?channel_id=" + this.props.channels[this.state.channelIndex].id;
```
corymcdonald: This is where I think there should be logic in place to check to see if the channelId exists, if it doesn't then we run `/publishers/:id/site_banners/create` without a channel id. If there's not a channel id then the controller knows that the user is trying to create a default channel

app/javascript/packs/banner_editor.jsx
dlipeles: Done ✔️

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,23 @@ class Publishers::SiteBannersController < ApplicationController
...
-    @site_banner = current_publisher.site_banner || SiteBanner.new
+  def fetch
+    site_banner = current_publisher.site_banners.find_by(channel_id: params[:channel_id])
+    if site_banner
+      data = site_banner.read_only_react_property
+      data[:backgroundImage] = data[:backgroundUrl]
+      data[:logoImage] = data[:logoUrl]
+      render(json: data.to_json)
+    else
+      render(json: nil.to_json)
```
yachtcaptain23: Might be easier to do `{}.to_json`

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,14 +5,23 @@ class Publishers::SiteBannersController < ApplicationController
...
+    end
   end
 
-  def create
-    site_banner = current_publisher.site_banner || SiteBanner.new
+  def save
```
yachtcaptain23: Rename to `def update`

app/models/json_builders/channels_json_builder.rb
```diff
@@ -26,15 +26,31 @@ def initialize
...
         if @excluded_channel_ids.include?(verified_channel.details.channel_identifier)
           excluded = true
           @excluded_verified_channel_ids.push(verified_channel.details.channel_identifier)
         else
           excluded = false
         end
-        channels.push([verified_channel.details.channel_identifier, true, excluded, verified_channel.publisher&.site_banner&.read_only_react_property || {}])
+
+        # Add banners - if default exists add to all channels, else add per channel
+        publisher = Publisher.find_by(id: verified_channel.publisher_id)
```
yachtcaptain23: `verified_channels.channel.publisher_id`

app/controllers/publishers_controller.rb
```diff
@@ -423,6 +423,43 @@ def balance
     end
   end
 
+  def banner_editor_data
+    # Handle edge case where user hasn't created channels yet.
+    no_channels_banner = current_publisher.site_banners.find_by(channel_id: "00000000-0000-0000-0000-000000000000")
```
yachtcaptain23: Would it be possible to get rid of lines 427 to 431?

app/controllers/publishers_controller.rb
```diff
@@ -423,6 +423,43 @@ def balance
...
+    if no_channels_banner && !current_publisher.channels.length.zero?
+      no_channels_banner.delete
+    end
+
+    channels = []
```
yachtcaptain23: 433 to 440 can be refactored to `channels = current_publisher.channels.map { |channel| {id: channel.id, name: channel.publication_title, type: channel.details_type} }`

app/controllers/publishers_controller.rb
```diff
@@ -423,6 +423,43 @@ def balance
...
+                "id" => c.id,
+                "name" => c.publication_title,
+                "type" => c.details_type
+                }
+      channels.push(channel)
+    end
+    # Handle edge case where user hasn't created channels yet.
+    if current_publisher.channels.length.zero?
+      data = no_channels_banner_data
+      return render(json: data.to_json)
```
yachtcaptain23: `render(json: data.to_json) and return` is the convention

app/controllers/publishers_controller.rb
```diff
@@ -554,4 +591,12 @@ def prompt_for_two_factor_setup
   def update_sendgrid(publisher:, prior_email: nil)
     RegisterPublisherWithSendGridJob.perform_later(publisher.id, prior_email)
   end
+
+  def no_channels_banner_data
+    return {
+      "channels": [{"id" => "00000000-0000-0000-0000-000000000000", "name" => " ", "type" => "SiteChannelDetails"}],
```
yachtcaptain23: Preferably we use symbols as opposed to strings. Aside from the slight performance improvement, they're immutable.

app/models/json_builders/channels_json_builder.rb
```diff
@@ -26,15 +26,31 @@ def initialize
   def build
     channels = []
 
-  [Channel.left_joins(publisher: :site_banner).verified.site_channels, Channel.left_joins(publisher: :site_banner).youtube_channels, Channel.left_joins(publisher: :site_banner).twitch_channels, Channel.left_joins(publisher: :site_banner).twitter_channels].each do |verified_channels|
+  [Channel.verified.site_channels, Channel.youtube_channels, Channel.twitch_channels, Channel.twitter_channels].each do |verified_channels|
```
yachtcaptain23: Nit: Would you fix the indentation?

db/migrate/20181116173430_add_channel_id_to_site_banner.rb
yachtcaptain23: Add an index here

app/controllers/publishers_controller.rb
corymcdonald: Really? I know it's a convention throughout the project but I was thinking that Rubocop/the community didn't like it.

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/AndOr

dlipeles:
Rewrite is in - by making a subtle design change as mentioned above, I was able to fix many of the issues we were seeing before. 

#### In the schema 

* Instead of putting a `default (boolean)` on each site_banner, 
we add `default_banner (uuid)` to the publishers table
* Site_banners now have their own `id's` which are generated uuid's
* Publisher has a `default_banner_mode (boolean)`, this is needed since we need to persist both the default banner, and publisher's choice to use one.

#### Improvements

* Site_banners are now oriented around their ids instead of their channel_id fields, retrieving them via the controller is much more natural this way
* Rails resource friendly methods - Instead of `fetch`, `save`, etc. We can now take advantage of the built in `show`, `update` methods
* No-channel edge case is handled much more gracefully now, if a publisher hasn't added channels yet he can style his default banner. 

#### Backwards compatibility

This becomes much more feasible than before, the script to port old banners to the new codebase is roughly:

```
for each publisher
  site_banner = publisher.site_banner
  if site_banner
    uuid = generate_uuid()
    site_banner.update(id: uuid)
    publisher.update(default_banner: uuid)
  end
end
```

dlipeles:
#### Important Note
I am seeing a very unusual bug 

#### Youtube / Twitch channel banners are getting their images conflated with other banners

* I only see this with Youtube / Twitch banners, I tried this out with 7+ unverified site channel banners and did not see the issue at all.  
* AFAIK my code does not make any distinction between the type of channel. 
* It's also really interesting to note that this only affects the images of the banners - updates to the header and tagline text work normally. 
* I tried to compose the update_logo and update_backgroundImage methods after I noticed this bug to see if it would fix it, the results were exactly the same. 

This leads me to believe it may be a deeper issue somehow related to channels, I will continue to debug tomorrow, but I may need some help on this one.

dlipeles:
#### Important Update

Solved the bug above, it is incredibly interesting, there appears to be a bug in the ActiveStorage library itself, this is a relatively new addition to the Rails framework and some other have been reporting bugs, will comment more info shortly.

app/controllers/publishers_controller.rb
dlipeles: Done ✔️

app/controllers/publishers_controller.rb
dlipeles: Done ✔️

app/controllers/publishers_controller.rb
dlipeles: Done ✔️

app/controllers/publishers_controller.rb
dlipeles: Done ✔️

app/models/json_builders/channels_json_builder.rb
dlipeles: Done ✔️

dlipeles:
#### Deployment Plan

We are going to need to give this some thought - but with this rewrite it becomes easier to port existing site_banners to the new codebase. 

The idea will be that any existing site_banners will become that publisher's `default_banner`. 

#### Run rake db:migrate

#### Run the following script 
```ruby
Publisher.all.each do |publisher|
  site_banner = publisher.site_banner
  if site_banner
    uuid = SecureRandom.uuid <---- Do we need to use schema.rb's generate_uuid here?
    site_banner.update(id: uuid)
    publisher.update(default_banner: uuid)
  end
end
```

corymcdonald:
# SiteBanner migration strategy
Part 1 - Migration

We need to add a UUID column to the table as the primary ID. The problem is that there is already an ID that exists for that table
```ruby
 def change
    add_column :site_banners, :uuid, :uuid, default: "uuid_generate_v4()", null: false

    change_table :site_banners do |t|
      t.rename :id, :legacy_id
      t.rename :uuid, :id
    end
    execute "ALTER TABLE site_banners ADD PRIMARY KEY (id);"
  end
```

```ruby
 def change
    add_column :active_storage_attachments, :uuid, :uuid, null: false

    change_table : active_storage_attachments do |t|
      t.rename :record_id, :legacy_record_id
      t.rename :uuid, :record_id
    end
  end
```

Then we need to update all the `ActiveRecord::Attachments` to have the correct record id. As part of the next deployment we can then remove the `legacy_id` column on the table.

SiteBanner.where(

## Part 2-  Rake tasks
We’ll have two rake tasks

1. Migrate the existing attachments to the new site banner UUID field
```ruby
namespace :migrate_site_banners do
  task :migrating_existing_data => :environment do
    SiteBanner.each do |site|
      # legacy_id = big int
      # legacy_record_id =  big int
      attachment = ActiveStorage::Attachment.find(legacy_record_id: site.legacy_id)
      attachment.record_id = site.id # UUID
      attachment.update
    end
  end
end
```

2. Migrate the existing SiteBanners to be the default banner for publishers

dlipeles:
Agree with the migration strategy above, two questions

* `execute "ALTER TABLE site_banners ADD PRIMARY KEY (id);"` 
Do we need something like this in the `active_storage_attachments` table?

* `add_column :active_storage_attachments, :uuid, :uuid, null: false`
Should this have `default: null`?

corymcdonald:
@dlipeles 

I don't think we need to add a primary key, this is the SQL that defines the current table. The primary key is on the `id`, and not the `record_id`.

```sql
CREATE TABLE public.active_storage_attachments
(
    id bigint NOT NULL DEFAULT nextval('active_storage_attachments_id_seq'::regclass),
    name character varying COLLATE pg_catalog."default" NOT NULL,
    record_type character varying COLLATE pg_catalog."default" NOT NULL,
    record_id bigint NOT NULL,
    blob_id bigint NOT NULL,
    created_at timestamp without time zone NOT NULL,
    CONSTRAINT active_storage_attachments_pkey PRIMARY KEY (id)
)
```


As for the `add_column :active_storage_attachments, :uuid, :uuid, null: false`. I don't think we want a `default: null` because that's no according to our current table declaration I found in pgAdmin

dlipeles:
#### Migration strategy (updated)

#### SiteBanner Migration 
```ruby 
class AddUuidToSiteBanner < ActiveRecord::Migration[5.2]
  def change
   add_column :site_banners, :uuid, :uuid, default: "uuid_generate_v4()", null: false

   change_table :site_banners do |t|
     t.rename :id, :legacy_id
     t.rename :uuid, :id
   end
   execute "ALTER TABLE site_banners DROP CONSTRAINT site_banners_pkey;"
   execute "ALTER TABLE site_banners ADD PRIMARY KEY (id);"
 end
end
```

#### Active Storage Attachments 
```ruby
class AddUuidToActiveStorageAttachments < ActiveRecord::Migration[5.2]
  def change
     add_column :active_storage_attachments, :uuid, :uuid, null: false
     change_table :active_storage_attachments do |t|
       t.rename :record_id, :legacy_record_id
       t.rename :uuid, :record_id
     end
     execute "ALTER TABLE active_storage_attachments ALTER COLUMN legacy_record_id DROP NOT NULL;"
   end
 end
```

#### Rake Task
```ruby
namespace :site_banners do
  task :migrate_site_banners => :environment do
    # Update record_id's to use uuids instead of bigints
    SiteBanner.all.each do |site|
      attachment = ActiveStorage::Attachment.find(legacy_record_id: site.legacy_id)
      attachment.record_id = site.id
      attachment.update
    end

    # Set existing banners to default
    Publisher.all.each do |publisher|
      legacy_banner = publisher.site_banners.first
      if legacy_banner
        publisher.update(default_banner: legacy_banner.id)
      end
    end
  end
end
```

dlipeles:
This will be merged in stage 3 of #1406. Once stage 2 is completed, I'll pull in the changes from staging and ensure all migrations are consistent.

dlipeles:
Ready for review, one note - with the most recent refactor we will need to run a rake task to onboard our existing publishers. 

The reason is as follows - the way we create banners now is as an `after_save` function on the Publishers Model (for default banners) and an `after_save` function on the Channels Model (for channel banners). So when a Publisher is created, his default site banner is created. And when a channel is created, a site banner for that channel is created. In the case of our existing publishers and channels, we need to guarantee those banners are created before they use the editor. 

This is probably the cleanest way to do it but it does require a rake task for Publishers and Channels which already exist. Unfortunately we'll need to iterate over all Publishers in the rake task, but it will only need to be run once. 

The alternative would be to have the creation logic trigger before the tipping banner editor is opened, that is, check the current publishers default banner and channels and create and banners that don't exist yet. This is less clean and can slow down the UI, but it would be backwards compatible without requiring a rake task. 

I think most of us would agree that creating the site_banners in the Publisher and Channels model would be the best way to do it.

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -5,27 +5,14 @@ class Publishers::SiteBannersController < ApplicationController
...
-      description: sanitize(params[:description])
-    )
-    head :ok
-  end
-
-  def fetch
-    if current_publisher.site_banner.present?
-      data = current_publisher.site_banner.read_only_react_property
+  def show
+    if params[:default]
```
corymcdonald: Can we extract this out to a method called like `def site_banner`. The implementation would look something like this with memoization.

```ruby
def site_banner
  @site_banner ||= 
    if params[:default]
      current_publisher.default_site_banner
    else 
      current_publisher.site_banners.find_by(id: params[:id])
    end
end
```

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -34,32 +21,42 @@ def fetch
     end
   end
 
-  def update_logo
-    if params[:image].length > MAX_IMAGE_SIZE
-      # (Albert Wang): We should consider supporting alerts. This might require a UI redesign
-      # alert[:error] = "File size too big!"
-      head :payload_too_large and return
+  def update
+    if params[:default]
```
corymcdonald: This can use the new extracted site_banner

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -34,32 +21,42 @@ def fetch
...
-    head :ok
+  def set_default_site_banner_mode
+    current_publisher.update(default_site_banner_mode: params[:dbm])
   end
 
   private
 
   def update_image(attachment:, attachment_type:)
-    data_url = params[:image].split(',')[0]
+
```
corymcdonald: nit: remove newline at beginning of function

app/controllers/publishers_controller.rb
```diff
@@ -422,13 +422,23 @@ def balance
     end
   end
 
+  def get_site_banner_data
+    default_site_banner_mode = current_publisher.default_site_banner_mode
+    default_site_banner = {"id" => current_publisher.default_site_banner_id, "name" => "Default", "type" => "Default"}
+    channel_banners = current_publisher.channels.map { |channel| {id: channel.id, name: channel.publication_title, type: channel.details_type} }
+    channel_banners.each do |channel_banner|
+      channel_banner[:id] = current_publisher.site_banners.find_by(channel_id: channel_banner[:id]).id
```
corymcdonald: I'm a little bit confused about why you're replacing the id of each channel banner with the id of the site banner?
I think in your map above it would be easier to do this
```ruby
channel_banners = current_publisher.channels.map do |channel|
  {
    id: channel.id,
    site_banner_id: channel.site_banner.id,
    name: channel.publication_title,
    type: channel.details_type
  }
end
```

app/javascript/packs/banner_editor.jsx
```diff
@@ -180,19 +196,26 @@ export default class BannerEditor extends React.Component {
   async fetchBanner(){
 
     let that = this
+    let url;
+
+    if(this.state.defaultSiteBannerMode){
+      url = '/publishers/' + document.getElementById("publisher_id").value + "/site_banners/" + this.props.defaultSiteBanner.id + "?default=" + this.state.defaultSiteBannerMode
+    }
+    else{
+      url = '/publishers/' + document.getElementById("publisher_id").value + "/site_banners/" + this.props.channelBanners[this.state.channelIndex].id;
```
corymcdonald: This would change if we replace the `map` function i mentioned about with `site_banner_id`

app/javascript/packs/banner_editor.jsx
```diff
@@ -180,19 +196,26 @@ export default class BannerEditor extends React.Component {
   async fetchBanner(){
 
     let that = this
+    let url;
+
+    if(this.state.defaultSiteBannerMode){
+      url = '/publishers/' + document.getElementById("publisher_id").value + "/site_banners/" + this.props.defaultSiteBanner.id + "?default=" + this.state.defaultSiteBannerMode
```
corymcdonald: Does it matter if the default param is passed in? 
If the site banner belongs to the current publisher then we might not even need to pass the param 🎉

app/models/channel.rb
```diff
@@ -48,6 +50,7 @@ class Channel < ApplicationRecord
   validate :verified_duplicate_channels_must_be_contested, if: -> { verified? }
 
   after_save :register_channel_for_promo, if: :should_register_channel_for_promo
+  after_save :set_channel_banner, if: -> { SiteBanner.find_by(channel_id: self.id).nil? }
```
corymcdonald: Probably better to let the relation do it for you 
```suggestion
  after_save :set_channel_banner, if: -> { self.site_banner.nil? }
```
not sure if you have to have the `self` there or not

app/models/channel.rb
```diff
@@ -281,6 +284,20 @@ def register_channel_for_promo
     RegisterChannelForPromoJob.new.perform(channel: self)
   end
 
+  def set_channel_banner
```
corymcdonald: can we rename this something like `create_default_site_banner`

app/models/channel.rb
```diff
@@ -281,6 +284,20 @@ def register_channel_for_promo
...
+    headline = I18n.t 'banner.headline'
+    tagline = I18n.t 'banner.tagline'
+    channel_banner = SiteBanner.create(
+      publisher_id: self.publisher_id,
+        channel_id: self.id,
```
corymcdonald: what's up with the indentation

app/models/json_builders/channels_json_builder.rb
```diff
@@ -26,15 +26,32 @@ def initialize
...
           @excluded_verified_channel_ids.push(verified_channel.details.channel_identifier)
         else
           excluded = false
         end
-        channels.push([verified_channel.details.channel_identifier, true, excluded, verified_channel.publisher&.site_banner&.read_only_react_property || {}])
+
+        # Add banners - if default_site_banner_mode=true use default_site_banner for all channels, else add channel_banner
+        publisher = Publisher.find_by(id: verified_channel.publisher_id)
+
+        if publisher
```
corymcdonald: I don't think you need to check for this because if a channel doesn't have a publisher then there is something seriously wrong. Too defensive in my opinion

app/models/json_builders/channels_json_builder.rb
```diff
@@ -26,15 +26,32 @@ def initialize
...
+          default_site_banner_mode = publisher.default_site_banner_mode
```
corymcdonald: Lets' just do this so we don't have to create a publisher variable that isn't useful
```suggestion
          default_site_banner_mode = verified_channel.publisher.default_site_banner_mode
```

app/models/channel.rb
yachtcaptain23: Would you fix the indentation here?

app/models/channel.rb
```diff
@@ -281,6 +284,20 @@ def register_channel_for_promo
...
+        title: headline,
```
yachtcaptain23: Fewer lines the better IMO. e.g. `title: I18n.t 'banner.headline'` is the preferred style.

app/models/json_builders/channels_json_builder.rb
```diff
@@ -26,15 +26,32 @@ def initialize
...
+          default_site_banner = publisher.site_banners.find_by(id: publisher.default_site_banner_id)
+          channel_banner = publisher.site_banners.find_by(channel_id: verified_channel.id)
```
yachtcaptain23: I still need to think about this a bit more, but you can avoid this extra call assuming `default_site_banner_mode` exists. Perhaps you can create one block just for lookup and appending to your `channels` array. Preferably, before this block starting on line 30, you might be able to pre-calculate which channels have a publisher, which publishers have a default_site_banner, etc. You can see N+1 queries by running `
ActiveRecord::Base.logger = Logger.new(STDOUT)` in `rails console`.

app/controllers/publishers_controller.rb
dlipeles: Thanks for the tip on using channel.site_banner.id 😎

dlipeles:
### Deployment Plan

#### Enter maintenance mode

* Merge #1357 into `master`
* Once build and migrations are complete run ```rake database_updates:migrate_site_banners```

#### Exit maintenance mode (Estimated time 30 minutes)

lib/tasks/database_updates/create_default_and_channel_site_banners.rake
```diff
@@ -1,20 +1,14 @@
...
   task :create_default_and_channel_site_banners => :environment do
-    # Create default and channel banners for publishers who existed before multi-channel banners feature.
-
-      headline = I18n.t 'banner.headline'
-      tagline = I18n.t 'banner.tagline'
-
       total = Publisher.all.count
       Publisher.find_each.with_index do |publisher, index|
-        puts "#{index} / #{total} complete" if index % 5 == 0
+        puts "#{index} / #{total} complete" if index % 100 == 0
```
corymcdonald: Wanna put a complete message so we know that everything worked at the end?

app/controllers/publishers/site_banners_controller.rb
```diff
@@ -34,32 +16,35 @@ def fetch
...
   end
 
-  def update_logo
-    if params[:image].length > MAX_IMAGE_SIZE
-      # (Albert Wang): We should consider supporting alerts. This might require a UI redesign
-      # alert[:error] = "File size too big!"
-      head :payload_too_large and return
+  def update
+    if site_banner
+      site_banner.update(publisher_id: current_publisher.id, title: sanitize(params[:title]), description: sanitize(params[:description]), donation_amounts: JSON.parse(sanitize(params[:donation_amounts])), default_donation: JSON.parse(sanitize(params[:donation_amounts])).second, social_links: params[:social_links].present? ? JSON.parse(sanitize(params[:social_links])) : {})
```
corymcdonald: i'm normally a 120 characters line length kind of guy, but this is way too long even for me lol

app/controllers/publishers_controller.rb
```diff
@@ -422,13 +422,20 @@ def balance
     end
   end
 
+  def get_site_banner_data
+    default_site_banner_mode = current_publisher.default_site_banner_mode
+    default_site_banner = {:id => current_publisher.default_site_banner_id, :name => "Default", :type => "Default"}
+    channel_banners = current_publisher.channels.map { |channel| {id: channel.site_banner.id, name: channel.publication_title, type: channel.details_type} }
+    data = {:default_site_banner_mode => default_site_banner_mode, :default_site_banner => default_site_banner, :channel_banners => channel_banners}
```
corymcdonald: nit: no rocket (`=>`) syntax

app/controllers/publishers/site_banners_controller.rb
dlipeles: Let me backlog this one item, I agree we should extract out but I don't wanna break anything before deploy 👾

app/models/publisher.rb
```diff
@@ -274,6 +275,17 @@ def publisher?
...
   end
 
+  def default_site_banner
+    self.site_banners.find_by(id: self.default_site_banner_id)
+  end
+
+  def create_default_site_banner
+    headline = I18n.t 'banner.headline'
+    tagline = I18n.t 'banner.tagline'
+    default_site_banner = SiteBanner.create(publisher_id: self.id, channel_id: nil, title: headline, description: tagline, donation_amounts: [1, 5, 10], default_donation: 5, social_links: {youtube: '', twitter: '', twitch: ''})
```
yachtcaptain23: Nit, you can do this in one step:
```
self.default_site_banner = SiteBanner.create(publisher_id: self.id, channel_id: nil, title: headline, description: tagline, donation_amounts: [1, 5, 10], default_donation: 5, social_links: {youtube: '', twitter: '', twitch: ''})
```

dlipeles: Moved this logic to a helper method, the reason for 2 lines instead of 1 is readability

PR #1361, opened 2018-11-19
Remove old referral stats logic

### LINKED ISSUE #1337: Save all referral stats to the stats column in the promo_registrations table

nvonpentz:
Resolves #1337 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:
Resolves #1337 

Reviewer Checklist:

Tests

Security:


corymcdonald:
Can we also remove `PROMO_STATS_UPDATE_DELAY` from `app/models/publisher.rb`? Doesn't look to be used anywhere

corymcdonald:
Also removal of this method
```
app/helpers/promos_helper.rb
19:  def offline_promo_stats
```

nvonpentz:
Replaced by #1637

PR #1366, opened 2018-11-20
Display publisher Uphold authorization status from the Eyeshade wallet to admins

### LINKED ISSUE #1320: Display publisher Uphold authorization status from the Eyeshade wallet to admins
Currently, the admin dashboard reads "Uphold connected: true" for all publishers who have ever connected to Uphold.  This is misleading for publishers who were once connected but are not any more.  Instead of sourcing this from the  `uphold_verified` column in the publishers table, let's use the most up-to-date information which can be found in the Eyeshade wallet.

corymcdonald:
Closes #1320

### DIFF
--- a/app/views/admin/publishers/show.html.slim
+++ b/app/views/admin/publishers/show.html.slim
@@ -19,6 +19,9 @@
           .db-field = "Uphold connected:"
           .db-value = @publisher.uphold_verified?
         .db-info-row
+          .db-field = "Uphold authorized:"
+          .db-value = @publisher.wallet.authorized?
+        .db-info-row
           .db-field = "Uphold widget enabled?"
           .db-value = @publisher.excluded_from_payout? ? "No" : "Yes"
         .db-info-row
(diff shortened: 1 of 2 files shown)


app/views/admin/publishers/show.html.slim
```diff
@@ -18,6 +18,9 @@
         .db-info-row
           .db-field = "Uphold connected:"
```
nvonpentz: could we change the text to 'uphold_verified' to match the database? both 'uphold_verified' and 'uphold connected' don't really give a full picture, but at least 'uphold_verified' matches the database.  we also use the term 'uphold connected' in place of 'uphold authorized' when we display to the pub.

corymcdonald: Updated the nomenclature 🌮 
https://github.com/brave-intl/publishers/pull/1366/commits/43ebb4cfc125e9e6f5052b4f8edfdcb8ee21b8f3

PR #1368, opened 2018-11-20
Request referral code stats in batches of 1000

nvonpentz:
**NOTE** This PR is directly against master.

We are unable to successfully request/save the stats from the promo server because the request URI is too long.  It's too long because each referral code is supplied as a query string parameter,  e.g.

```
GET /api/2/promo/statsByReferralCode?referral_code=PRO123&referral_code=PRO456
```

This PR updates the PromoRegistrationsStatsFetcher to make the request in batches of 1000, so each URI is ~31 kb.  According to [stack overflow](https://stackoverflow.com/questions/32763165/node-js-http-get-url-length-limitation) the limit is 80kb, so we should be well under.

This PR is directly against master since https://github.com/brave-intl/publishers/pull/1350 was merged in to staging.  Since that was merged in to staging, we can't merge `master <- staging` without changing the source of publisher stats from the publishers table to the promo registrations table.  We can't do that since the stats won't be there (since the uri was too long).

We could make this against staging, but there could be some downtime (I think maybe a few minutes) for referral stats.  I think in this case it makes sense to merge into master, then `master -> staging`, but am open to other approaches.

## Tests

The tests insert 2000 new referral codes in the test db with raw sql in order to make a single query instead of 2000 queries for performance.  Unfortunately, this test still takes ~10 seconds to complete because of the subsequent queries at the end of the test and in the PromoRegistrationsStatsFetcher.

I have a couple ideas for ways to speed it up further, but want to hear your thoughts.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/services/promo_registrations_stats_fetcher.rb
+++ b/app/services/promo_registrations_stats_fetcher.rb
@@ -10,20 +10,24 @@ class PromoRegistrationsStatsFetcher < BaseApiClient
 
   def perform
     return perform_offline if Rails.application.secrets[:api_promo_base_uri].blank?
-    response = connection.get do |request|
-      request.options.params_encoder = Faraday::FlatParamsEncoder
-      request.headers["Authorization"] = api_authorization_header
-      request.headers["Content-Type"] = "application/json"
-      request.url("/api/2/promo/statsByReferralCode#{query_string}")
-    end
-    referral_code_events_date_list = JSON.parse(response.body)
+    @referral_codes.each_slice(1000).to_a.each do |referral_code_batch|
+      query_string = query_string(referral_code_batch)
+      response = connection.get do |request|
+        request.options.params_encoder = Faraday::FlatParamsEncoder
+        request.headers["Authorization"] = api_authorization_header
+        request.headers["Content-Type"] = "application/json"
+        request.url("/api/2/promo/statsByReferralCode#{query_string}")
+      end
 
-    @referral_codes.each do |referral_code|
-      promo_registration = PromoRegistration.find_by_referral_code(referral_code)
-      promo_registration.stats = referral_code_events_date_list.select {|referral_code_event_date|
-        referral_code_event_date["referral_code"] == referral_code
-      }.to_json
-      promo_registration.save!
+      referral_code_events_by_date = JSON.parse(response.body)
+
+      referral_code_batch.each do |referral_code|
+        promo_registration = PromoRegistration.find_by_referral_code(referral_code)
+        promo_registration.stats = referral_code_events_by_date.select {|referral_code_event_date|
+          referral_code_event_date["referral_code"] == referral_code
+        }.to_json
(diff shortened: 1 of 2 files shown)


yachtcaptain23:
How long does this take to run under normal circumstances? There's N * M lookups and N different saves. I'm afraid this job would time out.

yachtcaptain23:
One suggestion I can make is make the job call itself after every 100 entries or calls or have a wrapper job that splits the promo codes by into clumps of 100-1000.

app/services/promo_registrations_stats_fetcher.rb
```diff
@@ -10,20 +10,24 @@ def initialize(promo_registrations:)
...
-      request.url("/api/2/promo/statsByReferralCode#{query_string}")
-    end
-    referral_code_events_date_list = JSON.parse(response.body)
-
-    @referral_codes.each do |referral_code|
-      promo_registration = PromoRegistration.find_by_referral_code(referral_code)
-      promo_registration.stats = referral_code_events_date_list.select {|referral_code_event_date|
-        referral_code_event_date["referral_code"] == referral_code
-      }.to_json
-      promo_registration.save!
```
corymcdonald: One thing to note is that this method no longer returns true or false and instead returns nil
That will cause this job to obtain the incorrect result `app/jobs/sync/unattached_promo_registrations_stats_job.rb`

test/services/promo_registrations_stats_fetcher_test.rb
```diff
@@ -12,6 +12,15 @@ class PromoRegistrationsStatsFetcherTest < ActiveJob::TestCase
     Rails.application.secrets[:api_promo_base_uri] = @prev_promo_api_uri
   end
 
+  def query_string(referral_codes)
+    query_string = "?"
+    referral_codes.each do |referral_code|
+      query_string = "#{query_string}referral_code=#{referral_code}&"
+    end
+    query_string.chomp("&") # Remove the final ampersand
```
corymcdonald: feel free to keep this as it is but an alternative would be 
```ruby
codes = referral_codes.map { |r| "referral_code=#{r}" }.join("&")
"?#{codes}"
```

nvonpentz:
> How long does this take to run under normal circumstances? There's N * M lookups and N different saves. I'm afraid this job would time out.

Could you elaborate?  Running this in the console I see 1 lookup and 1 update for each referral code.  Testing locally with 2000 codes took about 10 seconds.

app/services/promo_registrations_stats_fetcher.rb
nvonpentz: hm actually this will return`@referral_codes.each_slice(1000).to_a` if successful, so `app/jobs/sync/unattached_promo_registrations_stats_job.rb` should be fine

PR #1369, opened 2018-11-20
Create channels controller with search

### LINKED ISSUE #1347: Admin search by channel
Let's add a `admin/channel#index` and have a search functionality similar to `admin/publishers#index`

corymcdonald:
Closes #1347

### DIFF
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -213,6 +213,27 @@ class Channel < ApplicationRecord
     Publishers::RestrictedChannels.restricted?(self)
   end
 
+  def self.search(query)
+    query = query.downcase
+
+    channel = Channel
+              .joins(:publisher)
+              .left_outer_joins(:site_channel_details)
+              .left_outer_joins(:youtube_channel_details)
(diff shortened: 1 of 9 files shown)


app/models/channel.rb
```diff
@@ -213,6 +213,27 @@ def needs_admin_approval?
     Publishers::RestrictedChannels.restricted?(self)
   end
 
+  def self.search(query)
```
yachtcaptain23: Perhaps I'm wrong, but I don't want this to be in the model. If another developer uses this function in the publishers controller that accepts from a url header, it's possible we'd leak data.

corymcdonald: I disagree because in the publishers controller we would write something similar to `@publisher.channels` which would not have access to this method since it is on the `Channel` object. 

I'm not sure that developers would write `Channel.search(params[:q])` and expect that it would be safe. In my mind it's the same as doing `Channel.where(params[:q])`

I'm okay with moving this to the controller but I really think this is a model-concern and not a controller concern.

PR #1380, opened 2018-11-27
Payout report refactor

### LINKED ISSUE #1285: [Performance] Payout report generation
Currently payout report generation takes 6 hours. We should address this by parallelizing the `GeneratePayoutReportJob`

### LINKED ISSUE #1397: Up worker thread concurrency count to 5

nvonpentz:
Resolves #1285
Resolves #1397 

This PR introduces major changes do how we request and save payout reports in order to speed up the report generation.  These changes will also allow us to generate reports for a subset of publishers, or add to an existing report without risk of adding duplicate payments.

We speed up the reports by peeling off a separate job for every publisher, so their transactions can be added to the report asynchronously.  Since the legacy reports were only single json field in the payout report table, we weren't able to do this since we'd be running the risk of having multiple jobs update the same database field concurrently and losing data.  The solution in this PR is to create a new `potential_payments` table, and have each payment be it's own row.  This way the concurrent jobs are not editing the same database entries at the same time.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/models/channel.rb
```diff
@@ -167,6 +169,8 @@ def self.find_by_channel_identifier(channel_identifier)
         Rails.logger.info("Unable to find channel for channel identifier #{channel_identifier}")
         nil
     end
+  rescue NoMethodError => e
+    nil
```
yachtcaptain23: Would you use `LogException` rather than allowing a silent error?

app/models/payout_report.rb
```diff
@@ -12,15 +14,36 @@ def encryption_key
     Rails.application.secrets[:attr_encrypted_key]
   end
 
-  def self.total_amount
-    PayoutReport.sum { |payout_report| payout_report.amount.to_i } / 1E18
+  def amount
+    PotentialPayment.where(payout_report_id: id).sum { |potential_payment| potential_payment.amount.to_i }
```
yachtcaptain23: `potention_payments.sum { |potential_payment| potential_payment.amount.to_i }` since you already have a `has_many` relationship.

app/models/payout_report.rb
```diff
@@ -12,15 +14,36 @@ def encryption_key
...
+  end
+
+  def fees
+    PotentialPayment.where(payout_report_id: id).sum { |potential_payment| potential_payment.fees.to_i }
```
yachtcaptain23: Ditto here

app/models/channel.rb
corymcdonald: Can we use the nil safe navigation operator to prevent this?

corymcdonald: Can we use the nil safe navigation operator to prevent this?

app/services/payout_report_publisher_includer.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def perform
+    return if @publisher.suspended? || !@publisher.has_verified_channel?
+    publisher_has_unsettled_balance = false
+
+    wallet = @publisher.wallet
+    return if wallet.nil?
+
+    probi = wallet.channel_balances[@publisher.owner_identifier].probi_before_fees # probi = balance
+    if probi > 0
```
corymcdonald: nit

```suggestion
    if probi.positive?`
```

app/services/payout_report_publisher_includer.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    @publisher.channels.verified.each do |channel|
+      probi = wallet.channel_balances[channel.details.channel_identifier].probi # probi = balance - fee
+      next if probi <= 0 || !@publisher.uphold_verified? || wallet.address.blank?
+
+      publisher_has_unsettled_balance = true
+      fee_probi = wallet.channel_balances[channel.details.channel_identifier].fee # fee = balance - probi
+
+      PotentialPayment.create(payout_report_id: @payout_report.id,
```
corymcdonald: I'm trying to understand the logic a little bit more, why do we create two potential payouts, one for just the publisher and then one for all the publishers channels?

I'm curious, if the publisher doesn't have upload verified should we be preventing these payouts from being created?

app/models/payout_report.rb
```diff
@@ -12,15 +14,36 @@ def encryption_key
...
+  def num_payments
+    PotentialPayment.where(payout_report_id: id).count
   end
 
-  def self.total_payments
-    PayoutReport.sum(:num_payments)
+  # Updates the JSON summary of the report downloaded by admins
+  def update_report_contents
+    payout_report_json = JsonBuilders::PayoutReportJsonBuilder.new(payout_report: self).build
+    self.contents = payout_report_json.to_json
```
yachtcaptain23: IIRC you don't need to actually cast this to JSON format:

https://edgeguides.rubyonrails.org/active_record_postgresql.html#json-and-jsonb

app/services/payout_report_publisher_includer.rb
evq: This is due to eyeshade accounting logic - referrals accrue to to the publisher whereas contributions accrue to the channel. This way if a channel is moved any unsettled balance moves with it.

app/models/channel.rb
nvonpentz: I do think we want to raise a RecordNotFound error if the record isn't found, but we could improve this by using the nil safe operator.  We could use do something like this:

```
result = YoutubeChannelDetails.where(youtube_channel_id: channel_details_type_identifier).
                       joins(:channel).                                joins(:channel).
                       where('channels.verified = true').first&.channel

...

raise ActiveRecord::RecordNotFound.new("Couldn't find Channel with 'channel_identifier'=#{channel_identifier}") if result.nil?
```

app/models/payout_report.rb
nvonpentz: Yes I think that's right, but we do have other jsonb columns which are in json so I did this to keep it consistent.  I'm not opposed to not casting it to json though and move other jsonb columns to do the same in the future.

app/services/payout_report_publisher_includer.rb
nvonpentz: Will change here and a few lines down

nvonpentz:
These changes rely heavily on our app being able to have many concurrent jobs.  Currently concurrency is set two 2, which means 2 threads per process https://github.com/brave-intl/publishers/blob/staging/config/sidekiq.yml#L2.  There is 1 Sidekiq process per worker dyno https://stackoverflow.com/a/43727289.

I think we'll want to increase this concurrency to at least 5, and maybe increase the number of worker dynos.

Redis could be another bottleneck that we may want to upgrade as well, same with the number of database connections.

Also note that we can set concurrency per queue, which we may want to use here.

https://github.com/mperham/sidekiq/wiki/Advanced-Options#concurrency.

Please weigh in.

corymcdonald:
Before we do anything to scale up these numbers we should first measure what our current performance with this pull request is. We won't know if it's within an acceptable threshold or not until we do some testing. I do agree that we will probably need to increase the SideKiq workers. For testing purposes though it might make sense to replicate the prod DB into staging to get and run this job.

app/services/payout_report_publisher_includer.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    wallet = @publisher.wallet
+    return if wallet.nil?
+
+    probi = wallet.channel_balances[@publisher.owner_identifier].probi_before_fees # probi = balance
+    if probi.positive?
+      publisher_has_unsettled_balance = true
+
+      if @publisher.uphold_verified? && wallet.address.present?
+        PotentialPayment.create(payout_report_id: @payout_report.id,
```
yachtcaptain23: Try to lock everything in a transaction: I'm afraid if this function fails after this block, we'd get an extra `PotentialPayment`. Preferably we refactor this class so lookups are the in the beginning, and writes are at the end.

nvonpentz: Hm I don't think we want a transaction here since that would result in a publisher not being paid at all rather them being paid of only some of their channels or referrals, but maybe I'm misunderstanding your comment.

Which line are you worried about failing, and what result would that produce?

app/models/potential_payment.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  REFERRAL = "referral"
+  CONTRIBUTION = "contribution"
+
+  belongs_to :payout_report
+  belongs_to :publisher
+  belongs_to :channel
+
+  validates :channel_id, presence: true, if: -> { kind == CONTRIBUTION}
+  validates :channel_id, presence: false, if: -> { kind == REFERRAL}
```
yachtcaptain23: Would you clarify your intention here? Did you want to make sure it's empty?

I'm going to test on my own side, but would you also verify that `presence: false` works? Looking here, I don't think `presence: false` works https://stackoverflow.com/questions/10070786/rails-3-validation-presence-false

app/models/channel.rb
nvonpentz: Since this is largely not relevant to the PR, I'm going to revert to the original behavior and we can discuss in a different issue.

app/services/payout_report_publisher_includer.rb
yachtcaptain23: So let's say this service fails in the middle (perhaps during a database lookup). `IncludePublisherInPayoutReportJob` will restart and call `PayoutReportPublisherIncluder`.

In finer detail: If `PotentialPayment` for Referrals on line 20 succeeds, and let's say there's a failure while `@publisher.channels.verified.each` is running for PotentialPayment's Contributions, then the retried `IncludePublisherInPayoutReportJob` will have duplicates for (1) `PotentialPayment` for Referrals (2) `PotentialPayment` for Contributions from the previously failed job.

nvonpentz: Thanks for the clarification.  That's a valid concern and one that I try to address in the PotentialPayment validations.

Part of the design goal of the refactor was to make it so you could run the job multiple times and not add any duplicate payments.  That way if a publisher was left out for whatever reason, we could just run the job again and supply the publisher (or just re run for all publishers) and include them without fear of a duplicate payment.

The validations are set up so that a channel can have at most one payment for a given payout report, https://github.com/brave-intl/publishers/pull/1380/files#diff-31551a82b5c4e5a8bcf61f64060eb102R12, and a publisher can only receive one referral payment for a given payout report https://github.com/brave-intl/publishers/pull/1380/files#diff-31551a82b5c4e5a8bcf61f64060eb102R18.

The tests for validations might also help demonstrate what I'm after https://github.com/brave-intl/publishers/pull/1380/files#diff-1db04edc63db4ff7684664f9115af41aR3.

app/models/potential_payment.rb
```diff
@@ -0,0 +1,28 @@
...
+  validates :channel_id, uniqueness: { scope: :payout_report_id }, unless: -> { channel_id.nil? }
+
+  validate :publisher_id_unique_for_referral_payments
+
+  private
+
+  def publisher_id_unique_for_referral_payments
+    referral_payment_for_publisher_already_exists = PotentialPayment.where(payout_report_id: payout_report_id).
```
yachtcaptain23: I believe this method can be refactored to 

`validates :publisher_id, uniqueness: {scope: :payout_report_id}, if: -> { kind == REFERRAL }`

app/models/payout_report.rb
nvonpentz: Will keep as json for now.

PR #1396, opened 2018-11-30
Public file uploads

corymcdonald:
Still waiting for creds for the private s3 bucket. I'll update the Environment Variables / Secrets.yml file when I do so.

Essentially when the developer defines the following

```ruby
class User < ActiveRecord::Base
 has_one_public_s3 :avatar
end
```

And then the developer can call with an upload function to upload the data to our public s3 bucket 

```ruby
User.upload_public_avatar(
  {
    io: open(padded_resized_jpg_path),
    filename: new_filename + ".jpg",
    content_type: "image/jpg"
  }
)
```

They can then request the url by calling
```ruby
User.public_avatar_url
```


.gitignore
```diff
@@ -29,3 +29,22 @@ yarn-error.log
 /node_modules
 .env
 docker-compose.override.yml
+
+
+# Swap
+[._]*.s[a-v][a-z]
```
yachtcaptain23: You can set a global .gitignore for your session:
```
~/.gitignore_global
*~
.DS_Store
```

Add the following to your `$HOME/.gitconfig`
```
[core]
	editor = vim
	excludesfile = /Users/albertwang/.gitignore_global
```

Gemfile
```diff
@@ -159,6 +159,9 @@ group :test do
   # Clean state in-between tests which modify the DB
   gem "database_cleaner"
 
+  # Create a temporary table-backed ActiveRecord model
```
yachtcaptain23: Would you move this gem to the development/test block?

.gitignore
corymcdonald: Addressed https://github.com/brave-intl/publishers/pull/1396/commits/754d9c0c9e09649452942f9dd7abff409cc69ef8

Gemfile
corymcdonald: Addressed 👍 
https://github.com/brave-intl/publishers/pull/1396/commits/1f3709dc373273bdbf7ceba0b395aad18c7b1313

app/models/concerns/public_s3.rb
```diff
@@ -0,0 +1,126 @@
...
+
+        def public_#{name}_url
+          return if #{name}.attachment.blank?
+          if Rails.application.config.active_storage.service == :amazon
+            #{name}.blob.service = public_s3_service
+          end
+
+          # ActiveStorage::Current.host = Rails.application.secrets[:s3_rewards_public_domain]
+          if Rails.env.development? || Rails.env.test?
+            #{name}.service_url
```
yachtcaptain23: I think you'd want to use the URL for this e.g. `localhost:3000`

app/models/concerns/public_s3.rb
```diff
@@ -76,13 +82,20 @@ def upload_public_#{name}(file)
...
           # ActiveStorage::Current.host = Rails.application.secrets[:s3_rewards_public_domain]
           if Rails.env.development? || Rails.env.test?
-            #{name}.service_url
+            ActiveStorage::Current.host = 'https://localhost:3000'
+            filename = ActiveStorage::Filename.wrap(#{name}.blob.filename)
+
+            # Supports both DiskService and the PublicS3 Service
+            public_s3_service.url(
+              #{name}.blob.key,
+              expires_in: public_s3_service.url_expires_in,
```
yachtcaptain23: `undefined method `url_expires_in' for nil:NilClass`

PR #1406, opened 2018-12-04
Migrations for UUID upgrade

### LINKED ISSUE #1400: Convert to UUID for site banners and active storage attachments
[Link to recommended documentation:](http://www.madebyloren.com/posts/migrating-to-uuids-as-primary-keys) 

Only need to implement:
`Add UUIDs to existing tables` and `Migrate Foreign Keys to UUIDs` and finally `Migrate Primary Keys to UUIDs`

dlipeles:
Closes #1400 

#### Migrations
```ruby 
class AddUuidToSiteBanner < ActiveRecord::Migration[5.2]
  def change
   add_column :site_banners, :uuid, :uuid, default: "uuid_generate_v4()", null: false

   change_table :site_banners do |t|
     t.rename :id, :legacy_id
     t.rename :uuid, :id
   end
   execute "ALTER TABLE site_banners DROP CONSTRAINT site_banners_pkey;"
   execute "ALTER TABLE site_banners ADD PRIMARY KEY (id);"
 end
end
```

```ruby
class AddUuidToActiveStorageAttachments < ActiveRecord::Migration[5.2]
  def change
     add_column :active_storage_attachments, :uuid, :uuid
     change_table :active_storage_attachments do |t|
       t.rename :record_id, :legacy_record_id
       t.rename :uuid, :record_id
     end
     publisher = Publisher.find_by(id: banner.publisher_id)
     if publisher
        publisher.update!(default_banner: banner.id)
     end
     execute "ALTER TABLE active_storage_attachments ALTER COLUMN legacy_record_id DROP NOT NULL;"
   end
 end
```

##### Enter Maintenance Mode on Production
```ruby
namespace :database_updates do
  task :migrate_site_banners => :environment do
    # Update record_id's to use uuids instead of bigints
    SiteBanner.find_each do |banner|
      attachments = ActiveStorage::Attachment.where(legacy_record_id: banner.legacy_id)
      attachments.each do |attachment|
        attachment.update!(record_id: banner.id)
      end
    end
  end
end
```
#### Migrations 

```ruby
class RemoveLegacyRecordIDForActiveStorageAttachments < ActiveRecord::Migration[5.2]
  def change
     remove_column :active_storage_attachments, :legacy_record_id
   end
 end
```

```ruby
class RemoveLegacyIDForSiteBanner < ActiveRecord::Migration[5.2]
  def change
     remove_column :site_banners, :legacy_id
   end
 end
```

```ruby
class AddIndexToActiveStorageAttachments < ActiveRecord::Migration[5.2]
  def change
     add_index :active_storage_attachments, [:record_type, :record_id, :name, :blob_id], name: "index_active_storage_attachments_uniqueness", unique: true
   end
 end
```

```ruby
class AddNullConstraintToActiveStorageAttachments < ActiveRecord::Migration[5.2]
  def change
    change_column_null :active_storage_attachments, :record_id, false
   end
 end
```

```Q. Would it be preferred to compose these four into one single migration? ```

#### Stage 3 -> 
* Merge PR #1357 
* Merge PR #1396

### CODE REVIEW

File: lib/tasks/migrate_site_banners.rake [hunk 1]
```diff
@@ -0,0 +1,18 @@
+namespace :site_banners do
+  task :migrate_site_banners => :environment do
```
nvonpentz: would this be better under the existing `database_updates` namespace?

lib/tasks/migrate_site_banners.rake
```diff
@@ -0,0 +1,18 @@
...
+    # Update record_id's to use uuids instead of bigints
+    SiteBanner.all.each do |site|
+      attachment = ActiveStorage::Attachment.find(legacy_record_id: site.legacy_id)
+      attachment.record_id = site.id
+      attachment.update
+    end
+
+    # Set existing banners to default
+    Publisher.all.each do |publisher|
```
nvonpentz: let's use `find_each` for collections > 1000 https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each

db/migrate/20181204022249_add_uuid_to_active_storage_attachments.rb
```diff
@@ -0,0 +1,10 @@
+class AddUuidToActiveStorageAttachments < ActiveRecord::Migration[5.2]
+  def change
+     add_column :active_storage_attachments, :uuid, :uuid, null: false
```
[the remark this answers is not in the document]
yachtcaptain23: Ah nevermind. I see that we want to keep the original `id` and use `uuid` as a `record_id`.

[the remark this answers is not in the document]
yachtcaptain23: Just tested this out, I believe you need to drop the `null: false` requirement and add it in later

lib/tasks/migrate_site_banners.rake
```diff
@@ -0,0 +1,18 @@
+namespace :site_banners do
+  task :migrate_site_banners => :environment do
+    # Update record_id's to use uuids instead of bigints
+    SiteBanner.all.each do |site|
+      attachment = ActiveStorage::Attachment.find(legacy_record_id: site.legacy_id)
```
yachtcaptain23: use `find_by` rather than `find`

lib/tasks/migrate_site_banners.rake
```diff
@@ -0,0 +1,18 @@
+namespace :site_banners do
+  task :migrate_site_banners => :environment do
+    # Update record_id's to use uuids instead of bigints
+    SiteBanner.all.each do |site|
```
yachtcaptain23: Actually you should iterate over ActiveStorage::Attachment because there's 1 logo and 1 background per SiteBanner.

lib/tasks/migrate_site_banners.rake
dlipeles: Done ✔️

lib/tasks/migrate_site_banners.rake
dlipeles: Done ✔️ I'm thinking using `find_each` is enough optimization here, given we'll only run this task once during a maintenance mode. Unless you think we'll run into problems iterating over Publishers, lmk your thoughts.

lib/tasks/migrate_site_banners.rake
```diff
@@ -0,0 +1,18 @@
...
+      attachment = ActiveStorage::Attachment.find(legacy_record_id: site.legacy_id)
+      attachment.record_id = site.id
+      attachment.update
+    end
+
+    # Set existing banners to default
+    Publisher.all.each do |publisher|
+      legacy_banner = publisher.site_banners.first
+      if legacy_banner
+        publisher.update(default_banner: legacy_banner.id)
```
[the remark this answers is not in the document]
dlipeles: Done ✔️

lib/tasks/migrate_site_banners.rake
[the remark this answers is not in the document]
dlipeles: Done ✔️

lib/tasks/migrate_site_banners.rake
dlipeles: Done ✔️

lib/tasks/migrate_site_banners.rake
dlipeles: Done ✔️

db/migrate/20181204022249_add_uuid_to_active_storage_attachments.rb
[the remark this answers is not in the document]
dlipeles: Done ✔️

db/schema.rb
```diff
@@ -19,11 +19,12 @@
...
     t.string "name", null: false
     t.string "record_type", null: false
-    t.bigint "record_id", null: false
+    t.bigint "legacy_record_id"
     t.bigint "blob_id", null: false
     t.datetime "created_at", null: false
+    t.uuid "record_id"
     t.index ["blob_id"], name: "index_active_storage_attachments_on_blob_id"
-    t.index ["record_type", "record_id", "name", "blob_id"], name: "index_active_storage_attachments_uniqueness", unique: true
+    t.index ["record_type", "legacy_record_id", "name", "blob_id"], name: "index_active_storage_attachments_uniqueness", unique: true
```
corymcdonald: I think I initially said this is probably fine to keep in the index but I thought about this more - 

We'll get in trouble with this because we require unique on legacy_record_id but now that it's `nil` and that won't be unique

lib/tasks/database_updates/migrate_site_banners.rake
```diff
@@ -0,0 +1,15 @@
+namespace :database_updates do
+  task :migrate_site_banners => :environment do
+    # Update record_id's to use uuids instead of bigints
+    SiteBanner.find_each do |banner|
```
corymcdonald: Can we a little progress meter to this? Something that will report how far it's gone?
Maybe something like 
```suggestion
    total = SiteBanner.all.length
    SiteBanner.find_each.with_index do |banner, index|
      puts "#{index} / #{total} complete" if index % 5 == 0
```
And then one at the end

db/schema.rb
```diff
@@ -95,7 +96,7 @@
     t.string "name"
     t.string "email"
     t.string "verification_token"
-    t.boolean "verified", default: false
+    t.boolean "verified", default: true
```
corymcdonald: Is this intended? I'm not seeing a correlating migration

nvonpentz: I'm getting this occasionally when running `rake db:migrate` locally.  Weird - this would be a bad bug.

db/schema.rb
dlipeles: Let me know if my understanding is correct here:

legacy record id would need to be unique if we are creating new site banners and attachments, but we are fine as long as we are only working with existing banners (this would all be happening during maintenance mode). Once we run the rake task, we can get rid of the legacy id fields, **then** we merge in the PR for multi channel mode and return from maintenance mode.

lib/tasks/database_updates/migrate_site_banners.rake
dlipeles: Done ✔️

db/schema.rb
dlipeles: Not intended, Nick do you ever see this in other branches too? I commented out my migrations and still saw this bit flip. 

We could fix this with the line 
```ruby
    change_column_default :legacy_publishers, :verified, false
```

I can put this in the ```RenameRecreatePublishers```, migration - where legacy publishers was created. Although I've been following Cory's advice not to modify older migrations. 

Alternatively, I can create a new migration for the above line, let me know which you guys would prefer.

db/schema.rb
dlipeles: We can do it this way

```ruby 
remove_index :active_storage_attachments, name: :index_active_storage_attachments_uniqueness
```

then add the index again once the second deploy is finished and the legacy id's are removed. 

```ruby 
add_index :active_storage_attachments, [:record_type, :record_id, :name, :blob_id], name: "index_active_storage_attachments_uniqueness", unique: true
```

dlipeles:
Updated the migration strategy to highlight the three stages of deployment.

db/migrate/20171230160309_rename_recreate_publishers.rb
```diff
@@ -33,5 +33,6 @@ def change
       t.index :pending_email
       t.index :created_at
     end
+    change_column_default :legacy_publishers, :verified, false
```
yachtcaptain23: Don't include this migration.

yachtcaptain23:
FYI (for my own benefit): the instructions to run the rake task: `rake site_banners:migrate_site_banners`

lib/tasks/database_updates/migrate_site_banners.rake
```diff
@@ -0,0 +1,17 @@
...
+    total = SiteBanner.all.count
+    SiteBanner.find_each.with_index do |banner, index|
+      puts "#{index} / #{total} complete" if index % 5 == 0
+      attachments = ActiveStorage::Attachment.where(legacy_record_id: banner.legacy_id)
+      attachments.each do |attachment|
+        attachment.update!(record_id: banner.id)
+      end
+      publisher = Publisher.find_by(id: banner.publisher_id)
+      if publisher
+        publisher.update!(default_banner: banner.id)
```
yachtcaptain23: This won't work since the table doesn't include the default banner. Would you move the migration to add default banner into this one?

dlipeles: That's a good point, let me remove that piece from the rake task, and I'll include it in the PR for #1357 when we get to stage 3 of the migration plan.

db/migrate/20171230160309_rename_recreate_publishers.rb
dlipeles: Ok I created a new migration for it, we need to ensure this null property is enforced.

dlipeles:
> FYI (for my own benefit): the instructions to run the rake task: `rake site_banners:migrate_site_banners`



> FYI (for my own benefit): the instructions to run the rake task: `rake site_banners:migrate_site_banners`

Additionally I think we'll want to enter maintenance mode on production while we're running this rake task.

db/migrate/20171230160309_rename_recreate_publishers.rb
yachtcaptain23: The changes to that column isn't relevant because it'll be changing the `legacy_publishers` table which we don't use.

dlipeles:
Tested this locally using a copy of our production database, the migration was successful and it looks like default banners were set correctly

However there was some error on the attachments, resulting in the error 
```
Module::DelegationError: record_id delegated to attachment, but attachment is nil
```

Not sure yet if the cause is in the migration or the rake task, will debug and report back.

db/migrate/20181205180923_add_default_banner_to_publishers.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultBannerToPublishers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :publishers, :default_banner, :uuid
```
yachtcaptain23: this should be `default_site_banner_id`

dlipeles: Done ✔️

PR #1407, opened 2018-12-04
Auto-select end-date for tomorrow to cover timezone issues

### LINKED ISSUE #1395: Promo Reports are possibly offset by a day
When we generate Promo Reports, we probably want to see all the metrics inclusive of this moment. The current logic suggests "Today", which might mean tomorrow given the data is stored in UTC and we're in America. 

I'll be submitting a fix to adjust the dropdown menu default to UTC rather than using an American timezone.

yachtcaptain23:
Default checked

Closes #1395

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/views/admin/unattached_promo_registrations/index.html.slim
+++ b/app/views/admin/unattached_promo_registrations/index.html.slim
@@ -103,7 +103,7 @@
               . = date_select :referral_code_report_period, :start
             .flex-two.align-self-flex-center
               . = label_tag "Report period end"
-              . = date_select :referral_code_report_period, :end, default: 1.day.from_now
+              . = date_select :referral_code_report_period, :end, default: Time.now.utc.to_date
             .flex-two.align-self-flex-center
               . = label_tag "Reporting interval"
               . = select_tag :reporting_interval, options_for_select([["Daily",PromoRegistration::DAILY], ["Weekly",PromoRegistration::WEEKLY], ["Monthly",PromoRegistration::MONTHLY], ["Running total", PromoRegistration::RUNNING_TOTAL]])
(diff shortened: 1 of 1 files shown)


app/views/admin/unattached_promo_registrations/index.html.slim
```diff
@@ -103,20 +103,20 @@
               . = date_select :referral_code_report_period, :start
             .flex-two.align-self-flex-center
               . = label_tag "Report period end"
-              . = date_select :referral_code_report_period, :end
+              . = date_select :referral_code_report_period, :end, default: 1.day.from_now
```
nvonpentz: So I think what `1.day.from_now` does is take the current day in PST and add one day to it.  What I think we want is to have it be the current day, but in UTC time.

yachtcaptain23: This is intentional. Preferably we do 2 days from now (I don't in what case it would happen) but I think this "good-enough" solution is ideal. 

Alternatively, if we start supporting promo codes in Japan/South Korea/Australia/China, the UTC date is wrong.

nvonpentz: Why is showing the date 2 days from the current time PST more preferable than the exact day that the referral server will be recording the downloads at the current time?  Seems like this would be adding ambiguity.

Don't we already support those countries? Why would the UTC date be wrong?

yachtcaptain23: I'm just highlighting that we should show more info rather than less.

Supposing that the promo server saves everything in UTC, then yes, you're right.

Issue #1385, opened 2018-11-29
Additional OpenSSL checks


corymcdonald:
We currently rescue `OpenSSL::SSL::SSLError` in `app/services/site_channel_host_inspector.rb` however we do not do this in `app/services/site_channel_verifier.rb`. This is causing some SSL errors to be raised and not presented in a user-friendly format. 

The impact of this is essentially that #253 has not been fully completed

Issue #408, opened 2017-12-18
HTTPS check fails for *.github.io sites if the root returns 404


nvonpentz:
### Description
When checking if a domain is using https, we do a lookup on the root url that returns false if we get back a 404.  This is a problem for many .github.io sites because many times the root does not have a dedicated page, and will return a 404.  At this point the https checker will will report that https is not enabled, and the user cannot register their site.

Example:
nvonpentz.github.io  -  _returns 404_
nvonpentz.github.io/publishers/ -  _returns 200_

### Steps to reproduce
1. Sign up a new publisher account, and verify the email
2. Click "Site Operator"
3. Enter "nvonpentz.github.io" as the domain, click "Continue"
4. Click "I'll Use a Trusted File"

#### Expected result
Instructions on adding the verification file to .well-known folder

#### Actual result
"Uh oh! Your domain is NOT using HTTPS. You will need to fix that before continuing."

### Potential solutions
1.  On the `verification_github` page, link to blog post explaining how make the root of the .github.io page return a 200
2. Change the domain normalizer to check .github.io/repo 
3. Change the https checker to return true for a domain with https even if it returns a 404
4. Some other strategy (please comment)

ayumi:
i think we only need to require that the verification file is served at the correct URL with HTTPS
i.e. https://{name}.github.io/.well-known/brave-payments-verification.txt

yachtcaptain23:
This should be fixed with https://github.com/brave-intl/publishers/pull/910

nvonpentz:
@yachtcaptain23 This is still reproducible.

PR #1410, opened 2018-12-04
Better handle HTTPS and other errors

corymcdonald:
Resolves the following two issues

https://github.com/brave-intl/publishers/issues/408
https://github.com/brave-intl/publishers/issues/1385

🎉

### DIFF
--- a/app/services/site_channel_verification_file_generator.rb
+++ b/app/services/site_channel_verification_file_generator.rb
@@ -14,9 +14,8 @@ class SiteChannelVerificationFileGenerator < BaseService
     "brave-payments-verification.txt"
   end
 
-  # NOTE: Tell user http:// is acceptable but https:// is preferred
-  def generate_url
-    "https://#{site_channel.details.brave_publisher_id}/.well-known/#{filename}"
+  def url
+    "#{site_channel.details.brave_publisher_id}/.well-known/#{filename}"
   end
 
   def generate_file_content
--- a/app/services/site_channel_verifier.rb
+++ b/app/services/site_channel_verifier.rb
@@ -107,30 +107,26 @@ class SiteChannelVerifier < BaseService
 
   def verify_site_channel_public_file
     generator = SiteChannelVerificationFileGenerator.new(site_channel: channel)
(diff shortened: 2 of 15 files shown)


corymcdonald:
Nick reported that `http://google.com` isn't resolving so I'm going to look into that

corymcdonald:
Ready for review 🎉

app/services/site_channel_verification_file_generator.rb
```diff
@@ -14,9 +14,8 @@ def filename
     "brave-payments-verification.txt"
   end
 
-  # NOTE: Tell user http:// is acceptable but https:// is preferred
-  def generate_url
-    "https://#{site_channel.details.brave_publisher_id}/.well-known/#{filename}"
+  def url
+    "#{site_channel.details.brave_publisher_id}/.well-known/#{filename}"
```
nvonpentz: Where is this method used?

app/services/site_channel_verifier.rb
```diff
@@ -107,30 +107,25 @@ def verify_site_channel_dns
...
   def verify_site_channel_public_file
     generator = SiteChannelVerificationFileGenerator.new(site_channel: channel)
-    uri = URI("https://#{channel.details.brave_publisher_id}/.well-known/#{generator.filename}")
-    response = fetch(uri: uri)
-    if response.code == "200"
-      token_match = /#{channel.details.verification_token}/.match(response.body)
-      if token_match
+
+    host_inspector = SiteChannelHostInspector.new(
+      brave_publisher_id: generator.url,
```
nvonpentz: `brave_publisher_id` is used to refer to the root domain of the site.  It looks like `generator.url` includes the path to the public file. If the we will be using the SiteChannelHostInspector more generally, can we change the name of this parameter?

corymcdonald: Renamed it to be `url`

https://github.com/brave-intl/publishers/pull/1410/commits/8720d181801543d2ad5240b6e54fcc1d9371f31c

PR #1414, opened 2018-12-05
Add test coverage reports

corymcdonald:
When you run `rails test` there will be a coverage report in `coverage/index.html`

We're currently at 76.19%

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -1,4 +1,7 @@
 ENV["RAILS_ENV"] ||= "test"
+require 'simplecov'
+SimpleCov.start 'rails'
+
 require File.expand_path("../../config/environment", __FILE__)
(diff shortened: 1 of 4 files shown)


test/test_helper.rb
```diff
@@ -1,4 +1,7 @@
 ENV["RAILS_ENV"] ||= "test"
+require 'simplecov'
+SimpleCov.start 'rails'
+
```
dlipeles: When we get back up to 90% let's revisit this and add [minimum coverage](https://github.com/colszowka/simplecov#minimum-coverage)

```ruby
SimpleCov.minimum_coverage 90
```

There are other options too if we want to get more fancy 🍸

corymcdonald: Agreed!

PR #1417, opened 2018-12-05
Create organization permissions

### LINKED ISSUE #1351: Specify Partner Organization Permissions
Preferably the permissions are specified in a new table. Admin will have to edit the permissions after the account is created (by default, it has 0 permissions).

E.g.
```
PartnerOrganizationPermissions:
  t.uuid organization_id (To the Publishers table)
  t.boolean can_upload_offline_billing
  t.boolean can_upload_offline_invoice
  t.boolean can_upload_referral_codes
```

corymcdonald:
## Create Organization Permissions
An important thing to note is that this does not change anything that's on the user side. This is admin facing only. 

This additionally fleshes out the concept of an Organization by adding permissions to it.

## What was changed?

1. Added create/update workflows for organizations
1. Updated the "Make a partner" button in the publisher show page to direct to an update method on a partner. 
1. Partners now must be associated with an organization, and users can choose that relationship via drop down box.
1. Created link to organization from publishers show page
1. Added relationship to `Organization` so you don't have to type `organization.memberships.members` and now only have to type `organization.members`
1. Added relationship from Partner to organization so now you can simply type `partner.organization` to get the corresponding organization
1. Added a bunch of new tests to `partners_controller` and `organizations_controller` to increase test coverage to 100% 

Closes #1351

### DIFF
--- a/app/controllers/admin/organizations_controller.rb
+++ b/app/controllers/admin/organizations_controller.rb
@@ -40,9 +40,9 @@ module Admin
     end
 
     def permissions(organization_permissions)
-      organization_permissions.uphold_wallet = params[:uphold] == "1"
-      organization_permissions.referral_codes = params[:referral_codes] == "1"
-      organization_permissions.offline_reporting = params[:offline_reporting] == "1"
+      organization_permissions.uphold_wallet = params[:uphold].present?
+      organization_permissions.referral_codes = params[:referral_codes].present?
+      organization_permissions.offline_reporting = params[:offline_reporting].present?
 
       organization_permissions
     end
--- a/app/controllers/admin/organizations_controller.rb
+++ b/app/controllers/admin/organizations_controller.rb
@@ -4,8 +4,50 @@ module Admin
       @organizations = Organization.paginate(page: params[:page])
     end
 
+    def new
+      @organization = Organization.new
+    end
+
+    def create
+      @organization = Organization.new(organization_params)
+      @organization.permissions = permissions(OrganizationPermission.new)
+
+      if @organization.save
+        redirect_to admin_organization_path(@organization.id)
+      else
+        render :new
+      end
+    end
+
     def show
-      @organization = Organization.find_by(id: params[:id])
+      @organization = Organization.find(params[:id])
+    end
+
+    def edit
+      @organization = Organization.find(params[:id])
+    end
+
+    def update
+      @organization = Organization.find(params[:id])
+      @organization.permissions = permissions(@organization.permissions)
+
+      if @organization.update(organization_params)
+        redirect_to admin_organization_path(@organization.id)
+      else
+        render :edit
+      end
+    end
+
+    def permissions(organization_permissions)
+      organization_permissions.upload_offline_billing = params[:billing] == "1"
+      organization_permissions.upload_offline_invoice = params[:invoice] == "1"
+      organization_permissions.upload_referral_codes = params[:referral] == "1"
+
+      organization_permissions
+    end
+
+    def organization_params
+      params.require(:organization).permit(:name)
     end
   end
 end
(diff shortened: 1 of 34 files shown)


db/migrate/20181204201916_create_organization_permissions.rb
```diff
@@ -0,0 +1,10 @@
+class CreateOrganizationPermissions < ActiveRecord::Migration[5.2]
+  def change
+    create_table :organization_permissions do |t|
+      t.uuid :organization_id
+      t.boolean :uphold_wallet
+      t.boolean :offline_reporting
+      t.boolean :referral_codes
+    end
```
yachtcaptain23: Would you add an index to include `organization_id`?

app/controllers/admin/organizations_controller.rb
```diff
@@ -4,8 +4,51 @@ def index
...
+
+      if @organization.update(organization_params)
+        redirect_to admin_organization_path(@organization.id)
+      else
+        render :edit
+      end
+    end
+
+    def permissions(organization_permissions)
+      organization_permissions.uphold_wallet = params[:uphold] == "1"
```
yachtcaptain23: I believe the default value is `"1"` and you set the value to be `1` if checked. I'm under the impression that `params[:uphold].present?` is more conventional.

db/migrate/20181204201916_create_organization_permissions.rb
yachtcaptain23: Also, would you set default permissions to `false`?

app/controllers/admin/organizations_controller.rb
corymcdonald: ah! I forgot that when the input is unchecked it won't send that params
https://github.com/brave-intl/publishers/pull/1417/commits/c18483e3b2a1235d313ab4596da663d713fe7290

db/migrate/20181204201916_create_organization_permissions.rb
corymcdonald: sure thing! 
https://github.com/brave-intl/publishers/pull/1417/commits/c18483e3b2a1235d313ab4596da663d713fe7290

db/migrate/20181204201916_create_organization_permissions.rb
```diff
@@ -0,0 +1,12 @@
+class CreateOrganizationPermissions < ActiveRecord::Migration[5.2]
+  def change
+    create_table :organization_permissions do |t|
```
yachtcaptain23: default uuid.

I tried looking up a way to automatically add this for us. Do you have a suggestion?

app/views/admin/organizations/index.html.slim
```diff
@@ -3,16 +3,19 @@ div.row
     section.panel
       header.panel-heading.organization-container
         h4 Organizations
+        = link_to "New Organization", new_admin_organization_path, class: "btn btn-primary"
```
dlipeles: Do we need i18n here?

app/controllers/admin/organizations_controller.rb
```diff
@@ -4,8 +4,51 @@ def index
...
+    end
+
     def show
-      @organization = Organization.find_by(id: params[:id])
+      @organization = Organization.find(params[:id])
+    end
+
+    def edit
+      @organization = Organization.find(params[:id])
+    end
```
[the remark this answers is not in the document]
corymcdonald: Unfortunately not because the `routes.rb` looks for this method when the user goes to `organizations/:id/edit`

app/views/admin/organizations/index.html.slim
corymcdonald: I once argued that we should make our admin panel use i18n but I think since it's only internal and a few people have access then we don't have to worry about it.

app/controllers/admin/organizations_controller.rb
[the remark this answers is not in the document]
dlipeles: Nvm this is for the form 👍

corymcdonald:
Good catch! I didn't think to add it 😅

PR #1418, opened 2018-12-06
Public API for verified channels and publisher counts. Closes #1413

### LINKED ISSUE #1413: Issue: Need valid/correct number on verified publishers
We need to have a source of truth that consistently reflects;

a) total verified publisher counts
b) total verified channel counts
c) total YT channel counts
d) total Twitch channel counts

added to the Admin Console dashboard.
and sync'd with stats.brave.com for consistency reasons.

@yachtcaptain23 this is a P1 - would be good to get a fix quickly.

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/api/v1/public/channels_controller.rb
+++ b/app/controllers/api/v1/public/channels_controller.rb
@@ -7,4 +7,15 @@ class Api::V1::Public::ChannelsController < Api::V1::Public::BaseController
     end
     render(json: channels_json, status: 200)
   end
+
+  def totals
+    render(
+      json: {
+        all_channels: Channel.verified.count,
+        twitch_only: Channel.verified.twitch_channels.count,
+        youtube_only:  Channel.verified.youtube_channels.count
+      },
+      status: 200
+    )
+  end
 end
--- a/app/controllers/api/v1/public/channels_controller.rb
+++ b/app/controllers/api/v1/public/channels_controller.rb
@@ -9,6 +9,6 @@ class Api::V1::Public::ChannelsController < Api::V1::Public::BaseController
   end
 
   def totals
-    render(json: Channel.statistical_tools, status: 200)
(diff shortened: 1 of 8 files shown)


config/routes.rb
```diff
@@ -113,8 +113,13 @@
       # /api/v1/public/
       namespace :public, defaults: { format: :json } do
         get "channels", controller: "channels"
+        namespace :channels, defaults: { format: :json } do
```
corymcdonald: I think it might make more sense to have it like
```suggestion
        resources :channels, only: [:index] do
```
Then you can remove the route above it and it will be easier to add on more RESTlike routes later

yachtcaptain23: Good catch. I'll fix

app/controllers/api/v1/public/channels_controller.rb
```diff
@@ -7,4 +7,15 @@ def channels
...
   end
-end
\ No newline at end of file
+
+  def totals
+    render(
+      json: {
+        all_channels: Channel.verified.count,
+        twitch_only: Channel.verified.twitch_channels.count,
+        youtube_only:  Channel.verified.youtube_channels.count
```
corymcdonald: should we add websites? That number should add up to the `totals`

I'm not sure if i'm a big fan of appending `_only` at the end

config/routes.rb
yachtcaptain23: Actually I reverted this given that we'd have to attach a `id` in the curl request.

app/controllers/api/v1/public/channels_controller.rb
yachtcaptain23: I removed `_only`. `site` wasn't in the requirement document but I'll add anyway.

PR #1430, opened 2018-12-10
Add status to notes and index

### LINKED ISSUE #1422: Add PublisherStatusUpdates to the PublisherNotes index

### LINKED ISSUE #1427: Show latest Publisher Status in admin/publishers#index

corymcdonald:
This adds the current status to the admin publishers index and also adds the last found status when a note was created

Closes #1422 
Closes #1427


app/views/admin/publishers/show.html.slim
```diff
@@ -110,6 +110,8 @@
             = note.created_by.name
           .note-created-at
             = note.created_at.strftime("%B %d, %Y")
+          .note-created-at
```
yachtcaptain23: It's possible for a status change to appear after a note has been created. Additionally, it's possible for multiple status changes to occur between notes.

corymcdonald: This is just a CSS class, the actual functionality is that for every comment we'll look up the last corresponding publisher status that was created before the comment.
https://github.com/brave-intl/publishers/pull/1430/commits/59aac3dddc4207bec6b01d0710c4ab21380e63d6

yachtcaptain23: Oh I'm aware, I should have selected the line below. It's just that the logic for `note.publisher_status` reflects the most recent publisher status before the publisher note. It's preferable to see the sequence of actions taken against a publisher, e.g.

1) Publisher is created at a datetime 
2) Alice posts a comment at a datetime
3) Bob posts a comment at a datetime
4) Publisher's status is changed at a datetime
5) Charlie posts a comment at a datetime

This UX model is similar to Github in which we see all actions related to an issue, or if you used JIRA, it's similar to their UX model.

I realize it's a bit more work, but it's more familiar with how other issue tracking.

app/models/publisher.rb
```diff
@@ -150,6 +150,21 @@ def email_verified?
...
+  #
+  # Returns an array of PublisherNote and PublisherStatusUpdate
+  def history
+    # Create hash with created_at time as the key
+    # Then we can merge and sort by the key to get history
+    notes = self.notes.map { |n| { n.created_at => n} }
+    status = self.status_updates.map { |s| { s.created_at => s } }
+
+    combined = notes + status
+    combined = combined.sort { |x, y| x.keys.first <=> y.keys.first}.reverse
```
yachtcaptain23: If the key is in string notation for datetime, wouldn't you have to cast this the string?

yachtcaptain23:
Overall looks good. Left comment inline.

app/models/publisher.rb
corymcdonald: Actually Ruby allows for any object to be the key so fortunately this is in the right type when it's being sorted. 

```
[4] pry(#<Publisher>)> combined[0].keys[0].class
=> ActiveSupport::TimeWithZone
```

Issue #1404, opened 2018-12-03
Do not apply fee when displaying referral payment balance


nvonpentz:
The Eyeshade::Wallet is mostly composed of Eyeshade::Balances.  There is one balance per channel, and then one overall balance which includes all the channel accounts and the publisher account (for referrals).  The overall balance is known as the contribution balance:  https://github.com/brave-intl/publishers/blob/staging/lib/eyeshade/wallet.rb#L34.  At the moment, fees are automatically applied to any kind of Eyeshade::Balance during initialization: https://github.com/brave-intl/publishers/blob/staging/lib/eyeshade/balance.rb#L5.  The problem is that we are applying a fee for the entire the contribution balance which may also include referral payments.  We must only apply fees for channel accounts, not publisher (referral) accounts.

yachtcaptain23:
Solved in #1438

Issue #1309, opened 2018-11-02
Use the PublisherTransactionsGetter to determine last settlement balance


nvonpentz:
Instead of using eyeshade's /wallet endpoint, let's source the last settlement balance from the eyeshade transactions table so we can remove more legacy eyeshade code.

yachtcaptain23:
Per Nick, this is done by the wallet refactor

PR #1438, opened 2018-12-12
Wallet refactor

nvonpentz:
In this pull request we refactor how wallet and balance information is requested, stored in memory and displayed.

Resolves https://github.com/brave-intl/publishers/issues/1404
Resolves https://github.com/brave-intl/publishers/issues/1309

## Existing infrastructure

Eyeshade exposes three endpoints which are used to get wallet information:
  * `/wallet` - Used by the PublisherWalletGetter to get wallet information.  This is mostly information associated with the current wallet provider e.g. possible currencies, available currencies, authorization status.  Right now, that Uphold is the only wallet provider.  Before the eyeshade transactions table, balance information was served from this endpoint.
  * `/balances` - Used by the PublisherBalanceGetter to get balance information for a list of accounts.  An account can be a channel or a publisher.  Balances are returned in BAT.
  * `/transactions` - Used by the PublisherTransactionsGetter to get a the list of transactions executed for a publisher's eyeshade accounts.  This endpoint is currently used for publisher statements.  However after this PR it will also be used for last settlement balances instead of the `/wallet` endpoint where it's currently sourced from.

Information from these endpoints is used to create an Eyeshade::Wallet with one Eyeshade::Balance to represent the overall balance for a publisher and one Eyeshade::Balance for each of their accounts.

When a publisher confirms or selects a new default currency, we make a request to `/publishers/balance` asynchronously with js.  This endpoint uses the wallet json builder to create a summary of the wallet, and then send that json summary to the client.  The client then initializes the result of that using the wallet.js class, and updates the balances displayed on the page.

## Changes
As highlighted in https://github.com/brave-intl/publishers/issues/1404, a single Eyeshade::Balance class is not sufficient to cover the variations in the different types of balances.  For example, a channel balance needs to have fees taken out.  Referral balances to do not.  Overall balances need to be the correct sum of all balances while maintaining that fees are applied correctly for each of the sub balances. LastSettlementBalances need a notion of a settlement currency, while the other balances do not.  For these reasons I have created these classes:
  * Eyeshade::BaseBalance - The base balance that all the other balances inherit from.  This class covers fee calculation and the conversion of balances to different currencies.
  * Eyeshade::OverallBalance - This balance will be the main balance displayed to the publisher
  * Eyeshade::ReferralBalance
  * Eyeshade::ChannelBalance
  * Eyeshade::LastSettlementBalance

In addition to solving the problem of fee calculation, this change will allow us to add new types of balances, e.g. Eyeshade::AdsBalance fairly easily.

Another problem with the existing Eyeshade::Balance class, it that it's configured to accept balance information in the format that the `/wallet` sends.  However we no longer are using `/wallet` for balance information, so a lot of additional data processing is needed to get the `/balances` response into the old format.  This PR removes that additional processing and the new balance classes accept as input the responses from the `/balances` endpoint without additional processing.

Lastly, significant changes were made to the javascript representation of the wallet.  There are a number of problems with the current implementation.  First, the wallet json builder is creating a new representation of the wallet that's different from the Eyeshade::Wallet.  This makes it difficult to make code changes the Eyeshade::Wallet in isolation of the js wallet, and requires that devs maintain two mental models. Second, fee calculation is done client side and server side when it only needs to be server side.

In this PR I try to make wallet.js of the wallet tightly resemble wallet.rb.  To this end I removed the wallet json builder entirely, and replaced `/publishers/balance` with `/publishers/wallet` which simply sends `publisher.wallet.to_json` to the client.  Fee calculation is no longer done client side, we no longer will be summing per channel balances to get an overall balance client side.

## Testing
Ensure
* Overall balances are the sum of channel balances after fees + referral balance (no fee)
* Conversions to default currency / settlement currency are calculated correctly
* Safely handle cases when no default currency is present, or there is no rate for their default currency (e.g. display unavailable).
* All balances are displayed with two decimal places
* Last settlement balances correctly identifies only the most recent settlement transactions and displays those in BAT and in their settlement currency
* All of the above is true after a client side update to the balance (you can trigger by changing your default currency)
* Admin dashboard loads without crashing on the per publisher page, and displays the wallet information
* Payout reports capture the correct probi amounts with respect to fees for the referral and channel balances

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:
This is a work in progress.

Reviewer Checklist:

Tests

Security:


app/helpers/publishers_helper.rb
```diff
@@ -19,11 +19,19 @@ def publisher_can_receive_funds?(publisher)
...
-      '%.2f' % balance.convert_to(currency)
+
+  def next_deposit_date(today = DateTime.now)
+    if today.day > 8
+      today = today + 1.month
+    end
+    today.strftime("%B 8th")
+  end
+
+  def publisher_overall_bat_balance(publisher)
```
corymcdonald: Since we're refactoring this I think that it would be cleaner create an ActiveModel class for wallet. That would remove the need for having all these helpers which are scoped to `publisher.wallet`. 

I was thinking something like this. 

https://gist.github.com/corymcdonald/97914936a2f8b70e997e6a49a0f76e1a

Then in the `model/publisher.rb` file we could have something like this for the wallet
```ruby
  # API call to eyeshade
  def wallet
    @wallet ||= Wallet.new(self)
    @wallet = nil if @wallet.eyeshade_wallet.nil?
    @wallet
  end
```

However, it might not be worth creating an extra class just for the sake of isolation. 🤔

app/helpers/publishers_helper.rb
```diff
@@ -33,64 +41,77 @@ def publisher_humanize_balance(publisher, currency)
     I18n.t("helpers.publisher.balance_unavailable")
   end
 
-  def next_deposit_date(today = DateTime.now)
-    if today.day > 8
-      today = today + 1.month
+  def publisher_converted_overall_balance(publisher)
+    return if publisher.default_currency == "BAT" || publisher.default_currency.blank?
+    if overall_balance = publisher.wallet &&
+       publisher.wallet.overall_balance.is_a?(Eyeshade::OverallBalance) &&
```
corymcdonald: Can we remove all these `is_a?` checks. Let's just go with the [nil-safe-navigation operator](https://rubyinrails.com/2017/11/17/safe-navigation-operator-ampersand-dot-in-ruby/)

```ruby
  def publisher_overall_bat_balance(publisher)
    overall_ballance = publisher.wallet&.overall_balance
    if overall_balance&.amount_default_currency.present?
      I18n.t(
        "helpers.publisher.balance_pending_approximate",
        amount: '%.2f' % overall_balance.amount_default_currency,
        code: overall_balance.default_currency
      )
    else 
      I18n.t("helpers.publisher.conversion_unavailable", code: publisher.default_currency)
    end
  end
```

app/helpers/publishers_helper.rb
```diff
@@ -33,64 +41,77 @@ def publisher_humanize_balance(publisher, currency)
...
+        publisher.wallet.channel_balances[channel_identifier]
+    )
+
+      '%.2f' % balance.amount_bat
     else
-      I18n.t("helpers.publisher.conversion_unavailable", code: currency)
+      I18n.t("helpers.publisher.balance_unavailable")
     end
   rescue => e
     require "sentry-raven"
```
corymcdonald: I've pretty much checked with sentry and didn't find any errors in this helper function.Should we keep these exception trackers in?

lib/eyeshade/base_balance.rb
```diff
@@ -0,0 +1,51 @@
...
+                :default_currency,
+                :amount_probi,
+                :amount_bat,
+                :amount_default_currency,
+                :fees_probi,
+                :fees_bat,
+                :fees_default_currency
+
+    # Account types
+    CHANNEL = "channel".freeze
```
corymcdonald: Let's instead do this at the top of the file. [Ruby 3.0 is going to make strings immutable by default.](https://wyeworks.com/blog/2015/12/1/immutable-strings-in-ruby-2-dot-3)
```ruby
# frozen_string_literal: true
```

lib/eyeshade/base_balance.rb
```diff
@@ -0,0 +1,51 @@
...
+    OWNER = "owner".freeze
+
+    def initialize(rates, default_currency)
+      @rates = rates
+      @default_currency = default_currency || "BAT"
```
corymcdonald: In the `publisher_converted_overall_balance` we check to see if the default_currency is blank. Since this is defaulting to be BAT then we can remove that check ✔️

lib/eyeshade/base_balance.rb
```diff
@@ -0,0 +1,51 @@
...
+      "#{probi.to_i.to_d / 1E18}"
+    end
+
+    # Expects and returns a string
+    def bat_to_probi(bat)
+      "#{(bat.to_d * BigDecimal.new('1.0e18')).to_i}"
+    end
+
+    # Expects and returns values with probi unit
+    def calc_fees(total_probi)
```
corymcdonald: can we spell out the entire word here? `calculate_fees` Just makes it less confusing ya' know

lib/eyeshade/last_settlement_balance.rb
```diff
@@ -0,0 +1,38 @@
...
+      @settlement_currency = last_settlement["currency"]
+      @amount_settlement_currency = convert(@amount_bat, @settlement_currency)
+    end
+
+    private
+
+    def calc_last_settlement(transactions)
+      return {} if transactions == []
+       # Find most recent settlement transaction
+      last_settlement_date = transactions.select { |transaction|
```
corymcdonald: Let's change this to be `do |transaction|` instead of a `{ |transaction|`

lib/eyeshade/referral_balance.rb
```diff
@@ -0,0 +1,21 @@
+# No fees applied
+module Eyeshade
+  class ReferralBalance < BaseBalance
+    def initialize(rates, default_currency, accounts)
+      super(rates, default_currency)
+      owner_account = accounts.select { |account| account["account_type"] == OWNER}.first || {}
+
+      unless owner_account["balance"].nil?
```
corymcdonald: Let's swap this logic 

```ruby
if owner_account["balance"].nil?
  @amount_bat = "0.00"
else
  @amount_bat = owner_account["balance"]
end
```

app/helpers/publishers_helper.rb
nvonpentz: So I think rescuing the error here is important in the case that there is some unexpected state, or if there are issues with what eyeshade returns.  Since we're rescuing here, I think we might as well report the error to sentry.

I'm generally trying to be cautious until these changes have been stable in production for awhile.

lib/eyeshade/last_settlement_balance.rb
nvonpentz: I used this syntax so I can chain operations to the result of the block, e.g. the `.last["created_at"].to_date`:

```ruby
last_settlement_date = transactions.select { |transaction|
  transaction["settlement_amount"].present?
}.last["created_at"].to_date
```

When I wrote this I didn't think you could achieve the same thing with `do ... end`, but apparently you can.  I think it looks kind of funky though.

```ruby
last_settlement_date = transactions.select  do |transaction|
  transaction["settlement_amount"].present?
end.last["created_at"].to_date
```

What do you think?

corymcdonald: I think we should just do something like this

```ruby
last_settlement_date = transactions
	.select { |t| t["settlement_amount"].present? }
	.last["created_at"].to_date
```

lib/eyeshade/base_balance.rb
nvonpentz: I've removed defaulting to BAT since some users are unable withdraw cryptos because of regional restrictions.  Uphold does not have a [bitlicense](https://en.wikipedia.org/wiki/BitLicense).

corymcdonald:
Can we coordinate with eV to help us test this in a staging environment? I think it would be helpful to have a playground to test these changes.

dlipeles:
^ Agree, I would like to help test this on staging too

app/services/payout_report_publisher_includer.rb
```diff
@@ -12,8 +12,8 @@ def perform
     wallet = @publisher.wallet
     return if wallet.nil?
 
-    probi = wallet.channel_balances[@publisher.owner_identifier].probi_before_fees # probi = balance
-    if probi.positive?
+    probi = wallet.referral_balance.amount_probi # probi = balance
+    if probi.to_i.positive?
```
nvonpentz: this `to_i` isn't needed since probi should already be an integer

lib/eyeshade/base_balance.rb
```diff
@@ -0,0 +1,52 @@
...
+      fees_probi = (total_probi * fee_rate).to_i
+      amount_probi = total_probi - fees_probi
+      {"amount" => amount_probi, "fees" => fees_probi}
+    end
+
+    # Expects and returns BigDecimals
+    def convert(amount_bat, currency)
+      return amount_bat if currency == "BAT"
+      return if @rates[currency].nil?
+      amount_bat * @rates[currency]
```
nvonpentz: I'm finding that `@rates[currency]` is actually a string, not a decimal, so I'm getting
```
String can't be coerced into BigDecimal
```
on this line.

I think if we just cast `@rates[currency]` to a BigDecimal then we should be fine.

I think I didn't noticed this error when developing because the rates in the offline wallet are not strings.  It'd be great if we could change that too.

yachtcaptain23:
Merged in via https://github.com/brave-intl/publishers/pull/1644

PR #1448, opened 2018-12-14
Move no-channels edge case logic from UI to server

dlipeles:
Improves the flow for users in the case where they have:

`Not added any channels but created a default banner`
`Then have added their first channels`

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -423,7 +423,7 @@ class PublishersController < ApplicationController
   end
 
   def get_site_banner_data
-    prepare_site_banner_data()
+    prepare_site_banner_data
     default_site_banner_mode = current_publisher.default_site_banner_mode
     default_site_banner = {:id => current_publisher.default_site_banner_id, :name => "Default", :type => "Default"}
     channel_banners = current_publisher.channels.map { |channel| {id: channel.site_banner.id, name: channel.publication_title, type: channel.details_type} }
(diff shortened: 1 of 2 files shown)


app/controllers/publishers_controller.rb
```diff
@@ -423,7 +423,7 @@ def balance
   end
 
   def get_site_banner_data
-    create_site_banners()
+    prepare_site_banner_data()
```
corymcdonald: no parenthesis on methods

app/controllers/publishers_controller.rb
```diff
@@ -451,6 +451,9 @@ def create_site_banners
         channel.site_banner = SiteBanner.new_helper(current_publisher.id, channel.id)
       end
     end
+    if current_publisher.channels.length.zero?
+      current_publisher.update(default_site_banner_mode: true)
```
corymcdonald: let's do this 
```ruby
current_publisher.update(default_site_banner_mode: true) if current_publisher.channels.length.zero?
```

dlipeles: how bow dis? 

```
    if current_publisher.channels.length.zero?
      current_publisher.update(default_site_banner_mode: true)
    else
      current_publisher.channels.each do |channel|
        channel.site_banner = SiteBanner.new_helper(current_publisher.id, channel.id) if channel.site_banner.nil?
      end
    end
```

corymcdonald: sorry! I didn't see that it had an else statement. 👍

PR #1451, opened 2018-12-15
Feature/admin ui updates

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/views/admin/publishers/index.html.slim
+++ b/app/views/admin/publishers/index.html.slim
@@ -20,11 +20,17 @@ div.row
               th Name
               th Email
               th Status
+              th Number of verified channels
+              th Account creation time
+              th Last signed in at
             tbody
-              - @publishers.each do |publisher|
+              - @publishers.left_joins(:channels).each do |publisher|
                 tr.gradeX
                   td = link_to publisher.id, admin_publisher_path(publisher.id)
                   td = publisher.name
                   td = publisher.email
                   td = publisher_status(publisher)
+                  td = publisher.channels.verified.count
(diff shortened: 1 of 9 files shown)


app/views/admin/publishers/index.html.slim
```diff
@@ -20,11 +20,17 @@ div.row
...
-              - @publishers.each do |publisher|
+              - @publishers.left_joins(:channels).each do |publisher|
                 tr.gradeX
                   td = link_to publisher.id, admin_publisher_path(publisher.id)
                   td = publisher.name
                   td = publisher.email
                   td = publisher_status(publisher)
+                  td = publisher.channels.verified.count
+                  td = publisher.created_at
+                  td = publisher.last_sign_in_at
```
nvonpentz: I find dates with timezones kind of hard to read, what do you think about using [distance_of_times_in_words](https://apidock.com/rails/ActionView/Helpers/DateHelper/distance_of_time_in_words) and displaying their last sign in relative to the current time?

package.json
```diff
@@ -83,6 +83,6 @@
     "tslint-config-standard": "^7.0.0",
     "tslint-react": "^3.2.0",
     "webpack-cli": "^3.1.2",
-    "webpack-dev-server": "3.1.10"
+    "webpack-dev-server": "^3.1.10"
```
nvonpentz: Is this change intended?

nvonpentz: I think we'll want to revert the changes to yarn.lock as well

PR #1458, opened 2018-12-17
Uplift to Ruby 2.4.5

corymcdonald:
Upgrading to Ruby 2.4.5


app/models/payout_report.rb
```diff
@@ -3,7 +3,7 @@ class PayoutReport < ApplicationRecord
   self.per_page = 8
 
   LEGACY_PAYOUT_REPORT_TRANSITION_DATE = "2018-12-01 09:14:58 -0800"
-  
+
   attr_encrypted :contents, key: :encryption_key, marshal: true
```
yachtcaptain23: Do we need to do `attr_encrypted ... key: Base64.decode64(key)`?

test/fixtures/publishers.yml
```diff
@@ -1,4 +1,4 @@
-<% salt = ROTP::Base32.random_base32 %>
+<% salt = SecureRandom.random_bytes(12) %>
```
yachtcaptain23: Why 12?

yachtcaptain23:
Would you update the commit to say something along the lines of:

```
If you're seeing `SSL error, peer: 127.0.0.1, peer cert: , #<Puma::MiniSSL::SSLError: OpenSSL error: error:141F7065:SSL routines:final_key_share:no suitable key share - 337604709>`

Please try changing your Gemfile to use the fixes documented in https://github.com/puma/puma/issues/1670
```

PR #1462, opened 2018-12-18
Allow admins to generate geo referral reports

### LINKED ISSUE #1461: Allow admins to generate referral reports with country data
Currently admins can generate referral reports given a list of codes, a reporting interval (daily, monthly, running total), and a date range.  The reports include the number of downloads, installations, and confirmations for each referral code for every interval within the date range supplied.

We would like to extend this behavior to such that admins can see the breakdown of downloads,  installations, and finalizations per referral code within the interval supplied for each country.

First we must have the referral server expose this information (covered in https://github.com/brave-intl/vault-promo-services/issues/74).

Then we should integrate this information into the referral report generation flow:
* Add an option on the infinity code dashboard that allows admins to generate reports with country options. 
* Create a Promo::RegistrationsGeoStatsFetcher service to hit the new endpoint on the referral server
* Update the Promo::RegistrationStatsReportGenerator to be general enough to handle both standard reports and geo reports.
* Update  `/views/admin/unattached_promo_registrations/report.html.slim` to be general enough to handle both standard and geo reports.

nvonpentz:
Resolves #1461 
Depends on https://github.com/brave-intl/vault-promo-services/pull/76

## Changes
* Create a Promo::RegistrationsGeoStatsFetcher to hit the new geo stats endpoint
The bulk of the changes are to the Promo::StatsReportGenerator service:
* The service no longer pulls the recent stats from referral server into our database then uses the values in our database, instead we directly use the stats returned from the service.  We make this change because geo stats will not be saved in the publishers database, and we want the service to be general enough to cover both cases.  Generally I hope we move to not saving referral stats at all and pulling them on demand like we will be for admin geo stats.
* Modify the service to also allow for reports for both geo and standard reports

Understanding the report generation logic is fairly mentally taxing (at least for me) so I tried to make the Promo::StatsReportGenerator service more readable with private methods.  Any other advice is appreciated.

## Testing
The best testing would be to pull down https://github.com/brave-intl/vault-promo-services/pull/76 and a copy of the promo database, connect it to this branch of publishers and verify the numbers in the reports match the numbers in the `download` and `retrievals` tables in the promo services db.

Please see the tests for the format of the report contents.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/controllers/admin/unattached_promo_registrations_controller.rb
```diff
@@ -32,16 +32,18 @@ def report
     referral_codes = params[:referral_codes]
     @reporting_interval = params[:reporting_interval]
     @event_types = params[:event_types]
+    @geo = params[:geo].present?
```
nvonpentz: There's probably a better name for this param here and also in the Promo::RegistrationStatsGeneratorService

app/helpers/promos_helper.rb
```diff
@@ -94,7 +94,7 @@ def event_type_column_header(event_type)
     when PromoRegistration::FIRST_RUNS
       "Installs"
     when PromoRegistration::FINALIZED
-      "Confirmations"
+      "Confirms"
```
nvonpentz: This change was strictly for fitting things onto one line without wrapping

app/services/promo/registrations_geo_stats_fetcher.rb
```diff
@@ -0,0 +1,47 @@
...
+        request.headers["Content-Type"] = "application/json"
+        request.url("/api/2/promo/geoStatsByReferralCode#{query_string}")
+      end      
+      geo_stats = geo_stats + JSON.parse(response.body)
+    end
+    geo_stats
+  end
+
+  def perform_offline
+    true
```
nvonpentz: Hm, I will add an offline mode.

app/services/promo/registrations_stats_fetcher.rb
```diff
@@ -17,17 +18,18 @@ def perform
...
         promo_registration = PromoRegistration.find_by_referral_code(referral_code)
         promo_registration.stats = referral_code_events_by_date.select {|referral_code_event_date|
           referral_code_event_date["referral_code"] == referral_code
         }.to_json
         promo_registration.save!
       end
+      stats = stats + referral_code_events_by_date
     end
+
+    stats
```
nvonpentz: I made this change so we can directly use the result of this service without looking in the database for the promo registration stats.

app/services/promo/registration_stats_report_generator.rb
```diff
@@ -2,113 +2,140 @@
...
+        if @geo
+          events_by_referral_code_for_interval_by_country = group_events_by_country(events)
+
+          per_country_totals = {}
+          events_by_referral_code_for_interval_by_country.each do |country, events|
+            per_country_totals[country] = sum_totals(events)
+          end
+
+          # If there are any events for any country other than N/A, remove the blank country data
+          per_country_totals.delete("N/A") if per_country_totals.keys.count > 1
```
corymcdonald: Where is this `N/A` coming from?

nvonpentz: If there were no events for a given referral code on a specific day, then there will be no recording of that day in the /geoStatsByReferralCode endpoint.  However it's a requirement that we explicitly show 0 when there were no events on that day.

Without country data it was easy enough to just to create 0 values for all possible days, then insert the real data from the referral server:  If there was a download on that day, then we can sum 0 with the number of downloads and get the correct amount.  If there wasn't then 0 was the correct thing to display

With country data it's a little trickier, since we need to associate a country with these 0 inserts.  In this case I'm inserting "N/A".  If there are downloads for that day and there is a country associated (say United States), then we want to remove it since it doesn't make sense to have an entry with N/A if there is another entry on the same day for United States.  Hope that makes sense.

I'm inserting in these blank events here:
https://github.com/brave-intl/publishers/pull/1462/files#diff-bc3a9bb0922dd200752c37c9a23e4df3R122

There is probably some room for improvement here, though a better solution that's simple has escaped me thus far.

app/controllers/admin/unattached_promo_registrations_controller.rb
yachtcaptain23: Preferably `@geo` is `@is_geo` to indicate a boolean type.

app/services/promo/registrations_geo_stats_fetcher.rb
```diff
@@ -0,0 +1,66 @@
...
+      end      
+      geo_stats = geo_stats + JSON.parse(response.body)
+    end
+    geo_stats
+  end
+
+  def perform_offline
+    geo_stats = []
+    @referral_codes.each do |referral_code|
+      ((1.month.ago.to_date)..(Time.now.to_date)).each do |day|
```
yachtcaptain23: We might want to cast to `utc`

app/services/promo/registrations_geo_stats_fetcher.rb
```diff
@@ -0,0 +1,66 @@
...
+    Rails.application.secrets[:api_promo_base_uri]
+  end
+
+  def api_authorization_header
+    "Bearer #{Rails.application.secrets[:api_promo_key]}"
+  end
+
+  def query_string(referral_codes)
+    query_string = "?"
+    referral_codes.each do |referral_code|
```
corymcdonald: nit: this might be easier

```ruby
referral_codes.map { |r| "referral_code=#{referral_code}" }.join("&")
```

app/services/promo/registration_stats_report_generator.rb
```diff
@@ -2,113 +2,140 @@
...
-          existing_event[PromoRegistration::FIRST_RUNS] += event[PromoRegistration::FIRST_RUNS]
-          existing_event[PromoRegistration::FINALIZED] += event[PromoRegistration::FINALIZED]
+
+    events = fetch_stats
+    events = select_events_within_report_range(events)
+    events = fill_in_events_with_no_activity(@referral_codes, events)
+
+    events_by_referral_codes = group_events_by_referral_code(events)
+
+    events_by_referral_codes.each do |referral_code, events|
```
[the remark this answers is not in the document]
nvonpentz: I am not happy with `events` either.  We could be more explicit, but then we end up with names like `events_for_referral_code_for_interval_for_country`.  Very open to suggestions here

app/services/promo/registrations_geo_stats_fetcher.rb
nvonpentz: so this will not work since it does not have `?`, but we could tack it on at the beginning

app/services/promo/registration_stats_report_generator.rb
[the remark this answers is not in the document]
yachtcaptain23: In this loop, what's stored in `events`?

app/views/admin/unattached_promo_registrations/report.html.slim
```diff
@@ -13,12 +13,24 @@ scss:
...
         tbody
-          - events.each do |event|
-            tr
-              - if @reporting_interval != PromoRegistration::RUNNING_TOTAL
-                td = event.first
-              - @event_types.each do |event_type|
-                td = event.second[event_type]
\ No newline at end of file
+          - if @is_geo
+            - events.each do |date, events|
```
yachtcaptain23: IMO instead of `events` as the value, I think `events_on_date` sounds more intuitive.

app/services/promo/registration_stats_report_generator.rb
```diff
@@ -2,113 +2,137 @@
...
+    events = fill_in_events_with_no_activity(@referral_codes, events)
+    events_by_referral_codes = group_events_by_referral_code(events)
+
+    events_by_referral_codes.each do |referral_code, events|
+      events = order_events_by_date(events)
+      events_by_referral_code_for_interval = group_events_by_date(events)
+
+      # Sum the events per interval, or per interval per country if @is_geo
+      per_interval_totals = {}
+      events_by_referral_code_for_interval.each do |date, events|
```
yachtcaptain23: IMO the value should be `events_on_date` or something similar.

yachtcaptain23: And you could rename the function and save a line/variable declaration like so:

```
events_by_referral_codes.each do |referral_code, events|
  group_events_by_date(order_events_by_date(events)).each do |date, events_on_date|
    ...
```

@corymcdonald @nvonpentz let me know if you think this is more or less readable (I believe fewer lines and variables the better, but interpretations on readability can be different)

app/services/promo/registration_stats_report_generator.rb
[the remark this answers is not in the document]
nvonpentz: renamed with more specific names

PR #1472, opened 2018-12-19
Menu to split channels

### LINKED ISSUE #1463: Dropdown-menu to split channels
On the sidebar panel, it would be good to have submenus that accordion open/close like such:

- Channels
  - Websites
  - Twitch
  - YouTube

That way we can specify the columns in more detail based on the type it is.

corymcdonald:
Closes #1463 

screen shot 2018-12-19 at 5 00 49 pm
screen shot 2018-12-19 at 5 00 56 pm
screen shot 2018-12-19 at 5 00 58 pm
screen shot 2018-12-19 at 5 01 00 pm

### DIFF
--- a/app/controllers/admin/channels_controller.rb
+++ b/app/controllers/admin/channels_controller.rb
@@ -7,7 +7,7 @@ module Admin
       @channels = if sort_column&.to_sym&.in? Channel::ADVANCED_SORTABLE_COLUMNS
         Channel.advanced_sort(sort_column.to_sym, sort_direction)
       else
-        Channel.order(("#{sort_column} #{sort_direction}"))
+        Channel.order(sanitize_sql_for_order("#{sort_column} #{sort_direction}"))
       end
 
       if params[:q].present?
--- a/app/controllers/admin/channels_controller.rb
+++ b/app/controllers/admin/channels_controller.rb
@@ -26,7 +26,7 @@ module Admin
           @channels = @channels.twitch_channels
       end
 
-      @channels = @channels.paginate(page: params[:page])
+      @channels = @channels.group(:id).paginate(page: params[:page])
     end
 
     private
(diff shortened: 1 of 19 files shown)


app/controllers/admin/channels_controller.rb
```diff
@@ -3,16 +3,36 @@ class ChannelsController < AdminController
...
+      case params[:type]
+      when 'website'
+          @channels = @channels.site_channels
+      when 'youtube'
+          @channels = @channels.youtube_channels
+      when 'twitch'
+          @channels = @channels.twitch_channels
+      end
+
+      @channels = @channels.paginate(page: params[:page])
```
yachtcaptain23: We might need a `@channels.group(:id)` to prevent duplicates.

app/javascript/packs/admin.js
```diff
@@ -30,3 +30,27 @@ Rails.href = function Rails_href_override(element) {
 };
 
 Rails.start();
+
+
+document.addEventListener('click', function (event) {
+  // If the clicked element doesn't have the right selector, bail
+  if (event.target.matches('#sidebar-toggle, #sidebar-toggle *, i')) {
```
yachtcaptain23: This is fine for now, but down the road we should allow the accordion feature as part of a `class` rather than an `id`.

corymcdonald: Agreed, we could fairly easily extract this into a function but I decided not to do that for now 👍

app/controllers/admin/channels_controller.rb
corymcdonald: I'm not sure if we allow duplicate channels but it could hurt to ensure none pop up

https://github.com/brave-intl/publishers/pull/1472/commits/8b2d8a77652d756d19fb8c801e09a8a43c5fcf08

PR #1487, opened 2019-01-02
Re-create versions table 

### LINKED ISSUE #1421: PaperTrail versions item_id might be using 10 digit int

corymcdonald:
This pull request renames the current versions table to `legacy_versions` and creates a new `versions` table. A migration script has been written to convert the previous version table to the new version table by looking at the `version.object` property and extracting the id.

I cloned our production database and it took around 3 hours to complete. We do not need to be in maintenance mode during this migration.

Closes #1421


corymcdonald:
The migration command is 
```sh
rails database_updates:migrate_versions
```

lib/tasks/database_updates/versions_backfill.rake
```diff
@@ -0,0 +1,55 @@
...
+
+      new_version = Version.new.tap do |v|
+        v.item_type = legacy_version.item_type
+        v.item_id = item_id
+        v.event = legacy_version.event
+        v.whodunnit = legacy_version.whodunnit
+        v.object = legacy_version.object
+        v.object_changes = legacy_version.object_changes
+        v.created_at = legacy_version.created_at
+        v.save!
```
yachtcaptain23: If there's really 6 million rows, then 100-1000 saves per second would take over an hour or even half a day. Would you refactor to do batch inserts? IIRC there's no clean way in ActiveRecord to do so.

corymcdonald: When I ran it locally it took almost 3 hours :shipit:

corymcdonald:
The updated version only took 33 minutes! ✨⌛️✨

corymcdonald:
Blocking until CRM 1.1 is finsihed

PR #1495, opened 2019-01-03
Foundational PR for new UI

dlipeles:
## Foundational PR for new UI

This PR brings a new structure to our app for future front end implementations, as well as brings a number of new technologies to Publishers

## Structure
New UI will live in the `app/javascript` directory, specifically in two locations

- `app/javascript/packs/*` will contain boiler-plate code that will be mounted to a view, these files are minimal and serve as a bridge between react and rails.

- `app/javascript/views/*` will contain the actual markup, styles, and functionality of our UI, this directory will be structured using the `Fractal` component system. It goes something like this

--`views`
----`component1`
--------`subcomponent1`
----------------`index.tsx`
----------------`style.ts`
----------------`spec.ts`
--------`subcomponent2`
----------------`index.tsx`
----------------`style.ts`
----------------`spec.ts`
--------`index.tsx` 
--------`style.ts` 
--------`spec.ts` 

----`component2` 
...etc
----`component3` 
...etc

## Technologies 

Bringing a number of new technologies to the table

#### TypeScript 
Superset of javascript, as its name implies this will let us add Type-ing to our code which can improve predictability and testing, as well as prevent some errors. It also can also allow the use of intellisense in your IDE. 

#### Styled Components 
Popular component focused styling system known also known as `css-in-js`. Reduces the amount of markup to apply styles, allows for easy dynamic prop-based styling

#### Jest
Test kit for React, requires minimal configuration, the `spec.ts` files will be devoted to testing. 

#### Webpack 4.0 
Faster and will have more support in the future, good idea to get a jump on the latest version before starting.

#### Access
I've created an e-mail whitelist for access to the new UI, that can help us push to GitHub without having users access it before it's ready. The env variable is `new_ui_email_whitelist`

#### Integration
The new UI calls for the following components, this could use some discussion but as I see it, we can organize them this way: 

-------------------- Not Necessary for CRM 1.1
-- Dashboard
-- Channels
-------------------- Necessary for CRM 1.1
-- Navbar
-- Referrals
-- Payments

#### So I believe the plan would be implement Referrals, Payments, and Navbar first, then remove the e-mail whitelist, we can update the Dashboard and Channels page once CRM 1.1 is complete. 

`Note on the navbar, we will temporarily need to include the Add Channel and Tipping Banner buttons until the new channels page is completed.`


app/javascript/views/referrals/referralsCard/style.jsx
```diff
@@ -0,0 +1,162 @@
...
+
+export const StyledWrapper = styled.div
+`
+  border-radius: 6px;
+  background-color: white;
+  box-shadow: rgba(99, 105, 110, 0.18) 0px 1px 12px 0px;
+  padding: 30px;
+`;
+
+export const StyledGrid = styled.div
```
corymcdonald: Do we have to name all of these `StyledX` I feel like we would know that it's styled based on the fact that we imported it from Styled?

dlipeles: I took this convention from Brave-ui, not sure if they had a reason for this, but I'd be okay with dropping the Styled prefix

dlipeles: I think it may be to prevent naming clash with Components 

for example if you have a component `<Card/>` and want to create a styled class called `Card` you may get an error

Although that's a lot of extra syntax for this benefit, I'm leaning towards removing the `Styled` prefix

app/javascript/views/referrals/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+export default class Referrals extends React.Component {
+
+  render () {
+    return (
+      <Wrapper>
+        <ReferralsNav/>
+        <Container>
+          <ReferralsHeader/>
+          <Grid>
+            <ReferralsCard/>
```
corymcdonald: Can we name this component `ReferralCard` as opposed to `Referrals`? Each of these items is one Referral so I think that would make sense.

corymcdonald:
I'm seeing this error in my console on the Referrals Page
```
Warning: Received `true` for a non-boolean attribute `title`.
```

app/javascript/views/referrals/style.ts
```diff
@@ -0,0 +1,29 @@
+import styled from 'styled-components'
+
+export const Wrapper = styled.div
+`
+  width: 100%;
+`
+
+export const Container = styled.div
+`
+  width: 1200px;
```
corymcdonald: I don't think we should be setting this value. I think instead we should probably set a padding on the left and right. Similar to [this article on the matter](https://css-tricks.com/best-way-implement-wrapper-css/). 

I was also looking at how we implement the wrapping on the grid. If we remove this style and the `min-width` in the `main-content.scss` then we can apply the following style to the grid 
```ts
export const Grid = styled.div
`
  display: grid;
  grid-template-columns: repeat(auto-fit, minmax(22rem, auto));
  grid-gap: 1rem;
  margin-top: 1rem;
`
```

Which will allow our tiles to be responsive and go to the next row when there is not enough space anymore.

app/javascript/views/referrals/referralsNav/style.ts
```diff
@@ -0,0 +1,57 @@
...
+      margin-bottom: auto;
+      font-size: 28px;
+      color: white;
+      padding-top: 4px;
+      padding-left: 7px;
+    `
+  }
+`
+
+export const Button = styled.div
```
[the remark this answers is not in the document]
corymcdonald: discussed online, brave-ui does not have a white background button

app/javascript/views/referrals/style.ts
dlipeles: This is a good solution 👍

app/javascript/views/referrals/referralsHeader/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+            <Text header>{locale.downloads}</Text>
+            <Text stat>9999</Text>
+          </TextWrapper>
+          <TextWrapper>
+            <Text header>{locale.installs}</Text>
+            <Text stat>999</Text>
+          </TextWrapper>
+          <TextWrapper>
+            <Text header>{locale.thirtyDay}</Text>
+            <Text stat purple>999</Text>
```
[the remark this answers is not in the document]
dlipeles: Lets table it for now, when we have a more concrete color palette we can refactor and figure out which will be primary, secondary, ternary etc.

app/javascript/views/referrals/index.tsx
dlipeles: Rethought this, is it okay if we keep as Referrals? Each card will be representing multiple codes, and I also like the consistency of `ReferralsX` components in the `Referrals` section

package.json
```diff
@@ -45,34 +45,44 @@
     "preset": "ts-jest/presets/js-with-babel"
   },
   "dependencies": {
-    "@rails/webpacker": "3.5",
+    "@babel/preset-react": "^7.0.0",
+    "babel-core": "7.0.0-bridge.0",
```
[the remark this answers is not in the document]
dlipeles: I think this might explain https://github.com/babel/babel-bridge/issues/5#issuecomment-406379073

app/javascript/packs/Referrals.jsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import * as ReactDOM from 'react-dom'
+import Referrals from '../views/referrals/index.tsx'
+
+document.addEventListener('DOMContentLoaded', () => {
+  ReactDOM.render(
+    <Referrals name="React" />,
```
corymcdonald: Do we need this prop?

PR #1511, opened 2019-01-08
Ensure emails are sent to publishers with only contributions but no wallet address

### LINKED ISSUE #1510: Publishers who have only contributions waiting and are not Uphold authorized are not being notified
We email publishers who have a balance waiting but who won't receive payment because we aren't authorized to send them money on their account.  Currently, only publishers with referral balances are being emailed.

nvonpentz:
Fix #1510 
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/services/payout_report_publisher_includer.rb
+++ b/app/services/payout_report_publisher_includer.rb
@@ -29,11 +29,9 @@ class PayoutReportPublisherIncluder < BaseService
 
     @publisher.channels.verified.each do |channel|
       probi = wallet.channel_balances[channel.details.channel_identifier].probi # probi = balance - fee
+      publisher_has_unsettled_balance = probi.positive? ? true : publisher_has_unsettled_balance
       next unless probi.positive? && @publisher.uphold_verified? && wallet.address.present?
-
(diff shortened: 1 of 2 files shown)


app/services/payout_report_publisher_includer.rb
```diff
@@ -29,11 +29,9 @@ def perform
 
     @publisher.channels.verified.each do |channel|
       probi = wallet.channel_balances[channel.details.channel_identifier].probi # probi = balance - fee
```
yachtcaptain23: IMO the readability could be better where the first part is clearly the code looking through the publisher's balance and then looking at all of the publisher's channel balances.

Otherwise, it's good :+1:

nvonpentz: I agree.  This is because the owner/referral balance in the `channel_balances` hash in the wallet.

We rename the `channel_balances` array in wallet to `account_balances` in https://github.com/brave-intl/publishers/pull/1438.

Issue #1535, opened 2019-01-15
Publisher and Promo emails need updated


rossmoody:
Noticed all the emails I've ever gotten on any computer or client from Publishers ends up in spam. Also noticed the formatting in many instances breaks very badly. Additionally noticed that we have some dated copy (like saying publishers will have a verified icon in the toolbar which is Muon behavior only) so on the whole the emails just need updated to be current and consistent in their application.
<img width="730" alt="screen shot 2019-01-19 at 11 54 51 am" src="https://user-images.githubusercontent.com/29072694/51431710-9309df00-1be1-11e9-8b36-2f5c2ebc8a21.png">

PR #1536, opened 2019-01-15
Update Publisher and Promo Emails

rossmoody:
Previews for emails were hooked back up and all the changes in the PR can be previewed here: https://localhost:3000/rails/mailers

Fixes: https://github.com/brave-intl/publishers/issues/1535

## Updates
- consolidated typography sizing and weights. we had a pretty dramatic amount of sizes and instances of bold text
- limited the use of imagery so the emails load faster and format more consistently. it's important we aren't asking the user to download 8+ images per email. this will likely significantly help with spam filter flagging and formatting for gmail that auto blocks images into big blank boxes
- there is a small disconnect between users getting emails from Brave Rewards while being on a basic attention token website and then getting a brave logo in the header but being served bat links/logos in the footer while also occasionally getting support from a @basicattentiontoken address. it's important there is some semblance of addressing why this is happening and so the footer communication was adjusted
- emails have quite a bit of padding and default to 600px wide which is quite large for a standard min-width. especially with the complexity of content we are showing being really minimal. the approach was shifted to design for mobile and scaled up instead of the reverse.
- the red color isn't associated with Brave, BAT or Rewards branding and it was giving the button a "danger" intent feeling. color was pushed into a more blurple brand primary that we are pretty consistently associating with Rewards.
- omitted the signoff from the Brave team. it felt odd in many scenarios to have a "Sincerely" signoff when talking about security concerns and/or possible fraudulent activity. it came off like "You might be getting hacked. Sincerely, Brave."
- revised the subject lines in a few extremely key areas. the use of exclamation points in subject lines is an immediate red flag for spam so they were omitted everywhere. any marketing-speak that might trip a spam filter was reconsidered. this includes words like "limited time offer, free, hurry, quick, promo"
- adjusted subject headlines and button text to be sentence case instead of title case since there was a mix
- adjusted the verbiage to Creators instead of publishers since that is the direction the name is headed
- opted for explicit channel and email declarations in subject headlines and made them flow like sentences
- standardized the way we warn people about private access link usage. we had different wording, placements and application
- opted for buttons in place of bare private access url hyperlinks that were intimidating and unpolished
- added a more accessible snippet to each email including a support email for reaching out for help. this is really important to feeling accessible for help
- changed the footer social links to include brave and bat websites. also omitted medium since it hasn't been posted to in 8 months.

beforeafter


app/mailers/application_mailer.rb
```diff
@@ -21,25 +21,13 @@ def add_image(image_path)
...
-    add_image("mailer/logo-rocketchat.png")
-    add_image("mailer/logo-twitter.png")
-    add_image("mailer/footer-top-pattern.png")
-    add_image("mailer/logo-bat.png")
+    add_image("mailer/social_bat.png")
+    add_image("mailer/social_reddit.png")
+    add_image("mailer/social_brave.png")
+    add_image("mailer/social_twitter.png")
 
     if self.class == PromoMailer
```
corymcdonald: Let's refactor this to be one line like si
```ruby
raise "Invalid mailer class; can't run #add_images." unless self.class == InternalMailer || self.class == PromoMailer || self.class == PublisherMailer
```

app/views/promo_mailer/promo_activated_2018q1_verified.html.slim
```diff
@@ -32,8 +32,4 @@ hr style="border-top-width: 0px"
 
 p.promo--clear = t("promo.shared.tos_1")
 
-br= t("publisher_mailer.shared.signature_html")
```
corymcdonald: Can you take a look at the `switch_banner_1` here? It looks like it was removed so now it's causing this to encounter an error

config/locales/en.yml
```diff
@@ -170,10 +173,11 @@ en:
...
     mailer:
-      medium: Medium
-      twitter: Twitter
-      reddit: Reddit
-      rocketchat: Rocket.Chat
+      brave: Brave Website
+      twitter: Basic Attention Token Twitter
+      reddit: Basic Attention Token Reddit
+      bat: Basic Attention Token
+      footer_html: Brave Rewards is made by the fine folks at <a href="https://brave.com">Brave Browser</a> and powered by the <a href="https://www.basicattentiontoken.org">Basic Attention Token.</a><br />View the <a href="https://basicattentiontoken.org/publisher-terms-of-service/">Terms of Service.</a>
```
corymcdonald: I think we're called Brave Software?

app/views/promo_mailer/promo_activated_2018q1_verified.html.slim
corymcdonald: Adressed in this commit https://github.com/brave-intl/publishers/pull/1536/commits/c6db85bd227b5c29f9bbb3b45e4027d0c7635713

nvonpentz:
This stuff is looking really great.  Thanks for your work on this @rossmoody and @corymcdonald.

It looks like we're not getting the translations for the promo activated email for publishers with verified accounts.
image

It also looks like we won't have the share logos for twitter and attached since we've removed the logic that attaches them only when they're required [here](https://github.com/brave-intl/publishers/pull/1536/files#diff-7f4dc276bbcf65137ee281cff4ba324aL34).

rossmoody:
@nvonpentz this is worth another look. fixed up the promo mailers and revised the new `uphold_member_restricted` email html to match this styling.

dlipeles:
Looking great @rossmoody , the consistency was really needed, thanks so much for your help

Promo e-mail looks to be working now, as well as the social media share links
screen shot 2019-01-21 at 9 29 14 pm

dlipeles:
Code looks good to me, only feedback, do we still need 

https://github.com/brave-intl/publishers/blob/d0137c8e4f4c3ef662f3098e7bf34c391bb1cca1/app/mailers/application_mailer.rb#L38-L52

in `application_mailer`?

jenn-rhim:
@rossmoody Great work.  We need one more tweak. Maybe you can help us with that. The new branding stretched out of the content boundary on the mobile version of the email. Please take a look:

image

PR #1537, opened 2019-01-16
Ask users to go to uphold.com to fully verify. Closes #1513

### LINKED ISSUE #1513: remind publishers to KYC fully on Uphold upon login
when logging in a publisher who has linked their Uphold account but has not completed KYC should be reminded to complete KYC.

this can be determined by looking at the `isMember` field of the `wallet` object of the `/v1/owners/{owner}/wallet` endpoint, which returns `true` if KYC is complete

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/helpers/publishers_helper.rb
+++ b/app/helpers/publishers_helper.rb
@@ -159,7 +159,9 @@ module PublishersHelper
 
   def uphold_status_class(publisher)
     case publisher.uphold_status
-    when :verified
+    when :verified, Publisher::UpholdAccountState::BLOCKED
+      # (Albert Wang): We notify Brave when we detect a login of someone with a blocked
+      # Uphold account
       'uphold-complete'
     when :code_acquired, :access_parameters_acquired
       'uphold-processing'
@@ -169,8 +171,6 @@ module PublishersHelper
       'uphold-incomplete'
     when Publisher::UpholdAccountState::RESTRICTED
       'uphold-' + Publisher::UpholdAccountState::RESTRICTED.to_s
-    when Publisher::UpholdAccountState::BLOCKED
-      'uphold-' + Publisher::UpholdAccountState::BLOCKED.to_s
     else
       'uphold-unconnected'
     end
--- a/app/helpers/publishers_helper.rb
+++ b/app/helpers/publishers_helper.rb
@@ -167,8 +167,6 @@ module PublishersHelper
       'uphold-processing'
     when :reauthorization_needed
       'uphold-reauthorization-needed'
-    when :incomplete
-      'uphold-incomplete'
     when Publisher::UpholdAccountState::RESTRICTED
       'uphold-' + Publisher::UpholdAccountState::RESTRICTED.to_s
     else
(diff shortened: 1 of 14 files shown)


app/helpers/publishers_helper.rb
```diff
@@ -138,6 +138,8 @@ def uphold_authorization_description(publisher)
     case publisher.uphold_status
     when :unconnected
       I18n.t("helpers.publisher.uphold_authorization_description.connect_to_uphold")
+    when Publisher::UpholdAccountState::RESTRICTED
+      publisher.wallet.is_a_member? ? "Visit Uphold's Support page" : "Go to uphold.com"
```
corymcdonald: Let's internationalize this

app/models/publisher.rb
```diff
@@ -249,11 +262,18 @@ def uphold_status
...
+    elsif self&.wallet&.uphold_account_status == UpholdAccountState::RESTRICTED || (self.wallet.present? && self.uphold_verified? && self.wallet.not_a_member?)
+      UpholdAccountState::RESTRICTED
+    elsif self&.wallet&.uphold_account_status == UpholdAccountState::BLOCKED
+      UpholdAccountState::BLOCKED
     else
       :unconnected
     end
   end
 
+  def not_kycd_by_uphold?
```
corymcdonald: Would we be able to use the wallet `is_a_member?` check?

app/models/publisher.rb
```diff
@@ -255,6 +268,12 @@ def uphold_status
       :access_parameters_acquired
     elsif self.uphold_code.present?
       :code_acquired
+    elsif self&.wallet&.uphold_account_status == UpholdAccountState::RESTRICTED || (self.wallet.present? && self.uphold_verified? && self.wallet.not_a_member?)
+      UpholdAccountState::RESTRICTED
```
corymcdonald: I'm not getting into this state because the incomplete is returning true first. I don't think we need the incomplete if we're going to be displaying this during restricted status

test/services/publisher_wallet_getter_test.rb
```diff
@@ -19,156 +19,196 @@ class PublisherWalletGetterTest < ActiveJob::TestCase
...
+          eyeshade_response[:status][:action] = action
+          stub_request(:get, %r{v1/owners/#{URI.escape(publisher.owner_identifier)}/wallet}).
+            with(headers: {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'User-Agent'=>'Faraday v0.9.2'}).
+            to_return(status: 200, body: eyeshade_response.to_json, headers: {})
+          PublisherWalletGetter.new(publisher: publisher).perform
+          assert_equal Publisher::UpholdAccountState::REAUTHORIZATION_NEEDED, publisher.uphold_status
+        end
+      end
+
+      test "blocked by uphold" do
```
corymcdonald: Should we take this out or flush out this test?

app/views/publishers/home.html.slim
```diff
@@ -108,6 +108,9 @@ script type="text/html" id="confirm_default_currency_modal_wrapper"
             .dashboard-panel--content#uphold_connect
               .status-description
                 = uphold_status_description(current_publisher).html_safe
+              = link_to(uphold_authorization_description(current_publisher),
+                        "https://uphold.com/dashboard",
```
corymcdonald: I think we should use the `uphold_dashboard_url` from the secrets file so in dev we can click it and go to the staging uphold site

corymcdonald:
screen shot 2019-01-25 at 11 47 58 am

Due to the weird CSS stuff when you initially connect and are restricted then you get a "reconnect to uphold" status

PR #1538, opened 2019-01-16
Expose Uphold Status and ID

corymcdonald:
Changed the following

* We can actually search by a UUID now
* We can search for identical Uphold accounts
* Rewrote the `payout_report_publisher_includer_test` to handle more permutations of statuses
* Sorting by Last Signed In will show the null values last so we can see who was the most recent signed in user

### DIFF
--- a/app/controllers/admin/publishers_controller.rb
+++ b/app/controllers/admin/publishers_controller.rb
@@ -96,6 +96,7 @@ class Admin::PublishersController < AdminController
   end
 
   def is_a_uuid?(uuid)
+    # https://stackoverflow.com/questions/47508829/validate-uuid-string-in-ruby-rails
     uuid_regex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/
     uuid_regex.match?(uuid.to_s.downcase)
   end
--- a/app/controllers/admin/publishers_controller.rb
+++ b/app/controllers/admin/publishers_controller.rb
@@ -7,12 +7,14 @@ class Admin::PublishersController < AdminController
     @publishers = if sort_column&.to_sym&.in? Publisher::ADVANCED_SORTABLE_COLUMNS
                     Publisher.advanced_sort(sort_column.to_sym, sort_direction)
                   else
-                    Publisher.order(sanitize_sql_for_order("#{sort_column} #{sort_direction}"))
+                    Publisher.order(sanitize_sql_for_order("#{sort_column} #{sort_direction} NULLS LAST"))
                   end
 
     if params[:q].present?
       # Returns an ActiveRecord::Relation of publishers for pagination
-      search_query = "%#{remove_prefix_if_necessary(params[:q])}%"
+      search_query = remove_prefix_if_necessary(params[:q])
+      search_query = "%#{search_query}%" unless is_a_uuid?(search_query)
+
       @publishers = @publishers.where(search_sql, search_query: search_query)
     end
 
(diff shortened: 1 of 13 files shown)


app/views/admin/publishers/show.html.slim
```diff
@@ -18,11 +18,19 @@
...
-          .db-field = "Uphold verified:"
+          .db-field = "Uphold once connected:"
           .db-value = @publisher.uphold_verified?
         .db-info-row
           .db-field = "Uphold authorized:"
           .db-value = @publisher.wallet.authorized?
+        - if @publisher.wallet.status.present?
+          .db-info-row
+            .db-field = "Uphold status:"
+            .db-value = @publisher.wallet.status
```
yachtcaptain23: IIRC this isn't necessarily the status we would want. I was getting `re-authorize` for my account, so I believe this comes from Eyeshade/Ledger and not from Uphold.

config/locales/en.yml
```diff
@@ -293,6 +293,15 @@ en:
         </p>
         <p>Your earnings will be paid out in the next payment cycle after you verify your identity with Uphold. In the meantime, your earnings will still be held safely in your account. We’re sorry for the inconvenience and delays this may cause. Thank you for being a Brave Creator and bearing with us as we improve the integrity of the payout system. Learn more about this new requirement <a href="https://support.uphold.com/hc/en-us/articles/202887565-What-are-the-benefits-of-becoming-a-verified-member-of-Uphold-">here</a>.</p>
         <p>Thank you,</p>
+    uphold_member_restricted:
+      subject: Please check Uphold Accoun
```
yachtcaptain23: Spelling

config/locales/en.yml
```diff
@@ -293,6 +293,15 @@ en:
...
+      body_html: |
+        <p>Thank you for being a Brave creator.</p>
+        <p>
+          We've noticed that your Uphold account is not able to receive funds. Plus check your Uphold account status and ensure that there are no actionable items and that you are within the transaction limit guidelines.
```
yachtcaptain23: Spelling

lib/eyeshade/wallet.rb
```diff
@@ -25,6 +27,8 @@ def initialize(wallet_json:, channel_json:)
       @possible_currencies = details_json["possibleCurrencies"] || []
       @address = details_json["address"] || ""
       @is_member = details_json["isMember"] || false
+      @status  = details_json["status"]
+      @uphold_id = details_json["id"]
```
yachtcaptain23: Do we know if it's `id` or `address`? @evq would you verify this?

evq: it's `id`. `address` is the card address of the first card that matches the user's preferred currency

config/locales/en.yml
corymcdonald: https://github.com/brave-intl/publishers/pull/1538/commits/e3cadf7bf81d81eb370b561dacc2d28ea59069b0

config/locales/en.yml
corymcdonald: https://github.com/brave-intl/publishers/pull/1538/commits/e3cadf7bf81d81eb370b561dacc2d28ea59069b0

app/controllers/admin/publishers_controller.rb
```diff
@@ -92,4 +94,9 @@ def update_params
   def sortable_columns
     [:last_sign_in_at, :created_at, Publisher::VERIFIED_CHANNEL_COUNT]
   end
+
+  def is_a_uuid?(uuid)
+    uuid_regex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/
```
nvonpentz: if you didn't write this, we should probably credit where it came from in a comment

corymcdonald: Ah good point - added the reference to the original location
https://github.com/brave-intl/publishers/pull/1538/commits/7deafa87df6a7870871f1f2ec588999d5c85e6b9

PR #1546, opened 2019-01-22
Do not destroy dependent promo registrations for a publisher

nvonpentz:
We had some reports of some publishers who lost all their promo stats after deleting a channel with no referrals credited to it.  When they added the channel back, their channels stats were reinstated.

Stepping through the flow I found that after a channel was destroyed on this line

https://github.com/brave-intl/publishers/blob/3069160c7e65efe81a406b72cf202e620e24bbcd/app/jobs/delete_publisher_channel_job.rb#L19

the total number of promo registrations in the database decreased by more than one.  It seems like this call was not only destroying the promo registration for that channel, but for all channels the publisher had.  When they re-added their channels, the Promo::PublisherChannelsRegistrar would recreate all the old promo registrations.

It looks like the `has_many :promo_registrations, dependent: :destroy` on the Publisher model is doing more than just deleting all promo registrations that belong to a publisher when the publisher is deleted. 
It's also deleting all the promo registrations that belong to a publisher when one of the publisher's channels are destroyed.

I've removed the `dependent: destroy` because this serves no function that I can see at the moment.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -24,7 +24,7 @@ class Publisher < ApplicationRecord
   has_many :login_activities
 
   has_many :channels, validate: true, autosave: true
-  has_many :promo_registrations, dependent: :destroy
+  has_many :promo_registrations
   has_many :site_banners
   has_many :site_channel_details, through: :channels, source: :details, source_type: 'SiteChannelDetails'
   has_many :youtube_channel_details, through: :channels, source: :details, source_type: 'YoutubeChannelDetails'


corymcdonald:
I have a couple of thoughts -

>the total number of promo registrations in the database decreased by more than one.

Did you have more than one channel that was going to be deleted? could that be why the promo registrations decreased by more than one?

i'm not sure if the `delete_publisher_channel_job.rb` is the cause of the promo registrations being deleted. we destroy the promo registration when the channel gets deleted here

https://github.com/brave-intl/publishers/blob/staging/app/models/channel.rb#L32

So I think that we should remove that instead of the publishers line and see if that fixes it

>It's also deleting all the promo registrations that belong to a publisher when one of the publisher's channels are destroyed.

Does this mean the promo registrations that are associated to other channels? If so that is very worrisome because the `dependent: destroy` should only get called when we call destroy on the publisher. 


My other concern is if we do intend not to delete the promo registrations do they get cleaned up on the promo registration server? Do we currently have promo codes that are orphaned?

nvonpentz:
Thanks for taking a look @corymcdonald.

Channels are deleted via `channels#destroy`, which calls the DeletePublisherChannelJob.

When I was reproducing this error I had two verified channels with promo registrations.  I put a debugger on line 17 of the job.  I removed one channel and on line 19, after the ` success = @channel.destroy! `, two channels had been removed, but the job was only called once.

You are correct that promo registrations are actually deleted as part of the callback on the channel model.  We don't want to remove this though, because we do want promo registrations to be deleted when a channel is deleted, otherwise they cannot be reregistered.  I did test the effects of removing this earlier, and it is what you'd expect: The channel was deleted but the promo registration was not.

> Does this mean the promo registrations that are associated to other channels

Yes, the other channels that are owned by this publisher.  When they re-add their deleted channel (or any other channel for that matter, the promo registrations will be recreated.  That's why there wasn't an obvious problem initially for the pubs that reported it.

> My other concern is if we do intend not to delete the promo registrations do they get cleaned up on the promo registration server?

We do intend to delete the registrations on publishers, but we persist them on the promo server.  When a channel is deleted, the promo server is updated that the current owner is `null`.  It's updated again when a publisher verifies that channel again.

The modifications in this PR do resolve the problem: only one promo registration is deleted when the channel is destroyed, though why exactly the `has_many :promo_registrations, dependent: :destroy` on the publisher model is causing this isn't entirely clear.

corymcdonald:
I'm thinking that the reason is because each promo_registration has a `publisher_id`. This is kind of strange because a `channel` is associated to a publisher. I'm going to test this locally and report back with my findings 👍

dlipeles:
This might be relevant too 

from stackoverflow:

```
delete will only delete current object record from db but not its associated children records from db.

destroy will delete current object record from db and also its associated children record from db.

Their use really matters:

If your multiple parent objects share common children objects, then calling destroy on specific parent object will delete children objects which are shared among other multiple parents.
```

I noticed I had to remove that `dependent: :destroy line as well when I was developing the referrals page. I wasn't able to delete promo_registrations with that code in.

nvonpentz:
Albert noticed this which probably shouldn't be there and might explain the weird behavior

https://github.com/brave-intl/publishers/blob/3e4c1b28b19ec9be3e8dc90f0436e0e7fccab41c/app/models/promo_registration.rb#L33

corymcdonald:
I think it would make more sense to remove that in the registration rather than modify publishers. Does making that change sound good @nvonpentz?

nvonpentz:
@corymcdonald Sounds good! (assuming it works).

nvonpentz:
@yachtcaptain23 can this be closed since https://github.com/brave-intl/publishers/pull/1591 was merged?

corymcdonald:
Seems like it! Going to close this out 😄

PR #1554, opened 2019-01-24
Preview emails rebase

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


rossmoody:
@dlipeles Just pushed a missing `table` declaration on btn container. Should center now. Also, I noticed on my preview renders that Outlook is still borked but it was borked before so I don't think we should block on it. Should probably address it in a follow up though with something this like this: https://templates.mailchimp.com/development/css/outlook-conditional-css/

dlipeles:
Sounds good @rossmoody I'll take another look later today

dlipeles:
Followup item - #1556

PR #1558, opened 2019-01-25
Refactor: Update BraveUI Icon Module

rossmoody:
This update should be merged with the Brave UI version update to `brave-ui@0.37.0`. The `VerifiedFillIcon` was changed to just `VerifiedIcon`. https://github.com/brave/brave-ui/pull/354


rossmoody:
Gonna close this one for now. With the amount of changes going into Brave-UI in the near future, might make more sense to do it then. Or since a bunch of icons just went in, if we want to leverage them in the Payments UI, could merge with that update.

PR #1567, opened 2019-01-31
CSS

rossmoody:
I was noodling around. Might be some CSS updates in here worth considering. I was just inspecting my way through screens so wanted to pass it along. Looks good. Hope to get Brave-UI in a place to leverage a little more in the future.


corymcdonald:
Thanks! I'll incorporate these changes on my branch

PR #1574, opened 2019-02-02
Reports and invoices

corymcdonald:
# Things left to do 

- [x] submit security review for new network requests
- [x] restrict access to the invoices controller down to new ui whitelist
- [x] figure out the z-index issue with the overview page modal
- [x] rails tests

## Overview page
screen shot 2019-02-01 at 7 14 13 pm

## Reports Page
screen shot 2019-02-01 at 7 20 47 pm
screen shot 2019-02-01 at 7 21 53 pm

# Invoices Page

screen shot 2019-02-01 at 7 20 13 pm


db/migrate/20190131170022_create_reports.rb
```diff
@@ -0,0 +1,17 @@
+class CreateReports < ActiveRecord::Migration[5.2]
+  def change
+    create_table :reports, id: :uuid, default: -> { "uuid_generate_v4()" } do |t|
+      t.uuid :partner_id
+      t.references :uploaded_by, index: true, foreign_key: { to_table: :publishers }, type: :uuid
+
+      t.string :amount_bat
+      t.datetime :created_at, null: false
```
yachtcaptain23: I generally prefer `t.timestamps`, but if this works with auto-update, then don't change.

app/models/report.rb
```diff
@@ -0,0 +1,18 @@
+class Report < ActiveRecord::Base
+  belongs_to :partner
+  belongs_to :uploaded_by, class_name: "Publisher"
```
yachtcaptain23: I believe it's `Partner` rather than Publisher.

app/models/invoice.rb
```diff
@@ -0,0 +1,16 @@
+class Invoice < ActiveRecord::Base
+  belongs_to :partner
+  belongs_to :uploaded_by, class_name: "Publisher"
```
yachtcaptain23: I believe it's `Partner` rather than Publisher.

app/javascript/packs/Reports.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+document.addEventListener("DOMContentLoaded", () => {
+  const reactElement = document.createElement("div");
+  reactElement.id = "react-content";
+  document.getElementById("main-content").appendChild(reactElement);
+
+  const data = JSON.parse(
+    document.getElementById("reactData").getAttribute("data-react")
+  );
+  console.log(data);
```
yachtcaptain23: Debug statement.

app/models/report.rb
corymcdonald: Ah, so both partners and admins will upload reports so I made this publisher since it refers to both

app/controllers/partners/invoices_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+      invoice.save
+    end
+
+    private
+
+    #Internal: only allow users who are on the new UI whitelist to be allowed to access controller
+    #
+    # Returns nil
+    def filter_users
+      raise unless current_user&.in_new_ui_whitelist?
```
yachtcaptain23: Would you get rid of the `&`? Otherwise a `nil` `current_user` would bypass

app/models/publisher.rb
```diff
@@ -346,6 +346,8 @@ def can_create_uphold_cards?
 
   # Remove when new dashboard is finished
   def in_new_ui_whitelist?
+    # for testing purposes :)
+    return true if self.email == "partner@brave.com" 
     self.email.in?((Rails.application.secrets[:new_ui_email_whitelist] || "").split(","))
```
diracdeltas: can `self.email` be modified client-side? ex: can i run client-side javascript that will set self.email to `partner@brave.com` ?

app/javascript/components/Modal.tsx
```diff
@@ -0,0 +1,111 @@
+import * as React from "react";
```
dlipeles: Lets put this and style in their own folder ie. 

`app/javascript/components/modal/Modal.tsx`

db/migrate/20190131170022_create_reports.rb
```diff
@@ -0,0 +1,16 @@
+class CreateReports < ActiveRecord::Migration[5.2]
+  def change
+    create_table :reports, id: :uuid, default: -> { "uuid_generate_v4()" } do |t|
+      t.uuid :partner_id
+      t.references :uploaded_by, index: true, foreign_key: { to_table: :publishers }, type: :uuid
+
+      t.string :amount_bat
```
yachtcaptain23: IMO amounts being discussed and paid out should be a separate table. We can also track the state in the separate model too. 

Would you rename this column to `partner_requested_amount_in_bat`? Additionally, create another column called `admin_paid_out_in_bat`.

yachtcaptain23: Actually now that I think about it, this info should be stored in `invoices`. A `Report` will always have an `Invoice`, but the `Invoice` could or could not contain the `partner_requested_amount_in_bat`. The `revised` state of the `Invoice` will contain the admin amount.

app/controllers/partners/invoices_controller.rb
```diff
@@ -0,0 +1,36 @@
+module Partners
+  class InvoicesController < ApplicationController
+    before_action :filter_users
+
+    def index
```
dlipeles: Consider aggregating invoices across partners in the org 🔶

dlipeles: And reports too 🙈

corymcdonald:
Going to close this based on a new approach we're taking

PR #1578, opened 2019-02-04
Make BAT default currency

dlipeles:
Closes #1578 

Goal is to pre-select BAT when prompting for deposit currency

Caveats:

- Some publishers will not have the BAT card depending on their jurisdiction
- Should still pre-select their default currency if they've already chosen one. 

screen shot 2019-02-04 at 1 49 05 pm


app/javascript/publishers/home.js
```diff
@@ -97,20 +97,16 @@ function getPossibleCurrencies() {
...
-
   possibleCurrencies.forEach(currency => {
     let option = document.createElement('option');
     option.value = currency;
     option.innerHTML = currency;
-    option.selected = (currency === selectedCurrency);
+    if((!selectedCurrency || selectedCurrency.length === 0) && currency === DEFAULT_CURRENCY){
+      option.selected = true;
+    }
+    else{
```
corymcdonald: nit: space after this

PR #1588, opened 2019-02-05
Feature/top balance totals

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/admin/stats/top_balances_controller.rb
+++ b/app/controllers/admin/stats/top_balances_controller.rb
@@ -1,0 +1,12 @@
+class Admin::Stats::TopBalancesController < AdminController
+  def index
+    @result = case params[:type]
+    when Eyeshade::TopBalances::CHANNEL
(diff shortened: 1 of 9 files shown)


app/controllers/admin/stats/top_balances_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Admin::Stats::TopBalancesController < AdminController
+  def index
+    @result = case params[:type]
```
corymcdonald: Since there's not a default when you click the link on the navbar it results in an exception

app/views/admin/stats/top_balances/index.html.slim
```diff
@@ -0,0 +1,17 @@
+div.panel-body
+  div.adv-table
+    table.display.table.table-bordered.table-striped.dynamic-table id="dynamic-table"
+      tr
+        th Account Type
```
corymcdonald: Might make more sense to swap Account Type and Account ID. For example, against staging data

screen shot 2019-02-05 at 6 20 56 pm

PR #1589, opened 2019-02-05
Campaigns and Referrals

dlipeles:
### Todo
- [x] Refactor to 2 separate pages instead of one nested view, use our new modals
- [x] Port views to `/partners`
- [x] Refresh after modal action
- [x] Add description to registrations
- [x] Add logic to hardcoded values
- [x] Flesh out all modals
- [x] How will we calculate estimated earnings - (Discussed with mandar, out of scope for this release)
- [x] Clean up controllers I
- [x] Navbar
- [x] Edit Campaign Name
- [x] CSS touchups
- [x] Smooth out Promo Registration and Stats fetching
- [x] UX
- [x] Sanitize Inputs
- [x] Aggregate data across partners in organization
- [x] Max number of codes 
- [x] Add locales where needed 
- [x] Responsiveness
- [x] Clean up JS 
- [x] Unit Tests
- [x] Handle errors more gracefully

### Overview
screen shot 2019-02-05 at 6 27 10 pm

### Select View
screen shot 2019-02-05 at 6 34 36 pm


app/services/promo/registrations_stats_fetcher.rb
```diff
@@ -8,6 +8,7 @@ def initialize(promo_registrations:)
   end
 
   def perform
+    puts 'helo'
```
corymcdonald: hi 👋

config/secrets.yml
```diff
@@ -29,7 +29,7 @@ default: &default
   basic_auth_user: <%= ENV["BASIC_AUTH_USER"] %>
   fee_rate: <%= ENV["FEE_RATE"] %>
   internal_email: <%= ENV["INTERNAL_EMAIL"] %>
-  new_ui_email_whitelist: <%= ENV["NEW_UI_EMAIL_WHITELIST"] %>
+  new_ui_email_whitelist: user@test.com
```
corymcdonald: pls revert

app/controllers/partners/referrals_controller.rb
```diff
@@ -0,0 +1,61 @@
+module Partners
```
corymcdonald: This file is using 4 spaces and instead of 2

app/controllers/partners/referrals_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    def data
+        data = {}
+        unassigned_codes = current_publisher.promo_registrations.where(promo_campaign_id: nil)
+        campaigns = []
+    
+        current_publisher.promo_campaigns.each do |promo_campaign|
+        promo_registrations = current_publisher.promo_registrations.where(promo_campaign_id: promo_campaign.id)
+        campaigns.push(
```
corymcdonald: Instead of building this here, can we move this data aggregation to the model, similar to how [`Report.rb`](https://github.com/brave-intl/publishers/blob/f3b4ebd5d9646f2c7172b2bad8776d022a7eca60/app/models/report.rb#L8) does it?

It cleans up the controller a bit.

app/controllers/partners/referrals_controller.rb
```diff
@@ -0,0 +1,61 @@
+module Partners
+    class ReferralsController < ApplicationController
+    include PromosHelper
+
+    def index
+        render "referrals"
+    end
+
+    def data
```
corymcdonald: Instead of this being a separate route that doesn't really conform to anything let's move it to the index. You can respond to different formats like so

```ruby
respond_to do |format|
  format.html
  format.json  do 
    render json: method_producing_data
  end 
end
```

app/controllers/partners/referrals_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+        promo_registration.update(promo_campaign_id: params[:campaign])
+        end
+    end
+    
+    def delete_codes
+        promo_registration = current_publisher.promo_registrations.find(params[:id])
+        promo_registration.destroy
+    end
+    
+    def create_campaign
```
corymcdonald: This needs to be removed

corymcdonald:
btw can you make sure prettier is runnin on your tsx code?

app/controllers/partners/referrals_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+          format.html
+          format.json  do 
+            render json: prepare_json
+          end 
+        end
+    end
+
+    private
+    def prepare_json
+        membership = Membership.find_by(user_id: current_publisher.id)
```
corymcdonald: I think you can just do this
```suggestion
        partner_ids = Membership.find_by(user_id: current_publisher.id).organization.map { |membership| membership.user_id}
```

app/controllers/partners/promo_campaigns_controller.rb
```diff
@@ -0,0 +1,27 @@
+include ActionView::Helpers::SanitizeHelper
+module Partners
+    class PromoCampaignsController < ApplicationController
+    def create
+        promo_campaign = PromoCampaign.create(
+            name: sanitize(params[:name]),
```
corymcdonald: You don't have to sanitize the input here because rails should do it by default. If you are assigning the entire params then it might make more sense to use [strong parameters](https://edgeguides.rubyonrails.org/action_controller_overview.html#strong-parameters).

PR #1601, opened 2019-02-11
Link to Help Center

### LINKED ISSUE #1600: Link to Help Center in Publisher Dropdown
Add a link to the dropdown that directs to 

https://support.brave.com/hc/en-us/

screen shot 2019-02-10 at 7 37 51 pm

dlipeles:
Closes #1600 

screen shot 2019-02-10 at 7 37 51 pm


app/views/application/_nav.html.slim
```diff
@@ -27,6 +27,7 @@ nav.navbar.navbar-default.navbar-static-top.top-nav-collapse
                       = t ".security"
                   - if show_faq_link?
                     li= link_to t(".faqs"), faqs_path
+                  li= link_to t(".help"), "https://support.brave.com/hc/en-us/"
```
corymcdonald: In the future we should sending users to the locale that they're currently on

dlipeles: Yup good idea, as of right now HC is only en-us 🇺🇸

PR #1609, opened 2019-02-13
Referral charts on main dashboard

yachtcaptain23:
image

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/javascript/publishers/dashboard_chart.js
+++ b/app/javascript/publishers/dashboard_chart.js
@@ -50,8 +50,10 @@ function createChart(data, title, suggestedMax) {
   var wrapper = document.getElementById('channel-referrals-stats-chart');
   var canvas = document.createElement('canvas');
   canvas.setAttribute("width", "400");
-  canvas.setAttribute("height", "100");
+  canvas.setAttribute("height", "300");
   wrapper.appendChild(canvas);
+
+  Chart.defaults.global.defaultFontFamily = 'Poppins';
 
   new Chart(canvas, {
     type: 'line',
@@ -80,7 +82,6 @@ function createChart(data, title, suggestedMax) {
         mode: 'x'
       },
       title: {
-        fontFamily: 'Poppins',
         fontSize: 18,
         display: true,
         text: title.toUpperCase()
--- a/app/javascript/publishers/dashboard_chart.js
+++ b/app/javascript/publishers/dashboard_chart.js
@@ -10,31 +10,43 @@ let colors = [
   '255, 159, 64'
 ]
 
-function createLabels() {
-  var startingDate = new Date(document.getElementById('all-channels-referral-stats-starting-month').value);
-  var endDate = new Date();
+function createLabels(startingDate) {
   var loop = new Date(startingDate);
-
   var dates_array = [];
 
-  while (loop <= endDate) {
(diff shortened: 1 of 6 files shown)


jenn-rhim:
@yachtcaptain23  let's make the data plots within 90-95% of the grid (due to huge deviation). looks great. i'll modify the colors for the data points in the next iteration.

app/javascript/publishers/dashboard_chart.js
```diff
@@ -0,0 +1,105 @@
...
+
+  Chart.defaults.global.defaultFontFamily = 'Poppins';
+
+  new Chart(canvas, {
+    type: 'line',
+    data: {
+      labels: createLabels(data[0]['ymd']),
+      datasets: [
+        {
+          label: 'Retrievals',
```
dlipeles: Let's call this Downloads to be consistent with the referral promo. (I'll make an item to fix the referral promo  box naming, it should actually be installs).

app/javascript/publishers/dashboard_chart.js
```diff
@@ -0,0 +1,105 @@
...
+          data: data.map(x => x.retrievals),
+          borderColor: '#F88469',
+        },
+        {
+          label: 'First Runs',
```
dlipeles: Let's call this Installs

app/javascript/publishers/dashboard_chart.js
```diff
@@ -0,0 +1,105 @@
...
+          data: data.map(x => x.first_runs),
+          borderColor: '#7B82E1',
+        },
+        {
+          label: 'Finalized',
```
dlipeles: Let's use 30-Day-Use here (Jenn's term for this)

app/views/publishers/home.html.slim
```diff
@@ -23,6 +23,14 @@ script type="text/html" id="confirm_default_currency_modal_wrapper"
 
 
 = hidden_field_tag 'publisher_id', current_publisher.id
+- if current_publisher.promo_registrations.present?
+  = hidden_field_tag 'all-channels-referral-stats-starting-month', current_publisher.promo_registrations.order(created_at: :asc).first.created_at.strftime("%Y-%m-%d")
+  - referrals_hidden_tags = []
+  - current_publisher.promo_registrations.each do |promo_registration|
+    - hidden_tag_name = 'referral-stat-' + promo_registration.referral_code
```
dlipeles: small fix, let's adjust from the all caps to something like "Stats for referral code: BATS123" rather than "REFERRAL-STAT - BATS123"

dlipeles:
Two comments 
1. If the publisher has more than a few referral codes, the charts will stack too deep, consider adding a dropdown select that renders one chart at a time. 


feb-19-2019 14-09-45

2. Although it is probably fine for this iteration, consider limiting the date range to 1 month at a time. 

Otherwise this is probably good enough to get the information out, we'll refine it in the dashboard remake.

yachtcaptain23:
I agree that we should limit the number of promos shown. There's some with greater than 10 promos. Going to implement your suggestion with regards to that dropdown.

I'll implement date range too.

yachtcaptain23:
Merged in https://github.com/brave-intl/publishers/pull/1665

PR #1617, opened 2019-02-16
Wallet Refactor Rebase

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


app/helpers/publishers_helper.rb
```diff
@@ -33,64 +40,64 @@ def publisher_humanize_balance(publisher, currency)
...
-    return if currency == "BAT" || currency.blank?
-    if balance = publisher.wallet &&
-        publisher.wallet.contribution_balance.is_a?(Eyeshade::Balance) &&
-        publisher.wallet.contribution_balance
-      converted_amount = '%.2f' % balance.convert_to(currency)
-      I18n.t("helpers.publisher.balance_pending_approximate", amount: converted_amount, code: currency)
+  def publisher_channel_bat_balance(publisher, channel_identifier)
+    channel_balance = publisher.wallet&.channel_balances.dig(channel_identifier)
+    if channel_balance&.amount_bat.present?
+      '%.2f' % balance.amount_bat
```
nvonpentz: Whoops, this is my fault: `balance` here should be `channel_balance`, or else channel balances will display to the publishers as 'Unavailable'.

image

app/javascript/publishers/home.js
```diff
@@ -15,6 +15,20 @@ const SELECT_CURRENCY = '---';
 const BASIC_ATTENTION_TOKEN = 'BAT'
 const UNAVAILABLE = 'unavailable';
 
+function formatConvertedBalance(amount, currency) {
+  if (isNaN(amount) || amount === null ) {
+    return `${currency} ${UNAVAILABLE}`;
+  } else {
+    let amount = formatAmount(amount)
+    return `~ ${amount} ${currency}`;
```
nvonpentz: I think we need to change line 22 and 23 to:

```
let amountFormatted = formatAmount(amount)
return `~ ${amountFormatted} ${currency}`;
```
to avoid variable scoping problems.  Doing so fixed an issue I was seeing in which a converted overall currency was `NaN` after this js was run when changing default currency.

lib/eyeshade/overall_balance.rb
```diff
@@ -0,0 +1,27 @@
+# Fees applied to channel balances
+module Eyeshade
+  class OverallBalance < BaseBalance
+    def initialize(rates, default_currency, accounts)
+      super(rates, default_currency)
+
+      @amount_probi = 0
```
corymcdonald: It might make sense to have this property be able to be set in the `BaseBalance` class and then the `amount_bat` or `amount_probi` can be calculated based upon what is already defined in the class.

Although I suppose the next question is that what if there is both an `amount_probi` set and we're trying to set `amount_bat`. 

Maybe it's best to keep it like this for now and not assume anything?

app/javascript/publishers/home.js
yachtcaptain23: Fixed

lib/eyeshade/overall_balance.rb
yachtcaptain23: Fixed. 

@nvonpentz I agree, I think we should only track either `amount_probi` or `amount_bat` rather than tracking the same data twice.

app/helpers/publishers_helper.rb
```diff
@@ -33,64 +40,64 @@ def publisher_humanize_balance(publisher, currency)
...
+      I18n.t("helpers.publisher.balance_pending_approximate",
+             amount: '%.2f' % overall_balance.amount_default_currency,
+             code: overall_balance.default_currency)
+    else
+      I18n.t("helpers.publisher.conversion_unavailable", code: publisher.default_currency)
     end
-    today.strftime("%B 8th")
   end
 
-  def publisher_converted_balance(publisher)
```
corymcdonald: Can we rename the associated tests for this method?

app/javascript/publishers/home.js
```diff
@@ -26,39 +40,34 @@ function showPendingContactEmail(pendingEmail) {
...
-    lastDepositBatAmount.innerText = settlement.amount.bat.toFixed(2);
-    lastDepositConvertedAmount.style.display = settlement.amount.currency === "BAT" || settlement.amount.currency === null ? 'none' : 'block';
-    lastDepositConvertedAmount.innerText = formatBalance(settlement.amount.converted, settlement.amount.currency);
-  }
-  else {
+    lastDepositDate.innerText = formatFullDate(new Date(lastSettlementBalance.timestamp * 1000)); // Convert to milliseconds
+
+    lastDepositBatAmount.innerText = lastSettlementBalance.amount_bat;
+    lastDepositConvertedAmount.style.display = lastSettlementBalance.settlement_currency === "BAT" || lastSettlementBalance.settlement_currency === null ? 'none' : 'block';
+    lastDepositConvertedAmount.innerText = formatBalance(lastSettlementBalance.amount_settlement_currency, lastSettlementBalance.settlement_currency);
```
corymcdonald: We removed the `formatBalance` function but we're still calling it ⚠️

lib/eyeshade/base_balance.rb
```diff
@@ -0,0 +1,54 @@
...
+    private
+
+    # Expects an Integer and returns a BigDecimal
+    def probi_to_bat(probi)
+      probi.to_d / 1E18
+    end
+
+    # Expects a BigDecimal returns an Integer
+    def bat_to_probi(bat)
+      (bat * BigDecimal.new('1.0e18')).to_i
```
corymcdonald: consistency 🙂  
```suggestion
      (bat * BigDecimal.new("1.0e18")).to_i
```

lib/eyeshade/base_balance.rb
```diff
@@ -0,0 +1,54 @@
...
+      fees_probi = (total_probi * fee_rate).to_i
+      amount_probi = total_probi - fees_probi
+      {"amount" => amount_probi, "fees" => fees_probi}
+    end
+
+    # Expects and returns BigDecimals
+    def convert(amount_bat, currency)
+      return amount_bat if currency == "BAT"
+      return if @rates[currency].nil?
+      amount_bat * @rates[currency]
```
nvonpentz: I'm still getting this error on the latest commit:

I'm finding that `@rates[currency]` is actually a string, not a decimal, so I'm getting

```
String can't be coerced into BigDecimal
```
on this line.

I didn't noticed this error when developing because the rates in the offline wallet are not strings. It'd be great if we could change this too so we don't develop with bad data.

yachtcaptain23:
Merged in via https://github.com/brave-intl/publishers/pull/1644

PR #1620, opened 2019-02-19
Create Payments feature

### LINKED ISSUE #1333: Review reports and invoices
Action Items:
- [ ] Allow an admin review reports and invoices uploaded by a `publisher_partner`

Note: This effort is partially blocked by #1330, as in production, I'll prefer we use a private S3 bucket rather than store blobs in the database. Reading/writing blobs in a RDMS is not preferable should the file size be too big, as it'll consume too much of the memory in the database during the read part.

corymcdonald:
This is the PR for the Payments feature, handles both the administrator workflow and the client side workflow.

Closes #1620 
Closes #1333


app/javascript/components/Modal.tsx
```diff
@@ -0,0 +1,111 @@
+import * as React from "react";
```
dlipeles: Consider putting this and style in their own folder 📁 

`app/javascript/components/modal/Modal.tsx`
`app/javascript/components/modal/ModalStyle.ts`

app/controllers/admin/invoices_controller.rb
```diff
@@ -0,0 +1,83 @@
+class Admin
```
dlipeles: Consider aggregating invoices across partners in the org 🍊

This snippet might be useful
```ruby
        membership = Membership.find_by(user_id: current_publisher.id)
        organization = Organization.find(membership.organization_id)
        partners = Membership.where(organization_id: organization.id).map { |membership| membership.user_id}
```

test/controllers/partners/invoice_files_controller_test.rb
```diff
@@ -0,0 +1,53 @@
...
+require "webmock/minitest"
+
+class Partners::InvoicesControllerTest < ActionDispatch::IntegrationTest
+  include Devise::Test::IntegrationHelpers
+
+  describe '#create' do
+    before do
+      partner = partners(:default_partner)
+      sign_in partner
+      Invoice.destroy_all
```
dlipeles: Do we need to destroy all in the tests ? 💥

app/javascript/views/payments/Payments.tsx
```diff
@@ -1,28 +1,109 @@
...
 import UpholdCard from "./upholdCard/UpholdCard";
 
-export default class Payments extends React.Component {
+import routes from "../routes";
+
+export interface IInvoice {
+  id: string;
+  date: string;
+  url: string;
+  finalized_amount: string;
```
yachtcaptain23: Is this supposed to be in camelCase or snake_case?

db/schema.rb
```diff
@@ -78,6 +78,33 @@
...
+    t.boolean "archived", default: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["invoice_id"], name: "index_invoice_files_on_invoice_id"
+    t.index ["uploaded_by_id"], name: "index_invoice_files_on_uploaded_by_id"
+  end
+
+  create_table "invoices", id: :uuid, default: -> { "uuid_generate_v4()" }, force: :cascade do |t|
+    t.uuid "partner_id"
+    t.datetime "date"
```
yachtcaptain23: I believe we'd want `t.date` rather than `t.datetime`

db/schema.rb
```diff
@@ -78,6 +78,33 @@
...
+    t.string "amount", default: "0"
+    t.string "finalized_amount"
+    t.boolean "paid", default: false
+    t.uuid "paid_by_id"
+    t.datetime "payment_date"
```
yachtcaptain23: Ditto here for `t.date`

test/controllers/partners/invoice_files_controller_test.rb
corymcdonald: I'd rather have a clean slate each test case, so we can ensure that there is a correct state during the test 😄

app/controllers/admin/invoices_controller.rb
corymcdonald: So my only issue is that if I do that we'd need to display all the invoices on the organization page. Which is doable, but since each uphold account is linked to a user then we'd need to figure out which account is going to be the one that gets paid out. 

I think we can go back and iterate on this once we have some actual usage information. For now we should communicate with BizDev that it's one partner per account?

dlipeles: Isn't that also true as it stands right now? Any partner in an organization can upload invoices, so bd will need to decide which partner from the organization to payout.

test/controllers/partners/invoice_files_controller_test.rb
dlipeles: 👍

app/controllers/admin/invoices_controller.rb
corymcdonald: So the way that the system currently works is the invoice button displays the most recent invoice available to the partner, and this is per partner. So I'd have to figure out some way to display multiple invoices on the front end for the different partners. One solution is sharing the invoice across the organization

This would allow different users to login and upload their own reports to that one invoice - this is totally doable. Given Mandar's statement on how invoices and reports come from different areas of the same company it might make sense to have a couple of different accounts which can access the same invoice. Otherwise it's kind of a useless functionality 🤔 

I'd be worried about having to redo some the UI that we have already demo'd to make this possible, but I'm optionally for it if this is something that we thinks makes more sense

Curious on what your thoughts @yachtcaptain23

db/schema.rb
corymcdonald: This is it - https://github.com/brave-intl/publishers/pull/1620/commits/ab7050f639094d5c771697143338194f0046ad3c

db/schema.rb
corymcdonald: Agreed! https://github.com/brave-intl/publishers/pull/1620/commits/ab7050f639094d5c771697143338194f0046ad3c

app/javascript/views/payments/Payments.tsx
corymcdonald: Could work both ways

https://github.com/brave-intl/publishers/pull/1620/commits/ebcdf99c43312dfa555ca160238f83aed56b497c

PR #1622, opened 2019-02-20
Fix display of users if they only have a pending email

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/views/admin/publishers/index.html.slim
+++ b/app/views/admin/publishers/index.html.slim
@@ -26,7 +26,7 @@ div.row
             tbody
               - @publishers.left_joins(:channels).each do |publisher|
                 tr.gradeX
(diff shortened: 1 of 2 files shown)


corymcdonald:
Let's put this in the publisher model instead

```ruby
def email
  self[:email] || pending_email
end
```

yachtcaptain23:
@corymcdonald I disagree from a principle of least surprise. Perhaps another method name is more apt, but since it's a one-off issue, I prefer to keep the logic in the view.

corymcdonald:
@yachtcaptain23 Fine by me, but can we also update the show page header? If the user doesn't have a name we should display the email

PR #1628, opened 2019-02-20
Save why publishers and channels were not included in the payout report

### LINKED ISSUE #1618: Include all verified channels and publishers in payout report
Currently the payout report will only include verified channels and publishers that will receive payment.  As a result, there isn't an audit trail for publishers that were not paid out, and when publishers ask why they didn't get their payment, we don't have a document explaining why.  Ideally we should be tracking why each channel or publisher was not paid.  This will help us identify the source and scope of any bugs as well as provide better resources to our support team.

A verified channel / publisher will not be paid if 
* Their Uphold account `user status` is not `ok`, i.e. if it is `pending`, `restricted`, or `blocked`.
* They are not an Uphold member
* Brave is not authorized to read their Uphold account info
* They don't have a positive balance

Let's add these columns to the potential_payments table:
* `user_status` : string of (ok, pending, restricted, blocked)
* `was_member` : bool
* `reauthorization_needed` : bool

Instead of ignoring publishers that won't be paid out [here](https://github.com/brave-intl/publishers/blob/472654417873a56abc824a24c11423ac16339849/app/services/payout_report_publisher_includer.rb#L20) and [here](https://github.com/brave-intl/publishers/blob/472654417873a56abc824a24c11423ac16339849/app/services/payout_report_publisher_includer.rb#L35), let's create the potential payment for every verified channel and publisher. Then on the publisher page of the admin dashboard, an admin should be able to see a payment status section that displays why a publisher was or was not paid in the most recent final payout report.

### LINKED ISSUE #1631: Include uphold id in payout report

nvonpentz:
Resolves #1618 
Resolves #1631

Also, add uphold id to payout report.

Publisher page now displays report status information for the publisher and it's channels for the most recent final payout report:
image

Payout report page now displays the total amount of payments that are expected in the report as well as the actual number:
image

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

This can by tested by creating a publisher and a verified channel then generating payout reports with various states of suspension, uphold status, and balance amounts for the owner account and the channel.

Reviewer Checklist:

Tests

Security:


app/services/payout_report_publisher_includer.rb
```diff
@@ -10,40 +10,52 @@ def perform
     publisher_has_unsettled_balance = false
 
     wallet = @publisher.wallet
-    # If the wallet has been blocked from uphold's terms of service. In this state users are unable to login or access uphold.
-    return if wallet.nil? || wallet.blocked?
```
nvonpentz: The wallet should never be nil, and we want to record if it's blocked, so removing this line entirely

app/services/payout_report_publisher_includer.rb
```diff
@@ -10,40 +10,52 @@ def perform
...
 
+    uphold_status = wallet.uphold_account_status
+    uphold_member = wallet.is_a_member?
+    uphold_connected = wallet.uphold_id.present? # TODO consider adding sub category for when wallet.action == 're-authorize'
```
nvonpentz: I realized this is not needed and can also be expressed as `uphold_account_status == nil`, I will track reauthorization need instead.

app/helpers/admin_helper.rb
```diff
@@ -15,4 +15,14 @@ def nav_link(text, path)
       link_to text, path, options
     end
   end
+
+  def payout_report_status_header(account_type)
+    report_date = PayoutReport.most_recent_final_report.created_at.strftime("%b %d")
```
yachtcaptain23: Nit picking: Can we do `%Y-%m-%d`?

db/schema.rb
```diff
@@ -235,6 +236,10 @@
     t.string "amount", null: false
     t.string "fees", null: false
     t.string "url"
+    t.string "uphold_status_was"
+    t.boolean "reauthorization_was_needed", null: false
```
yachtcaptain23: Would you update the schema to make sure `null: false` isn't included?

db/migrate/20190219194426_add_report_status_info_to_potential_payments.rb
```diff
@@ -0,0 +1,8 @@
+class AddReportStatusInfoToPotentialPayments < ActiveRecord::Migration[5.2]
+  def change
+    add_column :potential_payments, :uphold_status_was, :string
```
yachtcaptain23: IMO you should drop `_was` in all these columns. It's fairly self explanatory it's a snapshot. 

@corymcdonald @dlipeles do you agree?

app/services/payout_report_publisher_includer.rb
```diff
@@ -10,40 +10,52 @@ def perform
...
+    uphold_status = wallet.uphold_account_status
+    uphold_member = wallet.is_a_member?
+    reauthorization_needed = wallet.action == "re-authorize"
+    suspended = @publisher.suspended?
+
+    # Create potential payment for referrals
     probi = wallet.channel_balances[@publisher.owner_identifier].probi_before_fees # probi = balance
-    if probi.positive?
-      publisher_has_unsettled_balance = true
+    publisher_has_unsettled_balance = probi.positive? ? true : false
```
yachtcaptain23: `publisher_has_unsettled_balance = probi.positive?`

db/schema.rb
nvonpentz: good catch!

app/helpers/admin_helper.rb
nvonpentz: I agree that your format is better for most things, but in this case I wonder if it makes it less readable.  The date here should really only be from the current month if everything is working, so I think including the year might be overkill.

image
vs
image

@corymcdonald @dlipeles do you have an opinion?

app/services/payout_report_publisher_includer.rb
nvonpentz: Resolved in 023ddd4

app/helpers/admin_helper.rb
dlipeles: BD has been asking us to make things more user friendly so I'll vote for month/day

db/migrate/20190219194426_add_report_status_info_to_potential_payments.rb
corymcdonald: aye 👍

nvonpentz: thanks guys, will update

nvonpentz:
This now also resolves #1631

app/helpers/admin_helper.rb
nvonpentz: I will keep it this way unless you want to veto @yachtcaptain23

db/migrate/20190219194426_add_report_status_info_to_potential_payments.rb
nvonpentz: fb4c7d7

PR #1639, opened 2019-02-22
Remove IP Address from Publishers Table

### LINKED ISSUE #1035: Stop storing publisher IP
We are currently storing `last_sign_in_ip` and `current_sign_in_ip` on the publishers table.  Currently these are not being used for anything.  In the interest of preserving user privacy, let's stop storing these until we have a defined use case.

corymcdonald:
Kept the trackable module but have `attr_accessor`s so that devise can still access these fields in it's module, however it won't save those values.
https://github.com/plataformatec/devise/blob/master/lib/devise/models/trackable.rb#L26

Closes #1035


app/controllers/admin/publishers_controller.rb
```diff
@@ -24,6 +24,7 @@ def index
 
   def show
     @publisher = Publisher.find(params[:id])
+    @potential_referral_payment = PayoutReport.most_recent_final_report&.potential_payments&.where(publisher_id: @publisher.id)&.first
```
corymcdonald: I moved this from the view to the controller

PR #1644, opened 2019-02-25
Refactor Eyeshade::Wallet and Eyeshade::Balance classes

yachtcaptain23:
Rebase of https://github.com/brave-intl/publishers/pull/1438 (Please follow link for details)

Closes https://github.com/brave-intl/publishers/issues/1404
Closes https://github.com/brave-intl/publishers/issues/1309

* Remove Eyeshade::Balance & tests
* Replace Eyeshade::Balance with:
  * Eyeshade::OverallBalance
  * Eyeshade::ChannelBalance
  * Eyeshade::ReferralBalance
  * Eyeshade::LastSettlementBalance
* Update Eyeshade::Wallet to use new new balances
Create EyeshadeHelper module for stubbing eyeshade requests for tests
Refactor Eyeshade wallet services to use new Eyeshade::Wallet and Balance format
Update helper methods that display publisher balances
Change how the wallet is sent to the client via js
* Remove PublishersController#balance, create PublishersController#wallet
* Remove the JSON builder, instead do `publisher.wallet.to_json` in
  PublishersController#wallet
* Remove methods that recalculate balances with js, instead
  just use the /wallet endpoint as source of truth
Update other tests to use EyeshadeHelper class
Miscellaneous admin style changes
Use to_time instead of to_date to ensure consistent last_settlement_timestamps regardless of the user's timezone

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


lib/eyeshade/last_settlement_balance.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    private
+
+    def calculate_last_settlement(transactions)
+      return {} if transactions == []
+       # Find most recent settlement transaction
+      last_settlement_time = transactions.select { |transaction|
+        transaction["settlement_amount"].present?
+      }.last["created_at"].to_time # Eyeshade returns transactions ordered_by created_at
```
nvonpentz: this `.last["created_at"]` will be a problem when there are transactions, but none of them are settlement transactions

image

yachtcaptain23: I think we're fine. The prod data yields the following:
```
select count(*) from account_transactions
where created_at is null;
```
0

nvonpentz: replied via slack

lib/eyeshade/last_settlement_balance.rb
```diff
@@ -0,0 +1,45 @@
...
+    private
+
+    def calculate_last_settlement(transactions)
+      return {} if transactions == []
+       # Find most recent settlement transaction
+      last_transaction = transactions.select { |transaction|
+        transaction["settlement_amount"].present?
+      }.last # Eyeshade returns transactions ordered_by created_at
+
+      return {} if last_transaction.nil?
```
corymcdonald: If transactions is an array this won't ever be nil. Instead you should use `last_transactions.blank?` so that if last_transaction is nil or empty then this will return. 

You also can remove `return {} if transactions == []`

lib/eyeshade/channel_balance.rb
```diff
@@ -0,0 +1,17 @@
+# Fees applied
```
corymcdonald: Let's start adding this to the top of files
```suggestion
# frozen_string_literal: true
# Fees applied
```

Issue #538, opened 2018-01-26
Need the global nav on 2FA pages


jenn-rhim:
Global nav is meant to be persistent

Current 2FA pages:
![screen shot 2018-01-25 at 4 53 29 pm](https://user-images.githubusercontent.com/4369856/35420176-e771331a-01f0-11e8-8733-9d3ea40d0bd2.png)

Should be:
![image](https://user-images.githubusercontent.com/4369856/35420159-cc25326e-01f0-11e8-91fc-ac769ae62706.png)

nvonpentz:
I think this would be best addressed when we port the dashboard to bootstrap 4

yachtcaptain23:
Users can work around. Going to get rid of as we realistically won't ever get to this and there's a straight forward workaround for the user.

PR #1649, opened 2019-02-26
New Navigation 🏷 

dlipeles:
## New Navigation 🏷 

This PR takes us closer to UI 3.0 with a new navbar, `Partners` will find the new navigation on the UI pages (`Referrals` & `Payments`)

#### Features 
- New look and feel to navigation 
- Responsive design supports all viewports
- Selected page has a purple indicator 
- Fixes some display issues in legacy nav

#### Notes
- For now the `Dashboard` will still support legacy navigation to accomodate our regular pubs. 

screen shot 2019-02-26 at 6 07 00 pm


corymcdonald:
Is this new navigation going to be on the 2FA page? We could potentially resolve #538 if we address this here

dlipeles:
@corymcdonald I would like to solve that issue with the new nav, but it'd be best done after we complete the new `Dashboard` and `Channels` page because then we can delete the legacy nav completely. 

Once that's done, we can use this new nav for every page.

app/javascript/components/navbar/NavbarStyle.ts
```diff
@@ -0,0 +1,132 @@
...
+  open: boolean;
+}
+export const DropdownMenu = styled.div`
+  position: absolute;
+  background-color: white;
+  top: 50px;
+  right: 0px;
+  width: 220px;
+  border-radius: 4px;
+  box-shadow: 2px 2px 0 0 rgba(47, 48, 50, 0.15);
```
corymcdonald: I think we should change this box-shadow to be something more along the lines of 

```suggestion
  box-shadow: 0px 1px 2px 1px rgba(47, 48, 50, 0.15);
```

app/javascript/routes/routes.ts
```diff
@@ -0,0 +1,26 @@
+/* tslint:disable:object-literal-sort-keys */
```
corymcdonald: There's already this file in `app/javascript/views/routes.ts` should you use that one instead?

app/views/layouts/application.html.slim
```diff
@@ -29,10 +29,11 @@ html
       }
 
     #main
-      - if content_for?(:nav)
-        = yield(:nav)
-      - else
-        = render("nav")
+      - if !request.original_url.include? "/partners/"
```
corymcdonald: I think instead of this we should simply check to see if the user is a `partner?`

app/javascript/components/navbar/Navbar.tsx
```diff
@@ -0,0 +1,149 @@
...
+      <Wrapper>
+        <Container>
+          <BraveLogo />
+          <Navigation navbarSelection={this.props.navbarSelection} />
+          <DropdownGroup onClick={this.toggleMenu}>
+            <Dropdown menuOpen={this.state.menuOpen} />
+            <AvatarContainer>
+              <AvatarIcon />
+            </AvatarContainer>
+            <Name>Partner</Name>
```
corymcdonald: Are we planning on passing this in via props

app/javascript/components/navbar/NavbarStyle.ts
```diff
@@ -0,0 +1,132 @@
...
+  padding-top: 6px;
+  color: #222326;
+  margin-right: 40px;
+  text-transform: uppercase;
+  user-select: none;
+  cursor: pointer;
+
+  ${(props: Partial<INavProps>) =>
+    props.selected === true &&
+    `
```
yachtcaptain23: I'm confused, what does this do?

app/javascript/routes/routes.ts
dlipeles: I'm agreed we should have a single location for `routes`, could we keep it outside of views, in its own `routes` folder?

app/views/layouts/application.html.slim
dlipeles: I would prefer to do it that way, but we need to provide legacy nav on the dashboard, for partners as well as publishers.

app/javascript/components/navbar/NavbarStyle.ts
dlipeles: Allows us to pass the `selected` prop into `<Nav>` components like so :

```tsx
<Nav selected={props.navbarSelection === NavbarSelection.Channels}>
        {locale.navbar.channels}
      </Nav>
```

When `selected === true` then we render the purple highlight underneath the nav

PR #1654, opened 2019-02-28
Fix nil details error

dlipeles:
Fix to resolve this [Sentry Error](https://sentry.io/organizations/brave-software/issues/750563805/?project=225116&query=is%3Aunresolved+json&statsPeriod=14d&utc=false)


nvonpentz:
There shouldn't be any verified channels that don't have details, so this is a good temporary fix if the channels json is not updating, but it does kind of ignore the root problem.

corymcdonald:
I wasn't able to find the exact case when this happens in prod 😕 I imagine there might be a race condition or something weird going on. Either way better to try and remove this from blowing up the cache job

PR #1660, opened 2019-02-28
Post settlement report

### LINKED ISSUE #1604: Support posting of settlement report via admin dashboard
Typically settlement reports are uploaded to Eyeshade through the swagger interface.  With manual payments, it makes sense to move this interface to the publishers dashboard since Publishers needs to process these payments by marking the unpaid invoices as paid before forwarding on to Eyeshade.  The flow will be:

1. Admin uploads settlement report to admin dashboard
2. For each manual settlement, Publishers looks up the corresponding unpaid invoice via the documentId, and marks the corresponding invoice as paid
3. Publishers forwards the settlement report to eyeshade via `POST /v2/publishers/settlement`

Related to https://github.com/brave-intl/bat-ledger/issues/519

corymcdonald:
Closes #1604

Creates the ability to upload a file which will then upload the report to Eyeshade 🎉


app/services/eyeshade/publishers.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class Eyeshade::Publishers < Eyeshade::BaseApiClient
+  RESOURCE = "/v2/publishers"
+
+  def create_settlement(body: )
+    return {} if Rails.application.secrets[:api_eyeshade_offline]
+
+    connection = Faraday.new(url: api_base_uri) do |config|
```
corymcdonald: Thinking about moving this connection up to the Eyeshade::BaseApiClient @yachtcaptain23 Thoughts?

Gemfile
```diff
@@ -38,7 +38,8 @@ gem "dnsruby", "~> 1.60.0", require: false
 gem "email_validator", "~> 1.6"
 
 # HTTP library wrapper
-gem "faraday", "~> 0.9.2", require: false
+gem "faraday", "~> 0.9.2"
+gem 'faraday_middleware', '~> 0.13.1'
```
yachtcaptain23: Do we need this?

app/services/eyeshade/publishers.rb
yachtcaptain23: IMO you should be able to copy the code style that uses `Eyeshade::BaseApiClient`. You can see a sample `post` in `uphold_services/card_creation_service.rb`

app/views/admin/payout_reports/index.html.slim
```diff
@@ -1,4 +1,17 @@
-h1 Payout Reports
+.d-flex
+  h1 Payout Reports
+  .ml-4
+    = form_tag upload_settlement_report_admin_payout_reports_path, id: "settlementForm", multipart: :true
+      label
+        = file_field_tag :file, { class: 'd-none', id: "settlementFile",  accept: ".json,application/json" }
```
yachtcaptain23: What class is `d-none`?

corymcdonald: `display: none` it comes from bootstrap

Gemfile
corymcdonald: We do need this to send JSON to eyeshade. I don't think we had any `POST` requests to eyeshade before but when I did not include this the request was unable to be sent properly.

corymcdonald: https://github.com/brave-intl/publishers/pull/1660/commits/8500d5d4ffa56959ddd1d4705a6c6eb481f2dd7c

corymcdonald:
Step missing:  need to look up and read the json to see if there are any manual payouts in the settlement report

app/services/eyeshade/publishers.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class Eyeshade::Publishers < Eyeshade::BaseApiClient
```
yachtcaptain23: IMO We should rename this to `Eyeshade::Settlements`. Thoughts?

app/services/eyeshade/publishers.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def create_settlement(body: )
+    return {} if Rails.application.secrets[:api_eyeshade_offline]
+
+    response = connection.post do |request|
+      request.headers["Authorization"] = api_authorization_header
+      request.headers["Content-Type"] = "application/json"
+      request.url("#{RESOURCE}/settlement")
+
+      request.body = body.to_json
```
yachtcaptain23: Would you write exception handling in case if we get a 503?

app/controllers/admin/payout_reports_controller.rb
```diff
@@ -31,6 +31,31 @@ def notify
...
+
+      invoice.update(
+        paid: true,
+        payment_date: Date.today,
+        status: "paid",
+        paid_by: current_publisher
+      )
+    end
+
+    Eyeshade::Publishers.new.create_settlement(body: json)
```
yachtcaptain23: Would you move this line above `invoice.update` block? Otherwise if Eyeshade throws an error, this would show the invoice as paid when it's not.

app/services/eyeshade/publishers.rb
corymcdonald: I think we should keep it as a publishers because that's what the actual resource on eyeshade is. I imagined that there would be other things under the same uri structure 🤔

app/controllers/admin/payout_reports_controller.rb
corymcdonald: Ah good catch!
https://github.com/brave-intl/publishers/pull/1660/commits/c26b02ec4f3dfe7db967669299103bd36bce630b

app/services/eyeshade/publishers.rb
corymcdonald: Sure can, 
https://github.com/brave-intl/publishers/pull/1660/commits/c26b02ec4f3dfe7db967669299103bd36bce630b

PR #1661, opened 2019-02-28
Update Promo Dashboard

### LINKED ISSUE #1383: Downloads / Installs - changes needed in Publisher dashboard
Two asks;

a) Need to include downloads (actual intent to download binary) into the Publisher dashboard with helper logos
b) Move the admin console functionality for codes to be the same for the Publisher dashboard

Downloads: Intent to download the binary
Install (First Open): Binary installed and opened first time
Confirmations: User is active for 30+ days

and c) need to email Publisher owners that terminology change as well.

cc @jenn-rhim @dlipeles

dlipeles:
image
### Update Promo Dashboard

closes #1383 

#### Features

- Adds the new `Installed` field to the promo widget
- Corrects the download field to use `retrievals` instead of `first_runs`
- Adds the corrected values to Admin Dashboard as well


corymcdonald:
Can we use the same color as the referral promo font over on the left? I think it will make it look a little bit better and more in line with the theme of the component

corymcdonald:
Can we add this to the admin dashboard?

PR #1663, opened 2019-03-01
Fix TypeError

corymcdonald:
Couple of sentry issues that were found in staging, might as well address them as well

### DIFF
--- a/app/controllers/admin/stats/top_balances_controller.rb
+++ b/app/controllers/admin/stats/top_balances_controller.rb
@@ -3,11 +3,11 @@ class Admin::Stats::TopBalancesController < AdminController
     @limit = params[:limit].present? ? params[:limit].to_i : 25
     @result = JSON.parse(case params[:type]
     when Eyeshade::TopBalances::CHANNEL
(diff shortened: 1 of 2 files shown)


app/controllers/admin/stats/top_balances_controller.rb
```diff
@@ -3,11 +3,11 @@ def index
     @limit = params[:limit].present? ? params[:limit].to_i : 25
     @result = JSON.parse(case params[:type]
     when Eyeshade::TopBalances::CHANNEL
-      Rails.cache.fetch(Cache::EyeshadeStatsJob::EYESHADE_TOP_CHANNEL_BALANCES) || []
+      Rails.cache.fetch(Cache::EyeshadeStatsJob::EYESHADE_TOP_CHANNEL_BALANCES) || "{}"
```
yachtcaptain23: There should be `"[]"`

corymcdonald: https://github.com/brave-intl/publishers/pull/1663/commits/5cfd73e7dc3e851dc331bb9194ecb0abd505ae08

PR #1665, opened 2019-03-04
Referral chart on the main dashboard

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:


yachtcaptain23:
screen shot 2019-03-05 at 9 44 23 am

yachtcaptain23:
Labels:
- [ ] Should be dots rather than a box for the label or time series
- [ ] Would be good to hide year
- [ ] Use dates like March 15 rather than 03-15
- [ ] Swapping colors between 30-day-use
- [ ] Hide gray area

app/javascript/publishers/dashboard_chart.js
```diff
@@ -0,0 +1,91 @@
...
+    currentMax = value.retrievals > currentMax ? value.retrievals : currentMax;
+    currentMax = value.first_runs > currentMax ? value.first_runs : currentMax;
+    currentMax = value.finalized > currentMax ? value.finalized : currentMax;
+  });
+  return (currentMax * 100 / 95)
+}
+
+function createChart(data, title, suggestedMax) {
+  var wrapper = document.getElementById('channel-referrals-stats-chart');
+  var canvas = document.getElementById('chanel-referrals-stats-chart-canvas);
```
corymcdonald: I think you're missing a single quote

app/javascript/packs/referral_charts.jsx
```diff
@@ -0,0 +1,183 @@
...
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
+
+export function renderReferralCharts() {
+  let props = {
+    referralCodes: JSON.parse(document.getElementById('referrals-hidden-tags').value)
```
corymcdonald: might be better to do something like this

const { value } = document.getElementById('referrals-hidden-tags')
if(value === undefined) {
  return;
}
let referralCodes = JSON.parse(value);

app/javascript/packs/referral_charts.jsx
```diff
@@ -0,0 +1,183 @@
+import * as React from "react";
```
corymcdonald: Can you run prettier on this file?

Issue #1602, opened 2019-02-11
Do not use Eyeshade balance for partner referral balance


nvonpentz:
A partner's referral balances should be the sum of their unpaid authorized invoices, which is information only available on publishers.  Their contribution balance should still be retrieved from Eyeshade.  It may be best to merge the wallet refactor #1438 before making these changes to the wallet.

yachtcaptain23:
Blocked until wallet refactor is merged: https://github.com/brave-intl/publishers/pull/1438

PR #1667, opened 2019-03-05
Display Partner Balance on homepage

corymcdonald:
Closes issue #1602

Things left to do

- [x] Determine if prettier should have run on the README 🤔 
- [x] Write some tests


app/models/partner.rb
```diff
@@ -13,7 +13,26 @@ def ensure_role
...
     self[:name] || self[:email]
   end
+
+  private
+
+  def invoice_amount
+    invoices = Invoice.where(partner_id: id, paid: false)
+    amounts = invoices.map { |i| i.finalized_amount || i.amount }
+
+    amounts.map { |x| x.gsub(/,/, '').to_i }.reduce(:+)
```
dlipeles: Quick question, do we prefer .gsub or trim? I've been having good results with 
```ruby
str.tr(",", "")
```

But I'm not sure which is better

app/models/partner.rb
```diff
@@ -13,7 +13,26 @@ def ensure_role
     self.role = 'partner'
   end
 
+  def balance
+    wallet.contribution_balance.amount_bat + invoice_amount
+  end
+
+  def balance_in_currency
+    wallet.contribution_balance.add_bat(invoice_amount)
+
```
dlipeles: nit: maybe we can rm the whitespace on `line 22`

app/models/invoice.rb
```diff
@@ -24,6 +24,10 @@ class ReadOnlyError < StandardError; end
 
   validates :date, uniqueness: { scope: :partner_id }
 
+  # Ensure these two values are numbers even though field is a string
+  validates :amount,  numericality: true, allow_nil: true
+  validates :finalized_amount,  numericality: true, allow_nil: true
```
dlipeles: Maybe we can even validate further with one of these

 https://guides.rubyonrails.org/active_record_validations.html#numericality

Perhaps ```:greater_than_or_equal_to: 0```

Not sure if that would clash with nil though, totally your call, just food for thought 🍕

app/models/partner.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/1667/commits/092d85279ad37ad6a8ce82db79c7972594e84ead

app/models/partner.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/1667/commits/092d85279ad37ad6a8ce82db79c7972594e84ead

app/models/invoice.rb
corymcdonald: Unfortunately we can't allow_nil if we do this 😢

app/models/partner.rb
```diff
@@ -13,7 +13,25 @@ def ensure_role
...
+    wallet.contribution_balance.amount_default_currency
+  end
+
   def name
     self[:name] || self[:email]
   end
+
+  private
+
+  def invoice_amount
```
dlipeles: @yachtcaptain23 passed me this feedback the other day, I think it applies to this section 

consider 

```ruby 
def invoice_amount
    invoices = Invoice.where(partner_id: id, paid: false)
    amounts = invoices.map { |i| i.finalized_amount || i.amount }

    amounts.map { |x| x.tr(",", "").to_i }.reduce(:+) || 0
end
```

His rationale is prevent `add_bat` from throwing an error if the amount is `nil`

corymcdonald: Done! 
https://github.com/brave-intl/publishers/pull/1667/commits/91b10f64f07745e510dcba4d2625bb439f90a752

PR #1673, opened 2019-03-06
Manual Payout Report 🏷 

### LINKED ISSUE #1603: Support manual payout report generation
We'd like manual payouts to be tracked in Eyeshade, so we should include these transactions in the payout reports.  Let's create a new type of payout report entry, with the same fields as the rest except:
```
{
  "type": "manual", 
  "documentId": "fba98492-8ed7-4f9b-a2f6-dc62dcd0c1a3", // uuid of the authorized invoice
  "authority": "nick@brave.com" // email of BD team that approved the payment
}
```

Admins should be able to generate payout reports including only manual payments so we can they can do this outside of the monthly cadence easily.

dlipeles:
screen shot 2019-03-06 at 10 48 35 pm

Closes #1603 

#### Features
- New `Manual` type for payout reports, which contains partners that have finalized invoices. 
- Optionally generate a `manual` payout report for a single selected `partner`. 

#### Notes
- Regular payout reports will not include manual entries, they will be kept distinct. 
- Manual payout report `amount` will be the `finalized amount` set by BizDev
- Manual payout reports on publishers will not be generated monthly, although it is easy to generate them at any time. 

#### How to Test
- As an admin, request an invoice from a partner
- Partner may upload supporting documents
- As an admin, finalize the partner's invoice
- After finalizing, you will see a `generate manual payout` button, which will generate an single entry
- Alternatively admin can generate a `manual` type payout from the `payout reports` tab, for all Partners

#### Note

We need to make sure to nil out the invoice once the partner is paid so it is not included in the next report. 

#### TODO 

- [x] Wire in Generate Manual Report button
- [x] Ensure the rest of the payout report works as intended
- [x] Ensure only partners with finalized invoice are included 
- [x] Move manual payout json to its own file
- [x] Ensure finalized amount to probi handles floats/commas correctly 
- [x] Add unit tests


app/services/payout_report_publisher_includer.rb
```diff
@@ -60,6 +61,30 @@ def perform
...
+                                reauthorization_needed: reauthorization_needed,
+                                uphold_member: uphold_member,
+                                suspended: suspended,
+                                uphold_id: uphold_id,
+                                invoice_id: invoice.id,
+                                finalized_by_id: invoice.finalized_by_id
+                                )
+      end
+    end
+
```
nvonpentz: If a publisher was also a partner, I think this may attempt to create two potential referral payments for the same balance.  I think it would fail since we have restrictions on the number of referral payments a single publisher can have for a given payout report.  I think what we want is to conditionally include these extra fields (invoice_id, finalized_by_id, etc) to the referral potential payment depending on if the publisher is a partner or not.  


https://github.com/brave-intl/publishers/blob/98fb9788542ebc16fcd746f9836f82c2b7b73975/app/models/potential_payment.rb#L25

app/models/potential_payment.rb
```diff
@@ -20,6 +21,8 @@ class PotentialPayment < ApplicationRecord
     where("amount::numeric > ?", 0)
   }
 
+  scope :manual_to_be_paid, -> { where(kind: MANUAL) }
```
nvonpentz: Do we want to check if the amount is greater than zero and if the publisher is suspended?

app/services/payout_report_publisher_includer.rb
dlipeles: Thanks very much for the feedback @nvonpentz . Given our talk with @mandar-brave it seems there will be two payment flows, the regular one we are used to, and this manual one which addresses infinite codes. 

I took your advice and added a new `manual` parameter to `EnqueuePublishersForPayoutJob` which decides which type of potential payments to create. 

One thing we should think about is when codes become polymorphic, we should make sure they are not double counted in regular and manual payments. But for now I think we are okay. 

By the way, you had a good point that it would be tedious to generate a report for each partner individually. With the new `manual` parameter you can generate a report for all partners who currently have a `finalized invoice` 👍

dlipeles:
Thanks very much for the feedback @nvonpentz . Given our talk with @mandar-brave it seems there will be two payment flows, the regular one we are used to, and this manual one which addresses infinite codes. 

I took your advice and added a new `manual` parameter to `EnqueuePublishersForPayoutJob` which decides which type of potential payments to create. 

One thing we should think about is when codes become polymorphic, we should make sure they are not double counted in regular and manual payments. But for now I think we are okay. 

By the way, you had a good point that it would be tedious to generate a report for each partner individually. With the new `manual` parameter you can generate a report for all partners who currently have a `finalized invoice` 👍

app/controllers/admin/partners_controller.rb
```diff
@@ -65,6 +65,13 @@ def update
       end
     end
 
+    def generate_manual_payout
+      # Generate report for the selected partner
+      publisher_ids = [].push(params[:id])
+      EnqueuePublishersForPayoutJob.perform_later(final: true, manual: true, publisher_ids: publisher_ids)
```
yachtcaptain23: `EnqueuePublishersForPayoutJob.perform_later(final: true, manual: true, publisher_ids: [params[:id]])`

db/migrate/20190307022345_add_manual_to_payout_report.rb
```diff
@@ -0,0 +1,5 @@
+class AddManualToPayoutReport < ActiveRecord::Migration[5.2]
+  def change
+    add_column :payout_reports, :manual, :boolean
```
yachtcaptain23: Should add a default value.

app/services/manual_payout_report_publisher_includer.rb
```diff
@@ -0,0 +1,42 @@
+class ManualPayoutReportPublisherIncluder < BaseService
```
yachtcaptain23: @corymcdonald and I have thoughts that we should change the naming conventions that we currently have. Definitely worth it for us to have the discussion and make sure naming conventions are conventional and obvious.

I think this class should be `PayoutReport::ManualGenerationService` or `PayoutReport::ManualGeneration`. 

cc: @nvonpentz who will also have thoughts too.

app/models/json_builders/manual_payout_report_json_builder.rb
```diff
@@ -0,0 +1,25 @@
+class JsonBuilders::ManualPayoutReportJsonBuilder
+  def initialize(payout_report:)
+    @payout_report = payout_report
+  end
+
+  def build
+    contents = []
+    @payout_report.potential_payments.manual_to_be_paid.find_each do |potential_payment|
+      contents.push(
```
yachtcaptain23: You need to wrap as a hash. e.g. `contents.push({ ... })`

app/controllers/admin/partners_controller.rb
dlipeles: This is better, thanks Al @yachtcaptain23

app/services/manual_payout_report_publisher_includer.rb
dlipeles: I used naming conventions similar to the existing codebase 

thus `manual_payout_report_publisher_includer.rb` is the manual complement of the existing class `payout_report_publisher_includer.rb`

nvonpentz: I like adding more name spacing when we can like you have described @yachtcaptain23, but would advocate for keeping the syntax of the services that since that would keep everything consistent and because I don't see the downsides of the current method (please point them out).

app/jobs/enqueue_publishers_for_payout_job.rb
```diff
@@ -2,11 +2,13 @@
...
-  def perform(should_send_notifications: false, final: true, payout_report_id: "", publisher_ids: [])
+  def perform(should_send_notifications: false, final: true, manual: false, payout_report_id: "", publisher_ids: [])
     Rails.logger.info("Enqueuing publishers for payment.")
 
     if publisher_ids.present?
       publishers = Publisher.where(id: publisher_ids)
+    elsif manual
+      publishers = Publisher.partner
     else
       publishers = Publisher.with_verified_channel
```
yachtcaptain23: My impression is that the partner should be excluded from this segment. @nvonpentz would you confirm?

app/jobs/enqueue_publishers_for_payout_job.rb
```diff
@@ -2,11 +2,13 @@
...
   queue_as :scheduler
 
-  def perform(should_send_notifications: false, final: true, payout_report_id: "", publisher_ids: [])
+  def perform(should_send_notifications: false, final: true, manual: false, payout_report_id: "", publisher_ids: [])
     Rails.logger.info("Enqueuing publishers for payment.")
 
     if publisher_ids.present?
       publishers = Publisher.where(id: publisher_ids)
+    elsif manual
+      publishers = Publisher.partner
```
yachtcaptain23: You should also scope `publisher_ids` if they're included.

app/jobs/enqueue_publishers_for_payout_job.rb
```diff
@@ -15,14 +17,22 @@ def perform(should_send_notifications: false, final: true, payout_report_id: "",
...
                                           expected_num_payments: PayoutReport.expected_num_payments(publishers))
     end
 
     publishers.find_each do |publisher|
-      IncludePublisherInPayoutReportJob.perform_later(payout_report_id: payout_report.id,
-                                                      publisher_id: publisher.id,
-                                                      should_send_notifications: should_send_notifications)
+      if manual 
+        # We can consider using a job here if n is sufficiently large 
+        ManualPayoutReportPublisherIncluder.new(publisher: publisher,
```
yachtcaptain23: Also make this a perform_later otherwise we'll have a blocking call on the view

app/models/payout_report.rb
```diff
@@ -40,7 +40,11 @@ def num_payments_to_be_paid
   def update_report_contents
     # Do not update json contents for legacy reports
     return if created_at <= LEGACY_PAYOUT_REPORT_TRANSITION_DATE
-    payout_report_hash = JsonBuilders::PayoutReportJsonBuilder.new(payout_report: self).build
+    if manual
```
yachtcaptain23: Where's `manual` defined?

app/jobs/enqueue_publishers_for_payout_job.rb
dlipeles: Fortunately the `if publisher_ids.present?` will receive any included id's 👍

app/jobs/enqueue_publishers_for_payout_job.rb
dlipeles: This could be true in the future for specified `orgs` but in the current use case, partner's will be paid out as regular publishers with regards to their channels.

app/jobs/enqueue_publishers_for_payout_job.rb
dlipeles: I have considered that as well, we should consider using a job here when `n` becomes sufficiently large.

Issue #1592, opened 2019-02-07
New landing page


karenkliu:
## Description
Update the publishers landing page so that:
- The user understands what is a publisher, the benefits to being a publisher, and how Brave Rewards works
- Recent functionality is promoted; e.g. Twitter user can be a publisher (should be scaleable)
- The page provides help and support
- Sign up is highly visible as the main CTA
- The page is responsive across all breakpoints

## Designs
![pub landing - desktop hd](https://user-images.githubusercontent.com/2388823/52677205-dfdf8d80-2ee1-11e9-8b78-29167ce527f3.jpg)

The landing page has a sticker header/navigation bar so sign up is always in view:
![pub landing - sticky header](https://user-images.githubusercontent.com/2388823/52442994-660e6500-2ad9-11e9-8ed0-d35525238365.jpg)

The social proof section is a rotating carousel of testimonials that we can switch out to keep the content fresh.
![pub landing - social proof carousel](https://user-images.githubusercontent.com/2388823/52443026-80e0d980-2ad9-11e9-94ba-1e50cb577fd0.jpg)

## Specs
The page content follows a responsive, 12-column layout grid:

![responsive layout grid](https://user-images.githubusercontent.com/2388823/52444059-2b59fc00-2adc-11e9-929d-9300962814ea.png)

The typography on the page also has responsive rules:

![typography variables](https://user-images.githubusercontent.com/2388823/52450355-7e897a00-2aef-11e9-8127-9e3889dfffb0.png)

![responsive - navigation bar](https://user-images.githubusercontent.com/2388823/52677275-2503bf80-2ee2-11e9-9e54-43557190d387.png)

![responsive - section 1](https://user-images.githubusercontent.com/2388823/52677287-2d5bfa80-2ee2-11e9-8877-aadd4b8a6109.png)

For example:
![rotating title_example](https://user-images.githubusercontent.com/2388823/52679218-369c9580-2ee9-11e9-9e63-38c58dcf321b.gif)

![responsive - section 2](https://user-images.githubusercontent.com/2388823/52446857-108b8580-2ae4-11e9-8683-97ac6ed17836.png)

![responsive - section 3](https://user-images.githubusercontent.com/2388823/52677332-554b5e00-2ee2-11e9-843e-01645ab06603.png)

![responsive - section 4](https://user-images.githubusercontent.com/2388823/52677346-5f6d5c80-2ee2-11e9-9caf-223b68a6c790.png)

[...]

PR #1679, opened 2019-03-06
Creators Landing Page

rossmoody:
Refers: https://github.com/brave-intl/publishers/issues/1592

Preview build: https://bat-publishers-staging-2.herokuapp.com/

None of the actual input/button functionality is wired up in any way. I used my best judgement for the hrefs but they can be[ reviewed in locale.](https://github.com/brave-intl/publishers/blob/2cbf0d882181cfca344e3b6e95d6480ddf7ff585/public/creators-landing/src/locale/en.js)

## Unplanned additions

1. Sign up and sign in pages were added to the mix of this PR. Though not initially in the scope, it bugged me how different they looked and we had the style foundation setup. So my proposal for this would be:

- Old: `http://publishers.basicattentiontoken.org` AND `https://brave.com/creators`
New: `new Creators landing page / creators.brave.com`
- Old: `https://publishers.basicattentiontoken.org/publishers/sign_up`
New: `new Creators Sign Up page / creators.brave.com/sign-up`
- Old: `https://publishers.basicattentiontoken.org/publishers/log_in`
New: `new Creators Sign In page / creators.brave.com/sign-in`

2. The gradient background animates. Cause why not.


corymcdonald:
Ross, this looks great! 

Some thoughts in no particular order

- Now that we have are going to Reactify the sign up and sign in pages, we'll likely want to figure out a good place to show an error message on those screens.

- Have you discussed with David or product about the renaming of the URL? I don't think it will be such a trivial change. 

- One alternative to setting up A/B testing, is to just simply look at the current matomo statistics on the funnel and then compare / contrast. If we do intend to make this non-passive with the `creators.brave.com` domain replacement then we'll need to make another issue to discussing everything that could go into that.

- Have you thought about SEO with this? I'm not quite familiar with any tactics or anything but we might want to consider that. Ideally if someone searched "Earn money with youtube" then creators website would pop up.

public/creators-landing/src/components/icons/SwipeRight.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+
+export const SwipeRightIcon = () => (
```
corymcdonald: i'd swipe right on this re-design 👉😄👉

public/creators-landing/src/locale/en.js
```diff
@@ -0,0 +1,102 @@
...
+      },
+      btn: {
+        signup: 'Sign up now'
+      }
+    },
+    signin: {
+      heading: 'Sign in to your account',
+      subhead:
+        'You will receive an email momentarily with a secure login link.',
+      input_placeholder: 'Enter your email',
```
corymcdonald: Usually javascript uses `camelCase` for variable names

public/creators-landing/src/sections/main-section/index.js
```diff
@@ -0,0 +1,192 @@
...
+          </Box>
+          <Box direction='row' pad={{ vertical: '24px' }} width='100%'>
+            <PrimaryButton
+              label={locale.main.home.btn.signup}
+              a11yTitle='Sign up'
+            />
+          </Box>
+        </Box>
+      </Container>
+      <SwoopBottom />
```
corymcdonald: I was having problems where the swoop bottom was on top of the "Sign Up" button so I couldn't click on the bottom half when my web browser was like 75% full screen. Can you investigate?

public/creators-landing/src/sections/main-section/index.js
```diff
@@ -0,0 +1,192 @@
...
+        <Box
+          direction='column'
+          pad='large'
+          margin={{ top: '80px', bottom: '60px' }}
+        >
+          <Heading
+            level='1'
+            color='white'
+            margin={{ vertical: 'small' }}
+            a11yTitle='Headline'
```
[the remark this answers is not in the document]
rossmoody: got carried away with the a11y title. no idea why. omitting.

public/creators-landing/src/locale/en.js
rossmoody: fixed

public/creators-landing/src/sections/main-section/index.js
rossmoody: great spot. swoop was on top of bottom 1/6 of button. pushed down.

rossmoody:
After meeting as a team, it was decided to move the landing page to brave.com/creators via Wordpress and punt any subdomain change until a top level domain name can be agreed upon at a future date.

corymcdonald:
need to make the built with bat a link

rossmoody:
@dlipeles Done! though right as I was committing I realized, the transition might not be quite as smooth as hoped cause doesn't login happen via link from email that generates a new tab?

dlipeles:
@rossmoody Thanks, and yes you are right, the hope would be to move to a more traditional password-based sign in, in the future. 

Then no need to jump from app to e-mail client, back to app ⛳️

PR #1692, opened 2019-03-08
Enable ruby metrics for heroku

corymcdonald:
We are constantly seeing errors in heroku and this will enable us to have a more a clear idea of what is happening


corymcdonald:
Al pointed out that New Relic has this information so there's not that much of a need to include this on Heroku.

PR #1721, opened 2019-03-18
Bugfix/401 admin

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,7 +17,7 @@ class ApplicationController < ActionController::Base
 
   rescue_from CanCan::AccessDenied do |e|
     sign_out
-    redirect_to root_url, alert: e.message, status: :authorized
+    redirect_to static_index_url, alert: e.message
   end
 
   def no_cache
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -15,6 +15,11 @@ class ApplicationController < ActionController::Base
   before_action :set_paper_trail_whodunnit
   before_action :no_cache
 
+  rescue_from CanCan::AccessDenied do |e|
+    sign_out
```
corymcdonald: I don't know if you should sign out the user if they aren't on the whitelist. This would force users who aren't on the VPN to basically have to go through the annoying sign in workflow again

yachtcaptain23: @corymcdonald unfortunately they go into a redirect loop if they're signed in. That being said, we should increase the sign in limits as it's currently too restrictive.

corymcdonald:
Going to take this over, closing this PR for right now

PR #1735, opened 2019-03-20
Automatically create uphold cards during payout

### LINKED ISSUE #1731: Automatically create default currency cards when generating payouts if needed
so we protect ourselves from a situation in which a publisher isn't paid because we can't find a card that matches their currency at payout report generation time.

nvonpentz:
Resolves #1731 

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/jobs/create_uphold_cards_job.rb
+++ b/app/jobs/create_uphold_cards_job.rb
@@ -14,10 +14,6 @@ class CreateUpholdCardsJob < ApplicationJob
                                               currency_code: default_currency).perform
     end
 
-    if default_currency != "BAT" && publisher.wallet.address.blank?
-      UpholdServices::CardCreationService.new(publisher: publisher, currency_code: "BAT").perform
-    end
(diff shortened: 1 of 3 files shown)


app/jobs/create_uphold_cards_job.rb
```diff
@@ -14,10 +14,6 @@ def perform(publisher_id:)
                                               currency_code: default_currency).perform
     end
 
-    if default_currency != "BAT" && publisher.wallet.address.blank?
-      UpholdServices::CardCreationService.new(publisher: publisher, currency_code: "BAT").perform
-    end
-
```
nvonpentz: This change probably should have happened in #1726.  The idea previously was that we should always create the BAT card when it was possible but not available.  Since we don't use available currencies anymore, we should only create a BAT card when the user requests it.

PR #1750, opened 2019-03-22
Validate Publisher Names 🏷 

### LINKED ISSUE #958: Publishers should always have names
It is possible to not supply a `name` when signing up.  As a result, some publishers are not addressed by name in the app emails.  We should add validations to the Publisher model and database constraints to prevent this.

dlipeles:
#### Features
* Add database constraints to `name` on `publishers` -> `default: "", null: false`
* Add validations on `name` on `publishers` -> Name cannot be `nil`, but it can be the empty string.
* Closes #958 

#### Notes 
* We need to run this short script in the console before merging: 
```ruby 
Publisher.where(name: nil).update_all(name: "")
```

#### How to Test
* Enter an e-mail address in the sign up page
* Before opening the received e-mail, check the publisher in `rails console`
* `name` field should be `""` 
* Try updating the name to `nil` in rails console
* Validation should prevent `nil`


dlipeles:
Ran ```Publisher.where(name: nil).update_all(name: "")``` on `bat-publishers-staging` and will merge now.

Issue #1752, opened 2019-03-22
update dashboard mobile responsiveness


rossmoody:
After a discussion on Slack about the amount of traffic coming from mobile sources, a small initiative has been started to make the UI more mobile friendly for the very immediate future. Most of the updates will be replaced via upcoming feature work but this is low hanging fruit to improve the usability.

## This is the issue for the dashboard page portion of that.

![Screen Shot on 2019-03-22 at 10:29:57](https://user-images.githubusercontent.com/29072694/54841844-80696000-4c8d-11e9-806d-9ac1905477f5.png)
![Screen Shot on 2019-03-22 at 10:29:36](https://user-images.githubusercontent.com/29072694/54841845-80696000-4c8d-11e9-8683-7b073eaf6511.png)
![Screen Shot on 2019-03-22 at 10:29:16](https://user-images.githubusercontent.com/29072694/54841846-80696000-4c8d-11e9-8b0e-8c42476bc09f.png)

PR #1753, opened 2019-03-22
dashboard styling

rossmoody:
Fixes: https://github.com/brave-intl/publishers/issues/1752

Not very optimal css but the aim was to not have it look broken and I think this accomplishes that. Had to remove the images. Their sizes were explicitly set as background images causing the very janky mobile formatting. Removing them solved 70% of the issues. 

Video of responsiveness: https://d.pr/v/RffwDw

Screen Shot on 2019-03-25 at 10:58:20
Screen Shot on 2019-03-25 at 10:58:03

Screen Shot on 2019-03-22 at 10:21:52
Screen Shot on 2019-03-22 at 10:21:36


lib/tasks/referral_code_data.rb
```diff
@@ -0,0 +1,534 @@
+
+# I am sorry for the attrocities I have commited
```
nvonpentz: nit: atrocities

corymcdonald: https://github.com/brave-intl/publishers/pull/1753/commits/f65c1548ddf5453cb34ca330f30521f6a25aef8f

rossmoody: what an epic nit

PR #1759, opened 2019-03-26
Moving sign up to new controller

### LINKED ISSUE #1736: Refactor: Move the current user authentication and registration to a separate controller
The Publishers controller assumes way too much responsibility when it comes to managing authentication and registering users. We should refactor this into a new controller which leverages more RESTful routing.

corymcdonald:
Closes #1736

This refactor might be hard to follow due to the naming conventions that were previously in place

| Previous Url   | Previous Method Name  | New Method Name  |
| -------------- | --------------------- | ---------------- |
| GET /sign_up                 | sign_up               | sign_up                       |
| POST /publishers             | create                | create                        |
| GET /log_in                  | new_auth_token        | log_in                        |
| POST /log_in                 | create_auth_token     | update                        |
| GET /expired_auth_email      | expired_auth_email    | expired_authentication_email  |
| POST /resend_auth_email      | resend_auth_email     | resend_authentication_email   |

I also removed some dead code, including `create_done`, before_action methods that were dead, etc.


app/controllers/publishers/registrations_controller.rb
```diff
@@ -0,0 +1,135 @@
...
+      flash.now[:notice] = t("publishers.registrations.create.email_already_active", email: @publisher_email)
+      render :emailed_authentication_token
+    end
+
+    def enforce_throttle(throttled:, path:)
+      @should_throttle = throttled
+      throttle_is_legit = @should_throttle ? verify_recaptcha(model: @publisher) : true
+      return if throttle_is_legit
+
+      Rails.logger.info("User has been throttled")
```
yachtcaptain23: Doesn't identify user... which isn't useful. Perhaps we could include some info on who it is or publish in Sentry.

corymcdonald: True, but unfortunately there's not really a good way of determining what the user is because the user hasn't signed in. IP Addresses shouldn't be logged. I was just hoping to get a good idea on how many users are actually being hit by this enforce throttle. 🤔 

I don't think we should use sentry for things like this, I'd like to reserve sentry for only reporting errors

PR #1760, opened 2019-03-27
Manual Report Download Fix 🔧 

dlipeles:
#### Summary 
There was a small issue in downloading manual reports, the existing code would overwrite the `authority` field with the currently signed in `admin`. For `manual reports` we want the `authority` field to display the admin that `finalized the invoice`. 

#### Features
* Manual report `authority` is not overwritten by the currently signed in `admin`
* Manual report `authority` field now displays the `e-mail address` instead of the `name` 

#### Notes
* For regular payout reports, the default `authority` field is null, it is set when the `download` button is clicked. 

#### How to Test
* As an `admin` request an `invoice` on a `partner`
* Finalize the `invoice`
* Log in as a different `admin`
* Generate a `manual payout report`, ensure the `authority` field is the first `admin` 
* Generate a `regular payout report`, ensure the `authority` field is the current `admin`

### DIFF
--- a/app/models/json_builders/manual_payout_report_json_builder.rb
+++ b/app/models/json_builders/manual_payout_report_json_builder.rb
@@ -12,7 +12,7 @@ class JsonBuilders::ManualPayoutReportJsonBuilder
           "altcurrency" => "BAT",
           "probi" => potential_payment.amount.to_s,
           "fees" => potential_payment.fees.to_s,
-          "authority" => Publisher.find(potential_payment.finalized_by_id).to_s,
+          "authority" => Publisher.find(potential_payment.finalized_by_id).email,
           "transactionId" => potential_payment.payout_report_id.to_s,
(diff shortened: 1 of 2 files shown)


app/models/json_builders/manual_payout_report_json_builder.rb
```diff
@@ -12,7 +12,7 @@ def build
           "altcurrency" => "BAT",
           "probi" => potential_payment.amount.to_s,
           "fees" => potential_payment.fees.to_s,
-          "authority" => Publisher.find(potential_payment.finalized_by_id).to_s,
+          "authority" => Publisher.find(potential_payment.finalized_by_id).email,
```
corymcdonald: Could it be possible that the email could be nil?

dlipeles: I  would hope not, or we have a bigger problem than a nil error 😛 

The e-mail belongs to the admin that finalized the invoice, so it has to exist before it reaches this code.

PR #1761, opened 2019-03-27
Add note on dashboard for when payout is in progress

### LINKED ISSUE #1594: add a additional state for "Next Deposit Date:" display - "Payout In Progress"
when performing the monthly payout we often get user confusion as they may see their balances paid out in part, without the balance on their dashboard being updated. this is an artifact of the payout procedure.

it might be helpful if we were to have an alternate state for the "Next Deposit Date:" bit of text - whereby an admin can set a payout as "in progress" so that users know payout is ongoing

nvonpentz:
Resolves #1594 

image

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/admin/payout_reports_controller.rb
+++ b/app/controllers/admin/payout_reports_controller.rb
@@ -56,6 +56,12 @@ class Admin::PayoutReportsController < AdminController
     redirect_to admin_payout_reports_path, flash: { alert: "Eyeshade responded with a 400 🤷‍️" }
   end
 
+  def toggle_payout_in_progress
+    payout_status = Rails.cache.fetch('payout_in_progress')
+    Rails.cache.write('payout_in_progress', !payout_status)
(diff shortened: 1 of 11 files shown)


nvonpentz:
Cool, I will be adding support for a flag and will also remove 'Next deposit date..' text when payout is in progress.

For the flag it sounds like we want something that's automatically set at the beginning of the month then manually unset when the payout is completed.  Sounds like redis is a good choice.

corymcdonald:
Do you think we should separate out the pending amount and the amount that the user is going to get paid? I think that could be confusing as their balance could increment but they wouldn't get paid for that amount

nvonpentz:
That could be useful.  The publisher may infer on there own that the donations received while the payout is in progress will be deposited next month.

For this next month's payout I think this should suffice.

nvonpentz:
Updated dashboard looks like
image

PR #1762, opened 2019-03-27
Do not include publishers with missing addresses in payout report

### LINKED ISSUE #1685: Add presence of uphold card id (i.e. address) to potential payment to_be_paid scope
Some transactions without addresses made it into the payout report json.

Since this is a potential way a publisher might not be paid, let's include it in the payout report status on the admin per publisher page.

nvonpentz:
Resolves #1685 

A missing address causes the transaction to fail at execution time.  However that failure serves as an alert that there may be a bug with us or uphold, which we would likely miss if this pr goes in as is.

I think we may need one more component to this pr that makes sure we're aware if many people didn't make it into the payout report.  Maybe another column on the payout report page displaying # payments with of missing addresses for unknown reasons would do the job.

Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/views/admin/payout_reports/index.html.slim
+++ b/app/views/admin/payout_reports/index.html.slim
@@ -35,7 +35,7 @@ table.display.table.table-bordered.table-striped
         td = report.manual ? "Yes" : "No"
         td = report.expected_num_payments - report.num_payments
         td = report.num_payments_to_be_paid
-        td = report.potential_payments.to_be_paid.where(address: "").count
+        td = report.potential_payments.to_be_paid.unscope(where: :address).count
         td = "#{'%.2f' % (report.amount.to_d / 1E18)} BAT"
         td = form_tag refresh_admin_payout_report_path(report.id), method: :patch do
              = submit_tag "refresh", class: "btn btn-info"
--- a/app/views/admin/payout_reports/index.html.slim
+++ b/app/views/admin/payout_reports/index.html.slim
@@ -33,7 +33,7 @@ table.display.table.table-bordered.table-striped
         td = link_to report.id[0..6], admin_payout_report_path(report.id)
         td = report.final ? "Final" : "Temp"
         td = report.manual ? "Yes" : "No"
-        td = report.expected_num_payments - report.num_payments
+        td = "#{report.expected_num_payments} - #{report.num_payments}"
         td = report.num_payments_to_be_paid
         td = report.potential_payments.to_be_paid.unscope(where: :address).count
         td = "#{'%.2f' % (report.amount.to_d / 1E18)} BAT"
--- a/app/views/admin/payout_reports/index.html.slim
+++ b/app/views/admin/payout_reports/index.html.slim
@@ -20,45 +20,32 @@ table.display.table.table-bordered.table-striped
     th ID
     th Type
     th Manual
-    th Expected # of payments
-    th Total # of payments
+    th Expected - actual # of payments
     th # Payments to be paid
+    th # Legit payments missing addresses
     th Amount
-    th Fees
     th Refresh JSON
     th Download
   tbody
     - @payout_reports.each do |report|
       tr.gradeX
         td = report.created_at.strftime("%B %d, %Y, %H:%M")
-        td = link_to report.id, admin_payout_report_path(report.id)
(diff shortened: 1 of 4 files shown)


app/views/admin/publishers/show.html.slim
```diff
@@ -85,19 +85,23 @@
...
+                .db-field = "Reauthorization status"
                 .db-value = @potential_referral_payment.reauthorization_needed
             .db-info-row
-              .db-field = "Was Uphold member"
+              .db-field = "Uphold membership stattus"
               .db-value = @potential_referral_payment.uphold_member
             - if @potential_referral_payment.suspended
               .db-info-row
-                .db-field = "Was suspended"
+                .db-field = "Suspendension status"
```
yachtcaptain23: Spelling

nvonpentz: Suspendension lol

dlipeles:
LGTM

app/views/admin/payout_reports/index.html.slim
```diff
@@ -20,45 +20,32 @@ table.display.table.table-bordered.table-striped
...
         td = report.created_at.strftime("%B %d, %Y, %H:%M")
-        td = link_to report.id, admin_payout_report_path(report.id)
+        td = link_to report.id[0..6], admin_payout_report_path(report.id)
         td = report.final ? "Final" : "Temp"
         td = report.manual ? "Yes" : "No"
-        td = report.expected_num_payments
-        td = report.num_payments
+        td = report.expected_num_payments - report.num_payments
         td = report.num_payments_to_be_paid
+        td = report.potential_payments.to_be_paid.where(address: "").count
```
yachtcaptain23: You need to re-write this since `to_be_paid` contains `where.not(address: "")`

app/views/admin/payout_reports/index.html.slim
```diff
@@ -20,45 +20,32 @@ table.display.table.table-bordered.table-striped
...
     - @payout_reports.each do |report|
       tr.gradeX
         td = report.created_at.strftime("%B %d, %Y, %H:%M")
-        td = link_to report.id, admin_payout_report_path(report.id)
+        td = link_to report.id[0..6], admin_payout_report_path(report.id)
         td = report.final ? "Final" : "Temp"
         td = report.manual ? "Yes" : "No"
-        td = report.expected_num_payments
-        td = report.num_payments
+        td = report.expected_num_payments - report.num_payments
```
yachtcaptain23: I think this arithmetic is evaluated... do we really want this? IMO we should do "#{report.expected_num_payments} - #{report.num_payments}"

app/views/admin/payout_reports/index.html.slim
nvonpentz: Thanks, fixed.

```
irb(main):001:0> PotentialPayment.to_be_paid.to_sql
=> "SELECT \"potential_payments\".* FROM \"potential_payments\" WHERE \"potential_payments\".\"uphold_status\" = 'ok' AND \"potential_payments\".\"reauthorization_needed\" = FALSE AND \"potential_payments\".\"uphold_member\" = TRUE AND \"potential_payments\".\"suspended\" = FALSE AND (amount::numeric > 0) AND \"potential_payments\".\"address\" != '' AND \"potential_payments\".\"address\" IS NOT NULL"
irb(main):002:0> PotentialPayment.to_be_paid.unscope(where: :address).to_sql
=> "SELECT \"potential_payments\".* FROM \"potential_payments\" WHERE \"potential_payments\".\"uphold_status\" = 'ok' AND \"potential_payments\".\"reauthorization_needed\" = FALSE AND \"potential_payments\".\"uphold_member\" = TRUE AND \"potential_payments\".\"suspended\" = FALSE AND (amount::numeric > 0)"
```

app/views/admin/payout_reports/index.html.slim
nvonpentz: I care about the difference between expected and actual number of payments (if it's really high then there's a bug), but if others want to see the actual numbers, that's cool too.  I'm just trying to free up space on the table.

PR #1763, opened 2019-03-27
Make top balances more usable

yachtcaptain23:
Submitter Checklist:

- [ ] Tagged reviewers.

Test Plan:

Reviewer Checklist:

Tests

Security:

### DIFF
--- a/app/controllers/admin/stats/top_balances_controller.rb
+++ b/app/controllers/admin/stats/top_balances_controller.rb
@@ -12,6 +12,15 @@ class Admin::Stats::TopBalancesController < AdminController
     if @result.present?
       @result = @result.take(@limit)
     end
+
+    if params[:type] == Eyeshade::TopBalances::CHANNEL
+      @table_identifier = "youtube#channel"
+      @youtube_channels = YoutubeChannelDetails.joins(:channel).where(youtube_channel_id: @result.select { |entry| entry["account_id"].split(":")[0] == @table_identifier }.map { |entry| entry["account_id"].split(":")[1]}).pluck("youtube_channel_id, channels.id, title").map { |row| {row[0] => [row[1], row[2]]}}.reduce(:merge)
+      @site_channels = SiteChannelDetails.joins(:channel).where(brave_publisher_id: @result.map { |entry| entry["account_id"] }).pluck("brave_publisher_id, channels.id").to_h
+    elsif params[:type] == Eyeshade::TopBalances::OWNER
+      @publisher_emails = Publisher.where(id: @result.map { |entry| entry["account_id"].split(":")[1] }).pluck(:id, :email).to_h
+    end
+
     @type = params[:type]
   end
 end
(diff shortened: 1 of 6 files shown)


app/views/admin/stats/top_balances/_channels.html.slim
```diff
@@ -0,0 +1,20 @@
+- table_identifier = "youtube#channel"
```
corymcdonald: IMO These things should not be in the view

yachtcaptain23: It's localized logic. There's no existing constant elsewhere unless I do `YoutubeChannelDetails::YOUTUBE_PREFIX.chomp`. Thoughts?

corymcdonald: No, sorry I mean querying the database for these things. It violates the separation of MVC and it can make finding queries a little harder

app/controllers/admin/stats/top_balances_controller.rb
```diff
@@ -12,6 +12,15 @@ def index
     if @result.present?
       @result = @result.take(@limit)
     end
+
+    if params[:type] == Eyeshade::TopBalances::CHANNEL
+      @table_identifier = "youtube#channel"
+      @youtube_channels = YoutubeChannelDetails.joins(:channel).where(youtube_channel_id: @result.select { |entry| entry["account_id"].split(":")[0] == @table_identifier }.map { |entry| entry["account_id"].split(":")[1]}).pluck("youtube_channel_id, channels.id, title").map { |row| {row[0] => [row[1], row[2]]}}.reduce(:merge)
```
corymcdonald: This is really hard to read on one line, can you separate out similar to this?

```ruby
      account_ids = @result.select { |entry| entry["account_id"].split(":")[0] == @table_identifier }.map { |entry| entry["account_id"].split(":")[1]}
      @youtube_channels = YoutubeChannelDetails.joins(:channel)
        .where(youtube_channel_id: account_ids)
        .pluck("youtube_channel_id, channels.id, title")
        .map { |row| {row[0] => [row[1], row[2]]}}
        .reduce(:merge)
```

app/views/admin/stats/top_balances/_channels.html.slim
```diff
@@ -0,0 +1,18 @@
...
+    table.display.table.table-bordered.table-striped.dynamic-table id="dynamic-table"
+      tr
+        th Account ID
+        th Balance
+      tbody
+        - @result.each do |entry|
+          tr.gradeX
+            - if entry["account_id"].starts_with?(@table_identifier)
+              - youtube_channel_id = entry["account_id"].split(":")[1]
+              td = link_to @youtube_channels[youtube_channel_id]&.dig(1), admin_publishers_path(@youtube_channels[youtube_channel_id]&.dig(0))
```
corymcdonald: This should be `admin_publisher_path` same with below. I had some problems locally with not having the latest database so I just wrote some code like this
```
              - youtube_channel = @youtube_channels[entry["account_id"].split(":")[1]]
              td
                - if youtube_channel.present?
                  = link_to(youtube_channel&.dig(1), admin_publisher_path(youtube_channel&.dig(0)))
                - else
                  = entry["account_id"]
                end
```

PR #1780, opened 2019-04-04
Dashboard V2

rossmoody:
General consistency, simplification, responsiveness and usability updates to the CSS of the home dashboard Creators view.

**Start**
screencapture-localhost-3000-publishers-home-2019-04-04-13_58_18

**Round 1** 
The icons in the card sections are placed via background image. It's almost impossible to format the cards with any type of responsive rules with the way they are positioned and so they had to come out. 

screencapture-localhost-3000-publishers-home-2019-04-03-22_31_08

**Round 2**
This has a change to the background color. Somewhat subjective but I do think there is an accessibility and readability consideration to be made for having a loud gradient in the background of the dashboard that is likely to become dense with information. If we can find another example of a dashboard UI that uses a gradient (especially in a warm hue) as the background color maybe we could leverage the approach. For now I would stick to a neutral.   

Also got rid of the red buttons. Red is also exclusive to destructive intent. i.e. this button deletes something. 

screencapture-localhost-3000-publishers-home-2019-04-04-12_55_16

**Round 3**
More SASS consistency and general polish. 

screencapture-localhost-3000-publishers-home-2019-04-04-14_59_26


app/assets/stylesheets/theme/main-content.scss
```diff
@@ -3,7 +3,6 @@
   z-index: 0;
   padding-top: $spacer;
   padding-bottom: $spacer;
-  background: #eaedef;
-  background: $braveGradient-LegacyPink;
+  background: #d0d2dc;
```
dlipeles: This might be the happiest day of my life. Can we try this color out `#eceff1`

dlipeles: This might be the happiest day of my life. Can we try this color out `#eceff1`

dlipeles:
I am thrilled that we are taking a step in this direction, I think the end goal would be an app with a solid white background (`#ffffff`) or very light (`#fafafa`), they typically look great in the browser and on phones. 

Truthfully we'd have to make changes to the content so that it pops more against the white background. I'd be willing to pitch in effort and I'm sure @corymcdonald would too if you think this is possible, keep us posted @rossmoody 



There are some great case studies that I have been looking at.

[Coinbase](Coinbase.com)

Screen Shot 2019-04-04 at 4 19 34 PM

[Airbnb](Airbnb.com)
Screen Shot 2019-04-04 at 4 38 07 PM

[Earn](Earn.com)
Screen Shot 2019-04-04 at 4 40 49 PM

rossmoody:
I agree. Being more in line with standard dashboard aesthetic when it comes to the app seems like a necessity here. In particular if we ever want to offer a dark version of any kind.

Screen Shot on 2019-04-04 at 13:45:45

dlipeles:
@rossmoody Maybe the best time to put this in action would be as part of the `V3 Publishers UI`, let's meet up and plan it during bravetogether 👍

app/assets/stylesheets/theme/main-content.scss
rossmoody: In Round 4

rossmoody:
This is sort of addictive styling updating but I think maybe we should reign it in here soon. This wasn't meant to be an overhaul - just a general finesse update so it didn't look broken on mobile and the plan (I believe) is this will all be deprecated soon anyway so might be best to just sew this up as is.

yachtcaptain23:
Looks really good!

Would you be able to fix the bottom two issues? Those are the ones that stood out from the testing I've done.

Screen Shot 2019-04-05 at 9 40 26 PM
Screen Shot 2019-04-05 at 9 41 02 PM

rossmoody:
Last commit fixes: 

Screen Shot on 2019-04-08 at 13:28:57
Screen Shot on 2019-04-08 at 13:28:25
Screen Shot on 2019-04-08 at 13:28:04

corymcdonald:
on the width of 414 it kinda looks weird

corymcdonald:
Seems like the React code might be referencing an old font-family

corymcdonald:
image
This is probably my bad, but when you first verify a site it shows a little icon that pops up

rossmoody:
Agree about the breakpoint. It's global bootstrap breakpoint styling though driving that. I'd prefer to not go noodling with those.

Can't seem to replicate that header font-family problem. Mine is rendering ok. Added an explicit Poppins to the `Text` component anyway.

It's not the prettiest but it's legible. 
Screen Shot 2019-04-09 at 11 18 00 AM

app/views/publishers/home.html.slim
```diff
@@ -86,31 +87,27 @@ script type="text/html" id="confirm_default_currency_modal_wrapper"
...
                   = t("promo.dashboard.add_channel")
       - elsif current_publisher.channels.empty?
-        .col
+        .col.bot-marg
           .add-channel-cta
             = t ".add_channel_cta"
     .row
       - unless current_publisher.excluded_from_payout?
-        .col
+        .col.bot-marg
```
corymcdonald: Can we define this and the column below it as `col-md`?

app/views/publishers/home.html.slim
```diff
@@ -86,31 +87,27 @@ script type="text/html" id="confirm_default_currency_modal_wrapper"
...
                       = publisher_referral_totals[PromoRegistration::FINALIZED]
-                .promo-flex-row.promo-flex-row-sub
-                  =t("promo.dashboard.check_statement")
             - else
               .promo-panel-item.promo-panel-item-alert
                 #promo-panel-alert
                   = t("promo.dashboard.add_channel")
       - elsif current_publisher.channels.empty?
-        .col
+        .col.bot-marg
```
corymcdonald: Let's not use ".bot-marg", instead let's use the spacing utilities defined by bootstrap itself 
https://getbootstrap.com/docs/4.0/utilities/spacing/

`mb-1` or `mb-2`

PR #1805, opened 2019-04-10
Rate limit on IP for 2FA removal

yachtcaptain23:
## Pull Request Name

#### Features

#### How To Test

1. Step one
2. Step two
3. Step three

#### Pull Request Checklist

### DIFF
--- a/config/initializers/rack-attack.rb
+++ b/config/initializers/rack-attack.rb
@@ -65,6 +65,20 @@ class Rack::Attack
     end
   end
 
+  # Throttle send 2fa disable emails for an IP address
+  throttle("request_two_factor_authentication_removal/publisher_id", limit: 2, period: 24.hours) do |req|
+    if req.path == "/publishers/request_two_factor_authentication_removal" && req.post?
+      req.ip
+    end
+  end
+
+  # Throttle confirm 2fa disable emails for an IP address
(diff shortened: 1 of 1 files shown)


config/initializers/rack-attack.rb
```diff
@@ -65,6 +65,20 @@ class Rack::Attack
...
+    end
+  end
+
+  # Throttle confirm 2fa disable emails for an IP address
+  throttle("confirm_two_factor_authentication_removal/publisher_id", limit: 2, period: 24.hours) do |req|
+    if req.path == "/publishers/confirm_two_factor_authentication_removal" && req.get?
+      req.ip
+    end
+  end
+
```
dlipeles: Could we throttle the `cancel_two_factor_authentication_removal` route also 

```ruby
  # Throttle cancel 2fa disable emails for an IP address
  throttle("cancel_two_factor_authentication_removal/publisher_id", limit: 2, period: 24.hours) do |req|
    if req.path == "/publishers/cancel_two_factor_authentication_removal" && req.get?
      req.ip
    end
  end
```

yachtcaptain23: I don't see a security reason to add this. What are your thoughts?

dlipeles: Seems to me like we should for the same reason as  `confirm_two_factor_authentication_removal`, they are both publicly available routes that could be spammed by an attacker

PR #1810, opened 2019-04-11
Minor fixes for invoices

corymcdonald:
#### Features

Due to the settlement tools removing the `documentId` attribute from the list we must find by the owner and the amount to mark things as paid.

Once an invoice was paid users were able to click "Edit" which was a mistake

#### How To Test

1. Create a partner
2. Create an invoice
3. Mark that invoice as ready to be paid out
4. Copy the publisher id and the amount put it in the JSON below
5. Upload this settlement report JSON file
```json
[
  {
    "altcurrency": "BAT",
    "authority": "user@brave.com",
    "amount": "4321",
    "commission": "0",
    "currency": "BAT",
    "address": "redacted",
    "owner": "publishers#uuid:d2d9c735-522c-484c-90a6-e2da78e12c99",
    "fees": "0",
    "probi": "redacted",
    "hash": "redacted",
    "publisher": "",
    "signedTx": "",
    "status": "completed",
    "transactionId": "redacted",
    "fee": "0",
    "type": "manual",
    "validUntil": "2019-04-09T01:15:21.434437557Z",
    "note": ""
  }
]
```

### DIFF
--- a/app/controllers/admin/payout_reports_controller.rb
+++ b/app/controllers/admin/payout_reports_controller.rb
@@ -1,4 +1,6 @@
 class Admin::PayoutReportsController < AdminController
+  MANUAL = "manual"
+
   def index
     @payout_reports = PayoutReport.all.order(created_at: :desc).paginate(page: params[:page])
   end
@@ -38,13 +40,21 @@ class Admin::PayoutReportsController < AdminController
     Eyeshade::Publishers.new.create_settlement(body: json)
 
     json.each do |entry|
-      next unless entry["documentId"]
-      invoice = Invoice.find_by(id: entry["documentId"])
+      next unless entry["type"] == MANUAL && entry["owner"] && entry["amount"]
+
+      partner_id = entry["owner"].sub("publishers#uuid:", "")
+
+      invoice = Invoice.where(
+        partner_id: partner_id,
+        finalized_amount: entry["amount"],
+        status: Invoice::IN_PROGRESS
+      ).order(:created_at).first
+
       next unless invoice.present?
 
       invoice.update(
         payment_date: Date.today,
-        status: "paid",
+        status: Invoice::PAID,
         paid_by: current_publisher
       )
     end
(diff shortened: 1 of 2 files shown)


app/controllers/admin/payout_reports_controller.rb
```diff
@@ -38,13 +40,21 @@ def upload_settlement_report
     Eyeshade::Publishers.new.create_settlement(body: json)
 
     json.each do |entry|
-      next unless entry["documentId"]
-      invoice = Invoice.find_by(id: entry["documentId"])
+      next unless entry["type"] == MANUAL && entry["owner"] && entry["amount"]
+
+      partner_id = entry["owner"].sub("publishers#uuid:", "")
```
yachtcaptain23: Preferably do `partner_id = entry["owner"].sub(Publisher::OWNER_PREFIX, "")`

app/controllers/admin/payout_reports_controller.rb
```diff
@@ -38,13 +40,21 @@ def upload_settlement_report
...
+
+      invoice = Invoice.where(
+        partner_id: partner_id,
+        finalized_amount: entry["amount"],
```
yachtcaptain23: Is there a more elegant way to look this up besides the amount? What happens if there are duplicates? (I realize this is unlikely, but it might bite us in the ass a year from now)

corymcdonald: Unfortunately not, I think this will be fine because we're basing it on the specific partner, filtering it down to finalized invoices,  and the amount will be the exact same. I think the chaces of duplicates are slim

app/controllers/admin/payout_reports_controller.rb
corymcdonald: Updated! https://github.com/brave-intl/publishers/pull/1810/commits/959badce161a9a8a5f0c0833566b134f1a35a26c

PR #1812, opened 2019-04-11
Add channel_stats to potential_payments

### LINKED ISSUE #1795: Add channel stats to potential payments
And save the stats at payout report generation time.  This way we keep a historical record of channel stats at for each potential payment.

Probably fine to just save as a column with type jsonb, but if we plan to further implement a scraper we will probably want to separate out into it's own table.

Related to https://github.com/brave-intl/publishers/pull/1774

nvonpentz:
Closes #1795

### DIFF
--- a/app/services/payout_report_publisher_includer.rb
+++ b/app/services/payout_report_publisher_includer.rb
@@ -22,6 +22,7 @@ class PayoutReportPublisherIncluder < BaseService
     probi = wallet.referral_balance.amount_probi # probi = balance
     publisher_has_unsettled_balance = probi.to_i.positive?
 
+    # Create the referral payment for the owner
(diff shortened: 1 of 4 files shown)


app/services/payout_report_publisher_includer.rb
```diff
@@ -58,7 +59,8 @@ def perform
                                 reauthorization_needed: reauthorization_needed,
                                 uphold_member: uphold_member,
                                 suspended: suspended,
-                                uphold_id: uphold_id)
+                                uphold_id: uphold_id,
+                                channel_stats: channel.details.stats)
```
yachtcaptain23: Would you do type check to make sure it's `{}` and not `'{}'`?

PR #1813, opened 2019-04-12
Add verdict and notes to potential_payment

### LINKED ISSUE #1801: Add verdict to potential_payments
Not required for this issue, but we should also expose the ability to update this via API.

nvonpentz:
Resolves #1801

### DIFF
--- a/db/migrate/20190412130309_add_verdict_and_notes_to_potential_payments.rb
+++ b/db/migrate/20190412130309_add_verdict_and_notes_to_potential_payments.rb
@@ -1,0 +1,6 @@
+class AddVerdictAndNotesToPotentialPayments < ActiveRecord::Migration[5.2]
+  def change
+    add_column :potential_payments, :verdict, :text
+    add_column :potential_payments, :notes, :text
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20190412130309_add_verdict_and_notes_to_potential_payments.rb
```diff
@@ -0,0 +1,6 @@
+class AddVerdictAndNotesToPotentialPayments < ActiveRecord::Migration[5.2]
+  def change
+    add_column :potential_payments, :verdict, :text
```
corymcdonald: Think this might be better suited to be a `string`? I don't think we'd need the verdict to be greater than 255 characters

db/schema.rb
```diff
@@ -282,6 +282,9 @@
     t.string "uphold_id"
     t.uuid "invoice_id"
     t.uuid "finalized_by_id"
+    t.jsonb "channel_stats", default: {}
```
corymcdonald: don't think this is supposed to be in this migration 😄

db/migrate/20190412130309_add_verdict_and_notes_to_potential_payments.rb
nvonpentz: I think we should be using `text` pretty much always as there is no difference in performance or storage

PR #1815, opened 2019-04-12
Prevent chaining of channel transfers

corymcdonald:
#### Features
This is the bug, where we saw multiple channels transfer to more than one user. 

The problem is that when the user was transferring their channel, we found the first Channel with the same id. However the channel that we found was not the original verified channel.

#### How To Test

1. Create 3 publisher accounts `a@a.com`, `b@b.com`, `c@c.com`
2. Add a channel to `a@a.com`
3. Sign into `b@b.com` add the same channel
4. Sign into `c@c.com`, add the same channel
5. Sign back into `b@b.com` channel should be gone.
6. Reject the transfer for `c@c.com`, channel should be successfully rejected.

### DIFF
--- a/app/models/json_builders/channels_json_builder.rb
+++ b/app/models/json_builders/channels_json_builder.rb
@@ -26,7 +26,7 @@ class JsonBuilders::ChannelsJsonBuilder
   def build
     channels = []
 
-    [Channel.verified.site_channels, Channel.youtube_channels, Channel.twitch_channels, Channel.twitter_channels].each do |verified_channels|
+    [Channel.verified.site_channels, Channel.verified.youtube_channels, Channel.verified.twitch_channels.where(verified: true), Channel.verified.twitter_channels].each do |verified_channels|
       verified_channels.find_each do |verified_channel|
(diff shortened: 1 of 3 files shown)


app/models/json_builders/channels_json_builder.rb
```diff
@@ -26,7 +26,7 @@ def initialize
   def build
     channels = []
 
-    [Channel.verified.site_channels, Channel.youtube_channels, Channel.twitch_channels, Channel.twitter_channels].each do |verified_channels|
+    [Channel.verified.site_channels, Channel.verified.youtube_channels, Channel.verified.twitch_channels.where(verified: true), Channel.verified.twitter_channels].each do |verified_channels|
```
yachtcaptain23: Duplicate `where(verified: true)`

corymcdonald: https://github.com/brave-intl/publishers/pull/1815/commits/d19a1c6131b88cc598950ceb1f92a7ce7c2aec0b

PR #1824, opened 2019-04-18
Patch for Partner Stats 🏷

dlipeles:
## Patch for Partner Stats

Screen Shot 2019-04-18 at 4 33 29 PM

#### Features

* Fix an issue partners were seeing with the stats on their dashboard. 
* Aggregate all stats for a referral code, instead of showing daily stats
* Add additional error handling for null or empty stats
* Ensure reporting works for partner generated campaigns/codes 

#### How To Test

1. As a `partner` create a campaign with promo codes 
2. As an `admin`, go to the referral promo page in the dashboard
3. Generate a report to update the stats
4. Ensure parity between partner stats, admin stats, and reports

#### Pull Request Checklist


app/services/promo/registrations_stats_fetcher.rb
```diff
@@ -62,6 +64,10 @@ def api_authorization_header
   end
 
   def query_string(referral_codes)
-    "?" + referral_codes.map { |referral_code| "referral_code=#{referral_code}"}.join("&")
+    query_string = "?"
```
yachtcaptain23: Preferably we use the older version since it's much more readable.

dlipeles: Sounds good, fixed.

dlipeles:
@yachtcaptain23 Thanks for reviewing, to your point about error handling. I agree with you, normally we display the error to the user in a dialogue as seen here https://github.com/brave-intl/publishers/blob/a858d27daa9ffcb5734b818a409934ab30cdf4fb/app/javascript/views/referrals/referralsInformation/editCampaignDialog/EditCampaignDialog.tsx#L108

This particular case is a bit easier, we only need to handle empty stats gracefully (it will display as a 0 if no stats collected yet). No need to dialogue the user, so I think we are okay.

app/javascript/views/referrals/referralsInformation/ReferralsInformation.tsx
```diff
@@ -344,6 +348,19 @@ function ReferralsTable(props) {
   const rows = [];
 
   props.referralCodes.forEach((referralCode, index) => {
+    let downloads = 0;
```
corymcdonald: Could you reuse the function you defined above?
`const { downloads, installs, thirtyDayUse } = processStats(referralCode)`

dlipeles: Added to my backlog.

dlipeles:
Removed the `owner_promo_registrations_stats_job.rb` and its call from `sidekiq` until we implement polymorphic codes. We'll keep the `registrar` because it does some useful things like add the publisher `id` and the code `description`. 

I swept our production db and updated the few owner codes that did exist.

PR #1825, opened 2019-04-18
Refactor Uphold and Cache KYC'd users on Uphold

### LINKED ISSUE #1739: New channels list and API that will need to be filtered to only include publishers with uphold accounts that are fully KYC'd (aka isMember)
Action items for channel list:
- [ ] New API
- [ ] Compression
- [ ] Add filters for uphold
- [ ] Add filters on ToS
- [ ] Add payout address

corymcdonald:
#### Features

* Extracted the fields that were uphold related into a separate table called `UpholdConnection`
* Cache the `isMember` and `address` fields
* Closes #1739

#### How To Test

1. Everything should be exactly the same as before
2. `isMember` and `address` are properly cached

Left to do
- [x] revert home.js back to normal 😄 
- [x] write some tests relating to caching 
- [ ] test on staging


db/migrate/20190408215416_create_uphold_connections.rb
```diff
@@ -0,0 +1,22 @@
+class CreateUpholdConnections < ActiveRecord::Migration[5.2]
+  def change
+    create_table :uphold_connections, id: :uuid, default: -> { "uuid_generate_v4()" } do |t|
+      t.string :uphold_state_token
+      t.boolean :uphold_verified, default: false
+      t.boolean :is_member, default: false
+
+      t.uuid :uphold_id
+      t.uuid :address
+      t.belongs_to :publisher, index: true, type: :uuid
```
yachtcaptain23: Would you add a uniqueness index to this?

db/schema.rb
```diff
@@ -369,10 +369,12 @@
     t.uuid "default_site_banner_id"
     t.boolean "default_site_banner_mode", default: false, null: false
     t.uuid "uphold_id"
+    t.uuid "uphold_connection_id"
```
yachtcaptain23: Unused code

app/controllers/publishers_controller.rb
```diff
@@ -373,6 +372,11 @@ def authenticate_via_token
         flash[:alert] = t(".email_confirmed", email: publisher.email)
       end
 
+      if publisher.uphold_connection.blank?
+        publisher.uphold_connection = UpholdConnection.new(publisher: publisher)
+        publisher.save
+      end
```
dlipeles: Consider using the .create(!) method

app/controllers/publishers_controller.rb
```diff
@@ -386,7 +390,7 @@ def authenticate_via_token
   end
 
   def create_uphold_card_for_default_currency_if_needed
-    if current_publisher.can_create_uphold_cards? &&
+    if current_publisher.uphold_connection&.can_create_uphold_cards? &&
```
dlipeles: Question, is it best practice to use `&` in situations like these, starting to see these everywhere I wonder if we should do more comprehensive checking / error catching.

app/jobs/exchange_uphold_code_for_access_token_job.rb
```diff
@@ -8,17 +8,17 @@ def perform(publisher_id:)
...
 
     if parameters
-      publisher.uphold_access_parameters = parameters
+      publisher.uphold_connection.uphold_access_parameters = parameters
       # The code acquired from https://uphold.com/authorize is only good for one request and times out in 5 minutes
       # it should now be cleared
-      publisher.uphold_code = nil
-      publisher.save!
+      publisher.uphold_connection.uphold_code = nil
+      publisher.uphold_connection.save!
```
dlipeles: Consider using .update!

app/models/uphold_connection.rb
```diff
@@ -0,0 +1,130 @@
...
+        UpholdAccountState::VERIFIED
+      end
+    elsif uphold_access_parameters.present?
+      :access_parameters_acquired
+    elsif uphold_code.present?
+      :code_acquired
+    else
+      UpholdAccountState::UNCONNECTED
+    end
+  end
```
dlipeles: Maybe this function could use some cleanup , consider something along the lines of this

lets make this a variable 

```ruby 
uphold_account_status = publisher&.wallet&.uphold_account_status&.to_sym
```

then let's put SlackMessenger in a function

```ruby
send_blocked_message if uphold_account_status == UpholdAccountState::BLOCKED
```

then 

```ruby 
if uphold_account_status == UpholdAccountState::RESTRICTED
    UpholdAccountState::RESTRICTED

elsif uphold_verified? && uphold_reauthorization_needed?
    UpholdAccountState::REAUTHORIZATION_NEEDED

elsif uphold_verified? && publisher&.wallet&.not_a_member?
    UpholdAccountState::RESTRICTED

elsif uphold_verified? && !uphold_reauthorization_needed? && !publisher&.wallet&.not_a_member?
    UpholdAccountState::VERIFIED

elsif uphold_access_parameters.present?
    :access_parameters_acquired

elsif uphold_code.present?
    :code_acquired

else
    UpholdAccountState::UNCONNECTED
end

File: app/controllers/publishers_controller.rb (hunk 4 above)
Comment (yachtcaptain23) [reply app/controllers/publishers_controller.rb]: It's more readable to do `&.` You can throw an error if the result is `nil`.

Comment (corymcdonald) [reply app/controllers/publishers_controller.rb]: I'm with you Dan, It's not necessarily good to have these nil-safe operators everywhere. Part of it is defensive coding and part of it is legacy. 

There's definitely more to refactor with this Uphold Connection so that we don't have to make sure that these things aren't nil. The way we're currently handling the uphold connections should be abstracted out to a controller so we can make sure that we would never call it on a nil uphold connection. In addition, we have code in a helper that calls a method on a model that updated a value 🤢

File: app/controllers/publishers_controller.rb (hunk 3 above)
Comment (corymcdonald) [reply app/controllers/publishers_controller.rb]: Good idea!
https://github.com/brave-intl/publishers/pull/1825/commits/2af6ad6de7c0c3bc4e7bf2fc716a670e007a4ac0

File: app/jobs/exchange_uphold_code_for_access_token_job.rb (hunk 5 above)
Comment (corymcdonald) [reply app/jobs/exchange_uphold_code_for_access_token_job.rb]: https://github.com/brave-intl/publishers/pull/1825/commits/2af6ad6de7c0c3bc4e7bf2fc716a670e007a4ac0

File: app/models/uphold_connection.rb (hunk 6 above)
Comment (corymcdonald) [reply app/models/uphold_connection.rb]: Glad you proposed this. I was debating changing it now or waiting till later. It looks really good

https://github.com/brave-intl/publishers/pull/1825/commits/ef4d1491cd420fd18872b426157f70e123dd8d91

File: db/schema.rb (hunk 2 above)
Comment (corymcdonald) [reply db/schema.rb]: I added this because in the `uphold_connection` model we call `publisher` which makes a call to the DB to get this information.

File: db/migrate/20190408215416_create_uphold_connections.rb (hunk 1 above)
Comment (corymcdonald) [reply db/migrate/20190408215416_create_uphold_connections.rb]: https://github.com/brave-intl/publishers/pull/1825/commits/0d6e0847d3e30393b4dc8d04eeb11821ae01e544

File: db/schema.rb (hunk 2 above)
Comment (yachtcaptain23) [reply db/schema.rb]: Sorry I didn't realize that you added a reference. That being said, my impression is that having rows that point at each other could result in bad state. Preferably we just establish a `belongs_to` and `has_one` in the models rather than store this in the database.

Comment (corymcdonald) [reply db/schema.rb]: Unified it here ✔️ 
https://github.com/brave-intl/publishers/pull/1825/commits/6df08b7dd73c63e4ce15c81926fa5f693979efaa

**Comment by corymcdonald** [comment]
This is ready for final review 🎉 I deployed it to staging and things seem to be good 👍

File: app/jobs/clean_stale_uphold_data_job.rb [hunk 7]
```diff
@@ -3,29 +3,29 @@ class CleanStaleUpholdDataJob < ApplicationJob
...
+    # clear uphold codes sitting for over 5 minutes
+    connections = UpholdConnection.has_stale_uphold_code
     n = 0
-    publishers.each do |publisher|
-      raise if publisher.uphold_status != :code_acquired
-      publisher.uphold_code = nil
-      publisher.save!
+    connections.each do |connection|
+      raise if connection.uphold_status != :code_acquired
+      uphold_code = nil
```
yachtcaptain23: I believe this should be `connection.uphold_code = nil`

corymcdonald: https://github.com/brave-intl/publishers/pull/1825/commits/2717f1080c4c35b859180a970396fb0c9702be5a

PR #1843, opened 2019-04-23
Feature/deleted account

yachtcaptain23:
## Pull Request Name

#### Features

#### How To Test

1. Step one
2. Step two
3. Step three

#### Pull Request Checklist


app/controllers/admin/publishers_controller.rb
```diff
@@ -22,6 +22,7 @@ def index
     if params[:two_factor_authentication_removal].present?
       @publishers = @publishers.joins(:two_factor_authentication_removal).distinct
     end
+    @publishers = @publishers.where("email IS NOT NULL or pending_email IS NOT NULL") # Don't include deleted users
```
corymcdonald: Let's try and use the ORM interface instead

```suggestion
    @publishers = @publishers.where.not(email: nil, pending_email: nil") # Don't include deleted users
```

app/models/publisher.rb
```diff
@@ -217,6 +217,10 @@ def history
     combined.map { |c| c.values.first }
   end
 
+  def deleted?
+    last_status_update.present? && last_status_update.status == PublisherStatusUpdate::DELETED
```
corymcdonald: I believe this will result in 2 db calls, probably better to use the following 
```suggestion
    last_status_update&.status == PublisherStatusUpdate::DELETED
```

app/jobs/publisher_removal_job.rb
```diff
@@ -0,0 +1,25 @@
...
+    publisher = Publisher.find_by(id: publisher_id)
+    return if publisher.suspended?
+    publisher.status_updates.create(status: PublisherStatusUpdate::DELETED)
+    ActiveRecord::Base.transaction do
+      publisher.update(email: nil)
+      publisher.update(pending_email: nil)
+      publisher.update(name: PublisherStatusUpdate::DELETED)
+      # If they're signed in, they should not longer be signed in
+      publisher.user_authentication_token.update(authentication_token_expires_at: Time.now) if publisher.user_authentication_token.present?
+      publisher.update(last_sign_in_ip: DELETED_IP_ADDRESS)
```
corymcdonald: Also `current_sign_in_ip` probably should be deleted

app/models/publisher.rb
```diff
@@ -59,12 +59,12 @@ class UpholdAccountState
   # Normalizes attribute before validation and saves into other attribute
   phony_normalize :phone, as: :phone_normalized, default_country_code: "US"
 
-  validates :email, email: { strict_mode: true }, presence: true, unless: -> { pending_email.present? }
-  validates :email, uniqueness: { case_sensitive: false }, allow_nil: true
-  validates :pending_email, email: { strict_mode: true }, presence: true, if: -> { email.blank? }
+  validates :email, email: { strict_mode: true }, presence: true, unless: -> { pending_email.present? || deleted? }
```
corymcdonald: I'm not a big fan of having a conditionals on validations because it makes it harder for our mental model to know what is being validated and when. 

Were these added because of the `PublisherRemovalJob`? Perhaps we should instead call `update_attribute` in the job because the method skips validation.

app/controllers/admin/publishers_controller.rb
yachtcaptain23: Your suggestion is an `AND` rather than `OR`. If you're able to figure out how to use this using the ORM interface, I welcome it, but the performance is still the same.

app/models/publisher.rb
yachtcaptain23: I'll do so.

app/controllers/admin/publishers_controller.rb
nvonpentz: I believe you can do `OR` with
```
@publishers.where.not(email: nil).or.@publishers.where.not(pending_email: nil)
```

app/models/publisher.rb
yachtcaptain23: Actually now that I think about it: If we want to make any changes to these deleted users, the validation would continuously fail. I realize it's messy, but it'll be a "wtf is happening" to the next developer who's trying to make a change on a deleted user. The next dev will then also need to remember to use `update_attribute`. 

We could specify individual function to skip these validations: https://richonrails.com/articles/skipping-validations-in-ruby-on-rails

Thoughts?

app/jobs/publisher_removal_job.rb
```diff
@@ -0,0 +1,28 @@
...
+    ActiveRecord::Base.transaction do
+      publisher.update(email: nil)
+      publisher.update(pending_email: nil)
+      publisher.update(name: PublisherStatusUpdate::DELETED)
+      publisher.update(phone: nil)
+      publisher.update(phone_normalized: nil)
+      # If they're signed in, they should not longer be signed in
+      publisher.user_authentication_token.update(authentication_token_expires_at: Time.now) if publisher.user_authentication_token.present?
+      publisher.update(current_sign_in_ip: DELETED_IP_ADDRESS)
+      publisher.update(last_sign_in_ip: DELETED_IP_ADDRESS)
```
dlipeles: Consider making this `nil` unless there's a reason not to.

app/controllers/admin/publishers_controller.rb
```diff
@@ -41,6 +42,12 @@ def update
     redirect_to admin_publisher_path(@publisher)
   end
 
+  def destroy
+    PublisherRemovalJob.perform_later(publisher_id: @publisher.id)
```
dlipeles: Question, should we consider making this a function? it seems like these will be one-off calls and don't need to be scheduled. Will defer to your judgement here @yachtcaptain23

yachtcaptain23: Sort of intended since I'd imagine there would be more API calls to delete more PII. By locking it all in the same thread, we can see which job succeeded and failed in Sidekiq.

app/controllers/admin/publishers_controller.rb
corymcdonald: Let's just keep it the same, that way it's more readable

app/models/publisher.rb
corymcdonald: Hmm, let's keep it the weird logic for now. Perhaps we can find a better way of handling it later

app/jobs/publisher_removal_job.rb
```diff
@@ -0,0 +1,26 @@
+class PublisherRemovalJob < ApplicationJob
+  queue_as :default
+
+  def perform(publisher_id:)
+    publisher = Publisher.find_by(id: publisher_id)
+    return if publisher.suspended?
+    publisher.status_updates.create(status: PublisherStatusUpdate::DELETED)
+    ActiveRecord::Base.transaction do
+      publisher.update(email: nil)
```
corymcdonald: Are there any view problems will having these be nil?

yachtcaptain23: There shouldn't be any problems given this will only render on admin.

Issue #1845, opened 2019-04-23
Improve homepage render speed


corymcdonald:
Using the [PageSpeed analysis by Google](https://developers.google.com/speed/pagespeed/insights/) we've found that our mobile performance is around 50% and 80% for desktop

![image](https://user-images.githubusercontent.com/5459225/56600868-57efc100-65bf-11e9-8502-e456f3c397c0.png)

This is bad and we should improve it. Since we use fastly and our site is static perhaps we can use their content to improve this. 
![image](https://user-images.githubusercontent.com/5459225/56601220-2aefde00-65c0-11e9-9709-859c7f070dfc.png)

PR #1846, opened 2019-04-23
Creator optimization

rossmoody:
Assists: https://github.com/brave-intl/publishers/issues/1845

- deleted png files in favor of webp
- added alt text for buttons and imgs
- added names to buttons


corymcdonald:
Just need to build the output on the public folder, then we'll merge

PR #1854, opened 2019-04-25
Improve landing page performance

### LINKED ISSUE #1845: Improve homepage render speed

### LINKED ISSUE #1847: Remove Vimeo logo from creators.brave.com
## Description
Bizdev wants to pull Vimeo from the landing page for now.

Current:
Screen Shot 2019-04-23 at 3 58 40 PM
Screen Shot 2019-04-23 at 3 58 46 PM

## Designs
Replace with updated logos so it looks like this:

Screen Shot 2019-04-23 at 4 00 48 PM

publogos-small

Use updated logos from this zipped folder: 
[2019-04-24_creators logos.zip](https://github.com/brave-intl/publishers/files/3114297/2019-04-24_creators.logos.zip)

corymcdonald:
## Improve landing page performance

* Gets the lighthouse speed up to an acceptable level
* Closes #1847 
* Closes #1845

#### Overview

I went through a lot of commits and deployments to ensure this was working as intended. The primary things that were changed are 

* Add a loading icon so before the React loads users on poor performing mobile devices won't just see a white screen
* Assets are being cached by CDN
* Add build script on main package.json so we will compile the homepage to use the right CDN (based on `PUBLIC_URL` env variable)
* Set fonts to be loaded after the page has already been loaded
* Defer almost all assets from being loaded so it does not block first paint
* Set a cache-control for public assets
* Add accessibility features per Chrome dev tools audit tool
* Add `Rack::Deflater` middleware to allow for GZipped responses on HTML

### Local chrome tools
| before | after |
| ------- | ---- |
|  |  |

### [Lighthouse PageSpeed](https://developers.google.com/speed/pagespeed/insights/)
| before | after |
| ------- | ---- |
|   |  |


config/environments/production.rb
```diff
@@ -43,7 +43,12 @@
   # `config.assets.precompile` and `config.assets.version` have moved to config/initializers/assets.rb
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  # config.action_controller.asset_host = "http://assets.example.com"
+  if ENV["PUBLIC_URL"].present?
```
corymcdonald: `PUBLIC_URL` is an environment variable that is consumed by Create-React-App to know where the primary assets are being hosted from. 

We updated the [Fastly configuration in this pull request](https://github.com/brave/devops/pull/902).

package.json
```diff
@@ -6,7 +6,8 @@
     "lint-js-css": "stylelint --config '.stylelintrc-jsx.json' 'app/**/*.{ts,tsx}'",
     "lint-ts": "tslint --project tsconfig.json 'app/javascript/**/*.{ts,tsx}'",
     "prettier": "prettier --check 'app/javascript/**/*.{ts,tsx}'",
-    "test": "jest"
+    "test": "jest",
+    "build": "cd public/creators-landing && yarn install && yarn build"
```
corymcdonald: Heroku's build platform will run `yarn build` during deployments and thus our creators-landing page is going to be built when we deploy to ensure that regardless if someone hasn't committed the latest built files the website will function as intended.

public/creators-landing/src/sections/main-section/index.js
```diff
@@ -26,13 +26,18 @@ const logAction = (action, value) => {
 export const MainHome = () => {
   return (
     <GradientBackground align="center">
-      <Container animation="fadeIn" role="main">
+      <Container role="main">
```
corymcdonald: Removed the animation here because it was slowing down the `First Input Delay (FID)` and `Estimated Input Latency`. The other pages fade in so I think we're good here.

app/javascript/packs/static.js
```diff
@@ -1 +1,11 @@
-import 'analytics';
+import "analytics";
+
+(function() {
+  // After the javascript has loaded request the fonts
+  let ss = document.createElement("link");
+  ss.type = "text/css";
+  ss.rel = "stylesheet";
+  ss.href =
+    "https://fonts.googleapis.com/css?family=Muli:300,400,700|Poppins:300,500";
```
dlipeles: May want to run this by the security team, my feeling is google fonts is a sneaky way for them to add tracking software to projects 😛

corymcdonald: Fixed this! :tada:
https://github.com/brave-intl/publishers/pull/1854/commits/c7b0953757d4347c6b23beb60bc5201469f95ec7

config/environments/staging.rb
```diff
@@ -39,7 +39,12 @@
   # `config.assets.precompile` and `config.assets.version` have moved to config/initializers/assets.rb
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  # config.action_controller.asset_host = "http://assets.example.com"
+  if ENV["PUBLIC_URL"].present?
+    # example: https://publishers-staging-distro.basicattentiontoken.org
+    config.public_file_server.headers = {
+      'Cache-Control' => 'public, max-age=31536000'
```
yachtcaptain23: I'm not certain this is the right configuration. My impression is that we should setup a CDN to speedup the load for assets. Would this affect all assets we upload inclusive of javascript?

corymcdonald: My original config was to actually serve all assets from publishers-distro with the following
```rb
config.action_controller.asset_host = ENV["PUBLIC_URL"]
```

However that caused issues when loading fonts. I don't think this cache control adds that much for our speed, it was just one of the recommendations that the PageSpeed utility told us that might have a benefit and seemed to increase the score a bit. 

This would impact assets we upload along with javascript. I don't think this configuration has very much impact though as the filenames receive a unique hash that vary from deploy to deploy.

corymcdonald: I'm just reverting this back to staging, no reason to keep this being a problem

PR #1881, opened 2019-05-13
Add Vimeo Integration

corymcdonald:
#### Features

- Hides behind feature flag `home?vimeo=1`
- Adds a little more robustness when adding new things
- Steals Albert's work

#### How To Test

1. Pull the keys from staging
2. Run the server and go to the homepage and pass `?vimeo=1` to the page.


app/controllers/api/v1/stats/channels_controller.rb
```diff
@@ -5,24 +5,15 @@ def index
...
-        channel_name = channel_details.name
-      when "TwitterChannelDetails"
-        channel_details = TwitterChannelDetails.find_by_id(channel.details_id)
-        channel_type = "twitter"
-        channel_name = channel_details.name
-      when "TwitchChannelDetails"
-        channel_details = TwitchChannelDetails.find_by_id(channel.details_id)
-        channel_type = "twitch"
+      else
+        channel_type = channel.details_type.gsub("ChannelDetails", "").downcase
```
dlipeles: I think I created a function for something like this way back when, check out https://github.com/brave-intl/publishers/blob/a5918610172eb28fea0c71e45e2438d9f4f0c87c/app/models/channel.rb#L306-L308

app/helpers/promos_helper.rb
```diff
@@ -60,16 +60,11 @@ def base_referral_url
...
-      "#{channel.publication_title.upcase} #{t("promo.shared.on_youtube")}"
-    when "TwitchChannelDetails"
-      "#{channel.publication_title.upcase} #{t("promo.shared.on_twitch")}"
-    when "TwitterChannelDetails"
-      "#{channel.publication_title.upcase} #{t("promo.shared.on_twitter")}"
     when "SiteChannelDetails"
       "#{channel.publication_title.upcase}"
     else
-      raise
+      integration_name = channel.details_type.gsub("ChannelDetails", "").downcase
```
dlipeles: Might be able to leverage it here too https://github.com/brave-intl/publishers/blob/a5918610172eb28fea0c71e45e2438d9f4f0c87c/app/models/channel.rb#L306-L308

app/controllers/api/v1/stats/channels_controller.rb
dlipeles: Although `type_display` probably wasn't the best name for it, feel free to rename (just search for any instances if you do)

PR #1907, opened 2019-05-23
2FA Removal Patch 🏷 

dlipeles:
#### Features

Handles the edge case where a user is attempting both a `Two Factor Removal` and `Channel Transfer` at the same time. 

Fixes - 
https://sentry.io/organizations/brave-software/issues/1016272053/?project=225116&query=transaction%3A%22Sidekiq%2FTwoFactorAuthenticationRemovalJob%22&statsPeriod=14d

#### How To Test

1. Try performing a 2FA Removal on an account with a Channel that's also contesting

#### Pull Request Checklist

### DIFF
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -256,7 +256,7 @@ class Publisher < ApplicationRecord
     )
   end
 
-  def is_registered_for_2fa_removal?
+  def registered_for_2fa_removal?
     two_factor_authentication_removal.present?
   end
 
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -256,6 +256,10 @@ class Publisher < ApplicationRecord
     )
   end
 
+  def is_registered_for_2fa_removal?
+    two_factor_authentication_removal.present?
+  end
+
   # Remove when new dashboard is finished
   def in_new_ui_whitelist?
     partner?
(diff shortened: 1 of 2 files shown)


app/models/publisher.rb
```diff
@@ -256,6 +256,10 @@ def register_for_2fa_removal
     )
   end
 
+  def is_registered_for_2fa_removal?
```
corymcdonald: I'm finding this logic kind of hard to read in the delete channels job because when the publisher is registered for 2fa it won't raise an error?

```ruby 
def not_registered_for_2fa_removal?
  two_factor_authentication_removal.blank?
end
```

yachtcaptain23: IMO the naming could be `registered_for_2fa_removal?` is preferable per rubocop style guide https://github.com/rubocop-hq/ruby-style-guide (I realize it's semi-arbitrary as Airbnb's Ruby style guide doesn't care but Rubocop does):

```
Avoid prefixing predicate methods with the auxiliary verbs such as is, does, or can. These words are redundant and inconsistent with the style of boolean methods in the Ruby core library, such as empty? and include?. [link]

# bad
class Person
  def is_tall?
    true
  end

  def can_play_basketball?
    false
  end

  def does_like_candy?
    true
  end
end

# good
class Person
  def tall?
    true
  end

  def basketball_player?
    false
  end

  def likes_candy?
    true
  end
end
```

corymcdonald: 👍

dlipeles: 👍 will update to `registered_for_2fa_removal?`

PR #1913, opened 2019-05-24
Stats for KYC'd Uphold accounts 🏷 

### LINKED ISSUE #1524: Track KYC'd Uphold verified accounts in Publisher.statistical_totals
`isMember` is fetched from Eyeshade which would result in a nasty N+1 query when we query. Going to elaborate on details later during scrum.

dlipeles:
#### Features

Opens up `/api/v1/stats/publishers/channel_and_kyc_uphold_and_email_verified_signups_per_day` which tracks publishers who have 

1. Verified their `e-mail`
2. Added a verified channel 
3. Added an uphold account
4. KYC'd their uphold account

Closes #1524 

#### How To Test

1. Create a publisher account with the above criteria satisfied
2. Request `/api/v1/stats/publishers/channel_and_kyc_uphold_and_email_verified_signups_per_day` and ensure the publisher is logged there

#### Pull Request Checklist

### DIFF
--- a/app/controllers/api/v1/stats/publishers_controller.rb
+++ b/app/controllers/api/v1/stats/publishers_controller.rb
@@ -67,6 +67,29 @@ class Api::V1::Stats::PublishersController < Api::V1::StatsController
     render(json: fill_in_blank_dates(result).to_json, status: 200)
   end
 
+  def channel_and_kyc_uphold_and_email_verified_signups_per_day
+    sql =
+    """
+      select p.created_at::date, count(*)
+      from (
+        select distinct publishers.*
+        from publishers
+        inner join channels
+        on channels.publisher_id = publishers.id and channels.verified = true
+        inner join uphold_connections
+        on uphold_connections.publisher_id = publishers.id
+        where role = 'publisher'
+        and uphold_connections.uphold_verified = true
+        and uphold_connections.is_member = true
+        and email is not null
+      ) as p
+      group by p.created_at::date
+      order by p.created_at::date
+    """
+    result = ActiveRecord::Base.connection.execute(sql).values
+    render(json: fill_in_blank_dates(result).to_json, status: 200)
+  end
+
   def totals
     if params[:up_to_date].present?
       up_to_date = Date.parse(params[:up_to_date])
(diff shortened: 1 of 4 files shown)


dlipeles:
> Actually would you group by the `uphold_connections` `created_by` date? It's a more accurate reflection of when the users completed KYC.

@yachtcaptain23 I think this is a really good observation, and I tried rewriting the sql query to base itself on `uphold_connections` 

```ruby
  def channel_and_kyc_uphold_and_email_verified_signups_per_day
    sql =
    """
      select p.created_at::date, count(*)
      from (
        select distinct uphold_connections.*
        from uphold_connections
        inner join channels
        on channels.publisher_id = uphold_connections.publisher_id and channels.verified = true
        inner join publishers
        on publishers.id = uphold_connections.publisher_id and publishers.role = 'publisher' and publishers.email is not null
        where uphold_connections.uphold_verified = true and uphold_connections.is_member = true
      ) as p
      group by p.created_at::date
      order by p.created_at::date
    """
    result = ActiveRecord::Base.connection.execute(sql).values
    render(json: fill_in_blank_dates(result).to_json, status: 200)
  end
```

However what I found is the results were actually similar, it appears uphold connections are `created` automatically for every publisher.

See: 
https://github.com/brave-intl/publishers/blob/7c4b4ce62fcd795728fb8e1c9f294ca6410c1607/app/controllers/publishers_controller.rb#L279-L301

It turns out the `created_at` attribute for `uphold_connections` doesn't tell us when `is_member` became `true`. 


But it got me thinking, we could try grouping by `updated_at`, which might reflect when `is_member` became `true`, but the problem is our numbers would shift with any update to the `uphold_connection`. 

---

After all that, I'm thinking it's best to continue grouping as we have been, because at least the numbers will be right, and should not shift too much. 

In the future we might create some new mechanism to track when a user KYC's. 

Thoughts, @yachtcaptain23 @corymcdonald ?

corymcdonald:
Uphold Diagram (1)

I've been working on an Uphold Diagram that can kind of describe the intricacies of the workflow. I wanna make it more detailed so that we can easily diagnose problems and eventually simplify it.

I think that in particular that we can create a field called `is_member_updated_at` which will be set when the value first becomes true.

PR #1973, opened 2019-06-14
Create "only  user funds" state

corymcdonald:
#### Features

- Create new state called "only user funds"
  - Remove promo registrations when they are being deleted
- Handle new 409 statuses being returned when promo server has user under `no-ugp` owner state.

#### How To Test

Prereqs
- Have the promo server running
- Be able to create many channels

1. Create a channel that will be able to generate a promo code
2. Insert into our publisher into the the promo server

```sql
INSERT INTO owner_states (owner_id, state) VALUES ('3a1dd608-7552-40f2-ad49-92439bb91fa2', 'no-ugp');
```

3. Try to add another channel
4. No promo registration will be created for it 🎉
5. Move publisher over to the `only user funds` state.
6. It deletes all the `PromoRegi`strations

### DIFF
--- a/app/controllers/admin/publishers/publisher_status_updates_controller.rb
+++ b/app/controllers/admin/publishers/publisher_status_updates_controller.rb
@@ -20,6 +20,12 @@ class Admin::Publishers::PublisherStatusUpdatesController < Admin::PublishersCon
       end
     end
 
+    # If the user is being transitioned to only user funds remove all their promo registrations
+    if @publisher.only_user_funds?
+      @publisher.promo_registrations.destroy_all
+      @publisher.channels.find_each { |c| c.promo_registration&.destroy }
+    end
+
     flash[:notice] = "Updated publisher's status to #{@publisher.inferred_status}"
     redirect_to admin_publisher_path(id: @publisher.id)
   end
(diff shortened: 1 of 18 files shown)


app/controllers/admin/publishers/publisher_status_updates_controller.rb
```diff
@@ -20,6 +20,12 @@ def create
       end
     end
 
+    # If the user is being transitioned to only user funds remove all their promo registrations
+    if @publisher.only_user_funds?
+      @publisher.promo_registrations.destroy_all
```
corymcdonald: Hey @aekeus - curious how this works on the promo server side of things, if we mark a user as `no-ugp` on the promo side, and then try to update the `/api/1/promo/publishers/:ref_code` to be `"removed"` will this result in a 409?

aekeus: Yes likely, but I will need to test.

corymcdonald: No worries, we won't actually delete these based on the conversation in slack. Instead we'll just hide them in the view that way the reference to the promo server isn't destroyed

PR #1978, opened 2019-06-19
Updated Terms of Service

corymcdonald:
## Updated Terms of Service
image

#### Features

Creates a rake task to email out every single user to inform them about updated terms of service

#### How To Test

`rake email:tos_update`

### DIFF
--- a/lib/tasks/email_tos_update.rake
+++ b/lib/tasks/email_tos_update.rake
@@ -1,0 +1,18 @@
+namespace :email do
+  task :tos_update => :environment do
+    puts "Emailing #{Publisher.count} users"
+
+    Publisher.find_each.with_index do |user, index|
+      begin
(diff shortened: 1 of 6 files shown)


lib/tasks/email_tos_update.rake
```diff
@@ -0,0 +1,18 @@
+namespace :email do
+  task :tos_update => :environment do
+    puts "Emailing #{Publisher.count} users"
+
+    Publisher.find_each.with_index do |user, index|
```
yachtcaptain23: I believe we're only emailing active publishers

corymcdonald: Mandar mentioned that "- we send it to everyone - including suspended and in setup (just no deletes or erasures)"

PR #1992, opened 2019-06-24
Admin charts

dlipeles:
## Pull Request Name

#### Features

#### How To Test

1. Step one
2. Step two
3. Step three

#### Pull Request Checklist


dlipeles:
@yachtcaptain23 Al I fixed the import/typescript errors, I was able to get the component to show up in admin, but I think the routes just need to be updated for admin, let me know if you want to push onto this pr, or if you would like me to jump back on and finish it.

app/controllers/admin/promo_registrations_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Admin::PromoRegistrationsController < AdminController
+    def for_referral_code
+      publisher = Publisher.find(params[:publisher_id])
+      promo_registration = publisher.promo_registrations.find_by(referral_code: params[:referral_code])
+      render :unauthorized and return if promo_registration.nil?
+      render json: promo_registration.stats_by_date.to_json
+    end
+end
+
```
corymcdonald: nit: need to remove trailing new lines 
```suggestion

```

app/controllers/admin/publishers/referrals_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Admin::Publishers::ReferralsController < Admin::PublishersController
```
corymcdonald: It's preferable to inherit from the administrator ApplicationController, rather than another models controller. When we inherit from another controller then we pollute our controller class somewhat. In addition all the before actions will be called on this making it somewhat more difficult to effectively determine what's going on and potentially leading to unintended side effects.

app/controllers/admin/promo_registrations_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Admin::PromoRegistrationsController < AdminController
+    def for_referral_code
```
corymcdonald: Curious, can we simply call the existing for_referral_code method that we've already defined instead of creating a new controller? 

I think that would prevent us from having to provide scope to all these different react components correct?

app/views/admin/publishers/referrals/index.html.slim
```diff
@@ -0,0 +1,11 @@
+
+- if @publisher.promo_registrations.present?
+  = hidden_field_tag 'all-channels-referral-stats-starting-month', @publisher.promo_registrations.has_stats.order(created_at: :asc).first.created_at.strftime("%Y-%m-%d")
```
corymcdonald: Not quite sure I see where this is being used

app/controllers/admin/promo_registrations_controller.rb
yachtcaptain23: I don't think this works since `for_referral_code` for publishers apply to that specific user, while `admin` scope works for all publishers.

corymcdonald: Seems straight forward enough to fix. The publisher's id is passed into the `for_referral_code` method already
The separation of the routes / additional code seems like it's harder to maintain than simply adding access parameters to the current solution

```ruby
if current_publisher.admin?
else 
  publisher = current_publisher
end
```

app/controllers/admin/publishers/referrals_controller.rb
yachtcaptain23: I don't agree with this, as wouldn't sub routes typically have the same before_action conditions as their parent routes? I'm trying to keep in mind DRY conventions.

That being said, I did fix an error on the next line.

PR #1998, opened 2019-06-25
Create owner_states client for Promo

corymcdonald:
#### Features

* Creates a basic service client for interacting with the promo server
* Adds `OwnerState` as a first endpoint to introduce

#### How To Test

1. Run `bundle exec rails console`
2. Here's some test code

```ruby
client = Promo::Client.new

# empty array
client.owner_state.find(id: '1234')

# returns true
client.owner_state.create(id: '1234', state: Promo::Models::OwnerState::State::SUSPEND)

# returns ["suspend"]
client.owner_state.find(id: '1234')

# returns true
client.owner_state.create(id: '1234', state: Promo::Models::OwnerState::State::NO_UGP)

# returns ["no_ugp", "suspend"]
client.owner_state.find(id: '1234')

# returns true
client.owner_state.destroy(id: '1234', state: Promo::Models::OwnerState::State::SUSPEND)

# returns true (it's not on the list)
client.owner_state.destroy(id: '1234', state: Promo::Models::OwnerState::State::SUSPEND)

```

### DIFF
--- a/app/services/promo/models/owner_state.rb
+++ b/app/services/promo/models/owner_state.rb
@@ -1,0 +1,59 @@
+require 'addressable/template'
+require 'json'
+
+module Promo
+  module Models
+    class OwnerState < Client
+      # For more information about how these URI templates are structured read the explaination in the RFC
+      # https://www.rfc-editor.org/rfc/rfc6570.txt
+      PATH = Addressable::Template.new("/api/2/promo/owners/{id}/states{/state}")
+
+      class State
+        SUSPEND = "suspend".freeze
+        NO_UGP = "no_ugp".freeze
+      end
+
+      def initialize(connection, params = {})
+        super(connection, params)
+      end
+
+      # Finds the current state of the owner on the promo server
+      #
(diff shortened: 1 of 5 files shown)


app/services/promo/models/owner_state.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+      # Removes the state for the specified owner
+      #
+      # @param [String] id The publisher id
+      # @state [Promo::Models::OwnerState::State] state The state to remove from the owner.
+      #
+      # @return [true] if destroy was a success
+      def destroy(id:, state:)
+        response = delete(PATH.expand(id: id, state: state))
```
yachtcaptain23: @aekeus @corymcdonald does this remove the last active state?

As an aside, I've been under the impression that it's better to add a state that says "active" for historical purposes.

aekeus: For the referral server, the absence of all other states is defined as 'active'.

PR #2002, opened 2019-06-26
Improve User Funds Only front end

corymcdonald:
## Improve User Funds Only front end
image

Screen Shot 2019-06-26 at 11 32 27 AM
Screen Shot 2019-06-26 at 11 32 10 AM

#### Features

- Interacts with the Promo server to put publishers in the correct state once 
- Only shows contributions for users who are in the "user_funds_only" state
- Show warning in the promo registrations 
- Creates partials for some of the panels on `home.html.slim` page (It was getting pretty large)

#### How To Test

1. Create a new user, activate the promo registrations
2. Add a few channels
3. Move the user to "User funds only"
4. See that the dashboard has changed


app/helpers/publishers_helper.rb
```diff
@@ -41,6 +41,9 @@ def publisher_overall_bat_balance(publisher)
       publisher = publisher.become_subclass
       amount = publisher.wallet&.overall_balance&.amount_bat
       amount = publisher.balance if publisher.partner?
+      # If the user is in the "user funds only" then only show contributions
```
yachtcaptain23: IMO this should be an if...else block

app/helpers/publishers_helper.rb
```diff
@@ -53,6 +56,8 @@ def publisher_converted_overall_balance(publisher)
     publisher = publisher&.become_subclass
     balance = publisher.wallet&.overall_balance&.amount_default_currency
```
yachtcaptain23: Another if...else block here too

corymcdonald: https://github.com/brave-intl/publishers/pull/2002/commits/75a524baf9b249a02aeabc542d24e76eb34eb794

app/models/publisher_status_update.rb
```diff
@@ -30,6 +32,32 @@ class PublisherStatusUpdate < ApplicationRecord
 
   validates :publisher_id, presence: true
 
+  # After a user creates a new status then we should check to see the previous staus and call backing server
+  after_create :update_services, if: :should_update?
+
+  # Calls the promo server to update the owner state for the publisher based on the status
+  #
+  # @return [true] if the services successfully update
+  def update_services
```
yachtcaptain23: API calls like this should be done in a job. It'll automatically handle any kind of failures that occur during downtime.

corymcdonald: I tend to prefer system correctness over reliability in these instances, however I'll move it to a job

config/locales/en.yml
```diff
@@ -684,6 +684,7 @@ en:
       confirmed: "confirmed"
       active_referral: "active"
       add_channel: "Add a web site or youtube channel below to get the promo link!"
+      violation: We regret to inform you that we've suspended your right to participate in Brave Rewards Referral Program due to violation of our <a href="https://basicattentiontoken.org/publisher-terms-of-service/">Terms of Service</a>. However, you can still earn and redeem rewards from from participating in our other programs.
```
dlipeles: Looks great Cory! Any chance we could run this copy by catherine corre ?

corymcdonald: Sent her a message 👍

app/models/publisher_status_update.rb
corymcdonald: Updated
https://github.com/brave-intl/publishers/pull/2002/commits/027da54e6b53f213d1e15db15d9d8381487bf6b3

app/jobs/promo/register_publisher_for_promo_job.rb
```diff
@@ -0,0 +1,19 @@
...
+    include PromosHelper
+    queue_as :default
+
+    def perform(publisher:)
+      return unless publisher.promo_enabled_2018q1
+      Rails.logger.info("Registering publisher #{publisher.id} for promo async.")
+
+      if Promo::PublisherChannelsRegistrar.new(publisher: publisher).perform
+        promo_enabled_channels = publisher.channels.joins(:promo_registration)
+        PromoMailer.promo_activated_2018q1_verified(publisher, promo_enabled_channels).deliver
```
yachtcaptain23: I believe this is `deliver_now` rather than `deliver`

config/locales/en.yml
corymcdonald: Text changed based on David & Catherine's feedback 
https://github.com/brave-intl/publishers/pull/2002/commits/67ab48f71a096ec69d1d6a1513fe6b33acc752a4

PR #2023, opened 2019-07-03
Attempt to fix https://github.com/brave-intl/publishers/issues/733. W…

yachtcaptain23:
…e have code expiration logic, not sure why we need a CustomFailureApp

#### Features

Follow the convention for `timeoutable` per instructions found on https://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Timeoutable

#### How To Test

1. Change the value for `timeout_in` to `1.minute` to see the effect.


corymcdonald:
When I register a new account I get `translation missing: en.publishers.show.verification_required`

corymcdonald:
One thought I had would be to test this out on staging where we use HTTP AUTH for logging in

PR #2034, opened 2019-07-09
Fix obscure channel transfer logic by not making circular calls

yachtcaptain23:
#### Features
Fixes tech debt found in channel transfer and when deleting a channel.


yachtcaptain23:
Accidentally closed.

PR #2036, opened 2019-07-09
Gemini v2 api for a single channel

yachtcaptain23:
#### Features

Closes https://github.com/brave-intl/publishers/issues/2035

### DIFF
--- a/app/models/json_builders/channels_json_builder_v2.rb
+++ b/app/models/json_builders/channels_json_builder_v2.rb
@@ -22,6 +22,7 @@ class JsonBuilders::ChannelsJsonBuilderV2
     @excluded_channel_ids = Publishers::ExcludedChannels.brave_publisher_id_list
     @excluded_verified_channel_ids = []
     @channels = []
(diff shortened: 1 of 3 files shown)


app/models/json_builders/channels_json_builder_v2.rb
```diff
@@ -53,13 +54,19 @@ def include_verified_channel(verified_channel)
     return if @excluded_channel_ids.include?(verified_channel.details.channel_identifier)
     # Skip if channel publisher has not yet KYC'd
     return unless verified_channel.publisher&.uphold_connection&.is_member
+
+    wallet_address_id = verified_channel.publisher&.uphold_connection&.address
+    return if wallet_address_id.nil?
+    return if existing_wallet_address?(wallet_address_id)
```
corymcdonald: Don't we want to push the channel even if the wallet address is existing?

yachtcaptain23: It's a hash to check to see if the address is a duplicate.

Issue #1952, opened 2019-06-10
Idempotent uphold_verified


corymcdonald:
We are getting a lot of alerts regarding `missing_access_parameters` this is because the `uphold_verified` endpoint is not idempotent.

When a user requests `uphold_verified` then we try and receive the code and do not check if any code has been received or accepted. We should try and rework this logic so that we do not encounter this anymore

corymcdonald:
Close https://github.com/brave-intl/publishers/pull/2094

PR #2042, opened 2019-07-11
Fix possible double calls to UploadUpholdAccessParametersJob. Related…

yachtcaptain23:
… to #1952

#### Features

Since UploadUpholdAccessParametersJob caller could be called multiple times, UploadUpholdAccessParametersJob might run multiple times. This adds a check to make sure we're not making noise about the uphold_access_parameters being gone when the connection is already verified.

### DIFF
--- a/app/jobs/upload_uphold_access_parameters_job.rb
+++ b/app/jobs/upload_uphold_access_parameters_job.rb
@@ -3,7 +3,7 @@ class UploadUpholdAccessParametersJob < ApplicationJob
 
   def perform(publisher_id:)
     publisher = Publisher.find(publisher_id)
-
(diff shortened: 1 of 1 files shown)


app/jobs/upload_uphold_access_parameters_job.rb
```diff
@@ -3,7 +3,7 @@ class UploadUpholdAccessParametersJob < ApplicationJob
 
   def perform(publisher_id:)
     publisher = Publisher.find(publisher_id)
-
+    return if publisher.uphold_connection.uphold_verified
```
corymcdonald: My only concern is that people who want to go and reconnect their uphold after they've revoked access

PR #2050, opened 2019-07-12
Upgrade puma gem from 3.12.1 to 4.0.1 which now has better support for local SSL certs.

yachtcaptain23:


Gemfile
```diff
@@ -95,7 +95,7 @@ gem "premailer-rails", "~> 1.9.4", require: false
 gem 'public_suffix', '~> 3.0.2'
 
 # Puma as app server
-gem "puma", "3.12.1"
+gem "puma"
```
corymcdonald: I think we should lock down Puma, if a major version gets released then we run into some problems while running `bundle update`

PR #2062, opened 2019-07-22
Move Uphold Card creation from Eyeshade

corymcdonald:
#### Features

Refactor the Uphold Card creation from Eyeshade to Publishers

#### What's left?

- Tests
- Initial review on refactoring


app/jobs/create_uphold_cards_job.rb
```diff
@@ -2,24 +2,23 @@
...
 
-  def perform(publisher_id:)
-    publisher = Publisher.find(publisher_id)
-
-    unless publisher.uphold_connection.can_create_uphold_cards?
-      Rails.logger.info("Could not create uphold card for publisher #{publisher.id}.")
-      SlackMessenger.new(message: "Could not create uphold card for publisher #{publisher.id}.", channel: SlackMessenger::ALERTS).perform
+  def perform(uphold_connection:)
+    unless uphold_connection.can_create_uphold_cards?
+      # Reasons might include that they are not in a district which allows for crypto, like tennesse
```
evq: I believe this particular error would not result in `can_create_uphold_cards` being false - the user will be able to create cards, just not ones denominated in crypto IIRC

app/models/uphold_connection.rb
```diff
@@ -51,50 +47,62 @@ def prepare_uphold_state_token
     save!
```
evq: while we're in here - I have a quick question about the logic to create the `state`. can you confirm that we will generate a new state for every time we attempt to link (whether we succeed in getting a code back or not)?

it looks to me like there may be possible re-use in the case where the user doesn't complete the linking flow

app/services/uphold/models/card.rb
```diff
@@ -0,0 +1,89 @@
...
+        else
+          cards = JSON.parse(response.body)
+        end
+
+        cards = cards.map { |c| Uphold::Models::Card.new(c) } if cards.present?
+
+        cards
+      end
+
+      # Finds a card given the uphold_connection and a currency
```
evq: these docstrings are out of date

evq:
are we planning to check an the eyeshade -> publishers migration script here or elsewhere?

app/models/uphold_connection.rb
corymcdonald: Yes, it certainly seems like we could get into an issue where if this was interrupted then we would be reusing the state token. The [state token is generated only once before they click on it](https://github.com/brave-intl/publishers/blob/store-uphold-in-publishers/app/models/uphold_connection.rb#L44).
We've actually seen the codes not work because of the user not completing which is why there is a `CleanStaleUpholdDataJob` which clears out the codes.

During the first refactor to migrate from storing all these properties directly on the Publisher's model I experimented with regenerating every time but it didn't work due to needing to store the code for more than one request., Thinking about this again I think when the user clicks on "Connect to Uphold" they should be directed to a Publisher's controller which then generated a code, then redirects them to uphold which then redirects back to `uphold_verified`

corymcdonald:
>are we planning to check an eyeshade -> publishers migration script here or elsewhere?

Not to sure how this should go - we'll need to migrate the `uphold_access_parameters` from Eyeshade to Publishers. It might be easy enough to code up a ruby script that would connect to Eyeshade's DB and pull the data and insert the relevant fields into Publishers.

@evq @yachtcaptain23 Do you have any opposition to direct DB connection to Eyeshade to read access parameter values?

app/models/uphold_connection.rb
corymcdonald: Here's the change that generates a new state token every time the user clicks 'Connect'
https://github.com/brave-intl/publishers/pull/2062/commits/58dda10651cfe2b9f3c1316b843a28f6cf86b633

app/views/admin/publishers/show.html.slim
```diff
@@ -53,9 +53,9 @@ hr
         .db-info-row
           .db-field = "Overall BAT balance:"
           .db-value = "#{publisher_overall_bat_balance(@publisher)}"
-        - if @publisher.default_currency.present?
+        - if @publisher.uphold_connection.default_currency.present?
           .db-info-row
-            .db-field = "Overall #{@publisher.default_currency} balance:"
+            .db-field = "Overall #{@publisher.uphold_connection. default_currency} balance:"
```
yachtcaptain23: extra space in `uphold_connection. `

corymcdonald: I ended up making the Uphold Connection it's own section on the show page of the administration dashboard, that way it's easier to view 

image

test/fixtures/uphold_connections.yml
```diff
@@ -1,8 +1,22 @@
+<% salt = SecureRandom.random_bytes(12) %>
+
 unconnected:
   created_at: 2019-01-01
   publisher: completed
-  is_member: true
-  address: "abcdabcd-0000-0000-0000-abcd00000001"
+  is_member: false
+
+suspended_connection:
```
yachtcaptain23: You can simplify the file if you use default args like `user_authentication_tokens.yml`

app/services/uphold/models/card.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2062/commits/0b073f817a0e365a584bed4247f1a52848921e68

corymcdonald:
@evq @yachtcaptain23 
Created a migration for existing access parameters in `lib/tasks/database_updates/migrate_uphold_access_parameters.rake` 

More testing is required to finesse the output but essentially 

1. Connect to Mongo
2. Go through each row
3. Update the publisher's uphold account
4. Pull latest information from Uphold
5. Find / update the address for the user's card based on the uphold information

left to do
- [ ]  Eventually read the mongo connection string from environment
more testing 
- [ ] Test for 429's? What happens if Uphold does throttle us?
- [ ] Validate the information from Eyeshade to Publishers to ensure things are synchronized correctly

test/fixtures/uphold_connections.yml
corymcdonald: Going to push this off till later because it doesn't hurt anything to keep this mock data like this

lib/tasks/database_updates/migrate_uphold_access_parameters.rake
```diff
@@ -0,0 +1,58 @@
...
+
+    Mongo::Logger.logger.level = Logger::FATAL
+
+    client = Mongo::Client.new(mongo_uri)
+    owners = client[:owners]
+
+    iterations = owners.count
+    puts "Anticipating queueing #{iterations} entries"
+
+    owners.find.each do |row|
```
yachtcaptain23: I believe this is `find_each`?

corymcdonald: Since this is a `Mongo::Collection` find_each doesn't work 😞 

Here's an example of how the Mongo connector works
https://github.com/mongolab/mongodb-driver-examples/blob/master/ruby/ruby-driver/ruby_simple_example.rb

app/models/uphold_connection.rb
```diff
@@ -108,17 +123,41 @@ def uphold_processing?
...
   def can_create_uphold_cards?
     uphold_verified? &&
-      wallet.present? &&
-      wallet.authorized? &&
-      wallet.scope &&
-      wallet.scope.include?("cards:write") &&
+      uphold_access_parameters.present? &&
+      scope.include?("cards:write") &&
+      status != "pending" &&
+      status != "blocked" &&
```
yachtcaptain23: I believe we can use `UpholdConnection::BLOCKED` or perhaps expand it to contain all states in https://uphold.com/en/developer/api/documentation/#user-object

app/controllers/publishers/uphold_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+    # If user does not have Uphold's `cards:write` scope, we redirect to Uphold to get authorization
+    # Card creation is done in #home
+    def confirm_default_currency
+      # TODO Consider refactoring this
+      uphold_connection = current_publisher.uphold_connection
+
+      uphold_connection.update(confirm_default_currency_params.merge(default_currency_confirmed_at: Time.now))
+
+      if uphold_connection.can_create_uphold_cards?
+        uphold_connection.create_uphold_card_for_default_currency
```
yachtcaptain23: Would this run for users who disconnected and re-connected?

app/services/initializable.rb
```diff
@@ -0,0 +1,8 @@
+module Initializable
+  def initialize(params = {})
+    params.each do |key, value|
+      setter = "#{key}="
```
yachtcaptain23: Would you elaborate on this function? I don't understand what's going on here. Are some of the params just methods of the particular object?

corymcdonald: So this is a ruby meta-programming method, basically anytime that you call initialize with a series of params like so
`Object.new(id: "1234", some_property: "foobar")`

Then for each of those params we'll iterate those and call the associated setter method. In this example this method would call the methods `id=`, and `some_property=`.

Now, this is where it kind of gets strange because you'll notice in `user.rb` and `card.rb` I do not define any methods like this. 

That's where `attr_accessor` comes in. This method allows for a shorthand for generating getters/setters. So if I have a class like 
```ruby
class Card
  def id
    @name
  end

  def id=(str)
    @id = str
  end
end
```

This can be shortened to
```ruby
class Card
  attr_accessor :id
end
```

So in this method we go through each of the passed hash params, define a string called `setter` for the method we want to call with our value being the hash value.

So in our example above 

```ruby
# passed in through the params
value = "1234" 
# the property we're setting
setter = "id=" 
# send the method the value
send(setter, value)
# calls the method id=("1234")
```

app/controllers/publishers/uphold_controller.rb
corymcdonald: I believe so, this might cause a 401 error if they are confirming a default currency. However from the UI we don't allow the user to change their default currency when they're in the "Reauthorize state.

app/models/uphold_connection.rb
corymcdonald: Good idea!
https://github.com/brave-intl/publishers/pull/2062/commits/58056406d5c24383360219451d242f01b329dcfb

app/services/initializable.rb
yachtcaptain23: Cool that makes sense. Would be good to add a small comment with your example.

app/jobs/create_uphold_cards_job.rb
```diff
@@ -2,24 +2,23 @@
 class CreateUpholdCardsJob < ApplicationJob
   queue_as :default
 
-  def perform(publisher_id:)
-    publisher = Publisher.find(publisher_id)
-
-    unless publisher.uphold_connection.can_create_uphold_cards?
-      Rails.logger.info("Could not create uphold card for publisher #{publisher.id}.")
-      SlackMessenger.new(message: "Could not create uphold card for publisher #{publisher.id}.", channel: SlackMessenger::ALERTS).perform
+  def perform(uphold_connection:)
```
yachtcaptain23: You need to lookup the `uphold_connection` by id since Sidekiq jobs only allow primitives (and dynos don't know each other memory space)

corymcdonald:
# Test Plan
Before we begin we need to run a payout report. Take screenshots of relevant balances, etc

## Post Deployment
Run the following migration command
`rake database_updates:migrate_uphold_access_parameters`

Monitor the sidekiq queue by going to `/magic`, check to see if there are any failures or retries


# Testing
## Homepage

### Prerequisites 
We should create 4 uphold accounts
1. A pending account, just one that hasn’t completed any of the sign up 
2. Complete sign up but not KYC
3. KYC the account
4. A blocked account


### New Publisher
1. Sign up with an email address that hasn’t been used before
2. Click the email link to login
3. The home page loads correctly

### New publisher can connect uphold

1. As the new publisher add a few channels
2. Connect Uphold
3. Select a default currency
4. A card should be created for the publisher under the account

### When a user disconnects and reconnect Uphold the same card is used

1. Go to the user’s uphold cards list
2. Disconnect Uphold
3. Reconnect Uphold
4. Ensure that if the same currency is selected another card is not created

### When a user changes their default currency a card is created
1. On the publishers dashboard click “Change default currency”
2. Select a unique default currency like LTC / DASH / UPBTC, etc. 
3. Go to the Uphold cards page and ensure a “Brave Rewards” card was created for the specified currency

### When a publisher changes their uphold password

1. Log into uphold
2. Change your password
3. Log into Publishers
4. Should be presented with a “Re-authorize” button

### A publisher can download their statements

1. Log into the publishers dashboard
2. Click on “Statements”
3. Click on “All statements”
4. Download is successful

## Administrator dashboard
### You can view publishers
1. Click on a publisher who has connected their uphold account
2. You can refresh to see if any properties have changed
3. Go to another publisher account which isn’t uphold connected, the page still loads
4. Go to a publisher who hasn’t signed in. The page still loads

### Payout report is equivalent
1. Compare the payout report that was previously generated
2. Ensure the contents are equal (minus minor amount changes)

### An overview of countries for Publishers is shown

1. Click on the publishers statistics page 
2. A table which populates the list of countries in which we have publishers is populated

corymcdonald:
Bugs 🐛found while testing on staging

1. Pending account doesn't get a valid JSON response back from `home/wallet`
2. Upon page refresh the entire page 500's

app/jobs/create_uphold_cards_job.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2062/commits/4929743cb20a2bffbaed09e9628689bf04e53d1d

corymcdonald:
The issues that were found 
[ActionView::Template::Error: undefined method `-' for nil:NilClass](https://github.com/brave-intl/publishers/issues/2092) and [NoMethodError: undefined method `default_currency' for nil:NilClass](https://github.com/brave-intl/publishers/issues/2093) was resolved by this commit https://github.com/brave-intl/publishers/pull/2062/commits/04038a0ccb2a6667bbac26b6a72c1ec663e85374 and https://github.com/brave-intl/publishers/pull/2062/commits/ebaf22d9f67ecbd55a3e2d34db6b8c2df1f7d31b

app/services/initializable.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2062/commits/0126df6ebaec13b1e1bfc8c6f9ae98517a045e65

corymcdonald:
Added a migration message 
image

PR #2089, opened 2019-08-01
Admin title polish

rossmoody:
General polish updates again. 

- consolidated page titles to h4 tags
- consolidated search and filter/sorting action items to one row
- updated default btn to feel more like a button

Screen Shot 2019-08-01 at 12 23 29 PM
Screen Shot 2019-08-01 at 12 23 33 PM
Screen Shot 2019-08-01 at 12 23 47 PM

### DIFF
--- a/app/assets/stylesheets/admin/style.scss
+++ b/app/assets/stylesheets/admin/style.scss
@@ -1,5 +1,31 @@
 @import "colors";
 
+// Heading and button row
+.panel-heading {
+  display: flex;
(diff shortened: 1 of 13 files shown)


corymcdonald:
Screen Shot 2019-08-01 at 4 31 08 PM

How do you feel about this search bar? It seems like maybe we would also want to be consistent with them

rossmoody:
so slick. love it

PR #2094, opened 2019-08-02
Create card per channel

corymcdonald:
#### Features

Creates a card per channel


app/helpers/channels_helper.rb
```diff
@@ -53,7 +53,7 @@ def channel_next_step_path(channel)
   end
 
   def channel_verification_status(channel)
-    if channel.verified?
+    if channel.verified? && channel.publisher.uphold_connection.is_member?
```
yachtcaptain23: I don't think we should add this, because having or not having verification with Uphold is independent of channel getting verified. Plus we should inform support team of this change

app/helpers/promos_helper.rb
```diff
@@ -51,6 +51,25 @@ def publisher_current_referral_totals(publisher)
     }
   end
 
+  def channel_referral_totals(channel)
```
yachtcaptain23: I believe this is a duplicate of `aggregate_stats` on the `promo_registrations` model.

app/helpers/channels_helper.rb
yachtcaptain23: Or I guess we should indicate elsewhere on a second field whether or not the channel connected to uphold.

db/migrate/20190730162819_create_uphold_connection_for_channel.rb
```diff
@@ -0,0 +1,13 @@
+class CreateUpholdConnectionForChannel < ActiveRecord::Migration[5.2]
+  def change
+    create_table :uphold_connection_for_channels, id: :uuid, default: -> { "uuid_generate_v4()"}, force: :cascade do |t|
+      t.belongs_to :uphold_connection, type: :uuid, index: true, null: false
```
yachtcaptain23: Let's add uniqueness constraints just in-case

corymcdonald: Should not be unique on this, because users will have multiple channels

app/helpers/channels_helper.rb
corymcdonald: Yeah, this was just for a UI representation but I think we'll do something different on the homepage. 
https://github.com/brave-intl/publishers/pull/2094/commits/1addcb99452d89472851a87dd6fa11bef67617fd

app/helpers/promos_helper.rb
corymcdonald: Fixed! https://github.com/brave-intl/publishers/pull/2094/commits/c2ca42aeb6d72513d0167f2d56d7651973e8e470

app/jobs/create_uphold_channel_card_job.rb
```diff
@@ -0,0 +1,71 @@
...
+    channel = Channel.find(channel_id)
+    return unless uphold_connection&.is_member? && channel&.verified?
+
+    unless uphold_connection.can_create_uphold_cards?
+      Rails.logger.info("Could not create uphold card for channel #{uphold_connection.publisher_id}. Uphold Verified: #{uphold_connection.uphold_verified}")
+      SlackMessenger.new(message: "Could not create uphold card for the channel #{uphold_connection.publisher_id}.", channel: SlackMessenger::ALERTS).perform
+      return
+    end
+
+    upfc = UpholdConnectionForChannel.where(
```
yachtcaptain23: It should only find one, so I think there should be some kind of uniqueness constraint somewhere, right?

app/models/channel.rb
```diff
@@ -376,6 +381,10 @@ def register_channel_for_promo
     Promo::RegisterChannelForPromoJob.new.perform(channel: self)
   end
 
+  def create_channel_card
+    CreateUpholdChannelCardJob.perform_now(uphold_connection_id: publisher.uphold_connection&.id, channel_id: id)
```
yachtcaptain23: If Uphold goes down, anyone adding a card would experience a major slowdown too on any page renders. Preferable that we make this ` perform_later`.

db/migrate/20190730162819_create_uphold_connection_for_channel.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.belongs_to :channel, type: :uuid, index: true, null: false
+
+      t.string :currency, index: true
+      t.string :channel_identifier, index: true
```
yachtcaptain23: It doesn't seem like we really use this except for lookups in the future, right? Might be worth removing or mentioning that it's de-normalized. Might be worth having a conversation related https://stackoverflow.com/questions/5437744/naming-conventions-for-non-normalized-fields

corymcdonald: I've not really given much thought to the naming that much for these denormalized fields. 

Would you like to rename this to `derived_channel_identifier`?

app/models/channel.rb
corymcdonald: I was debating about this one, but i'm good with moving it to `perform_later`

https://github.com/brave-intl/publishers/pull/2094/commits/e95259299e1a0f477d13b779ab7d0072878f5bd0

app/jobs/create_uphold_channel_card_job.rb
corymcdonald: Just added a validation
https://github.com/brave-intl/publishers/pull/2094/commits/e95259299e1a0f477d13b779ab7d0072878f5bd0

app/jobs/create_uphold_cards_job.rb
```diff
@@ -9,8 +9,17 @@ def perform(uphold_connection_id:)
       return
     end
 
+    card = nil
+
     # Search for an existing card
-    card = uphold_connection.uphold_client.card.where(uphold_connection: uphold_connection)&.first
+    cards = uphold_connection.uphold_client.card.where(uphold_connection: uphold_connection)
```
yachtcaptain23: Shouldn't there be a currency match too?

corymcdonald: In this case it uses the currency that is on the `uphold_connection` object, it's not very clear form the interface. Should I go ahead and pass the currency in?

PR #2104, opened 2019-08-05
Devise shows t.devise.failure.timeout after a user signs in. Removing…

### LINKED ISSUE #2082: Fix erroneous flash banners for users who just signed in
After we had resolved the expired login bug, we have erroneous flash messages at the top. It's a bit annoying to see these messages that are wrong, so we should fix or just remove them entirely.

yachtcaptain23:
… this line removes the message by not forcing a sign out. Closes #2082 

## Pull Request Name

#### Features

#### How To Test

1. Step one
2. Step two
3. Step three

#### Pull Request Checklist

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -266,8 +266,6 @@ class PublishersController < ApplicationController
   private
 
   def authenticate_via_token
-    sign_out(current_publisher) if current_publisher
-
     publisher_id = params[:id]
     token = params[:token]
     confirm_email = params[:confirm_email]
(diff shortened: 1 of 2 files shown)


app/controllers/publishers_controller.rb
```diff
@@ -266,8 +266,6 @@ def get_site_banner_data
   private
 
   def authenticate_via_token
-    sign_out(current_publisher) if current_publisher
```
corymcdonald: Could this raise an exception if we continue down with the method with an already signed in user? 

Perhaps we should just return if they are already authenticated? (Not sure about that one)

yachtcaptain23: I tested by: 
(0) Open mailcatcher in non-private window
(1) Signing in a non-private window for a@a.com
(2) Fetching a login email in a private window for b@b.com
(3) In non-private window, clicking on the link in mailcatcher for b@b.com
(4) See that session has switched to b@b.com

PR #2123, opened 2019-08-08
Revert "Create card per channel"

yachtcaptain23:
Reverts brave-intl/publishers#2094


corymcdonald:
would this revert would fail due to the db change? 

I think it would easier for longterm support to just modify the V2 API back to the original.

yachtcaptain23:
Closing for the time being as it doesn't seem we need this.

PR #2125, opened 2019-08-08
Fix manual payout publisher includer

### LINKED ISSUE #2120: Fix reauthorization_needed on ManualPayoutReportPublisherIncluder
In /app/services/manual_payout_report_publisher_includer.rb should be `.blank?` rather than `.present?`

corymcdonald:
#### Features

This fix the manual payout report. Before we were filtering out the reauthorization needed. This was a problem because the logic was reversed

closes #2120

#### How To Test

1. Kick off a manual payout report with people who should be on it.
2. Refreshing the JSON works
3. The report contains entrise

### DIFF
--- a/app/services/manual_payout_report_publisher_includer.rb
+++ b/app/services/manual_payout_report_publisher_includer.rb
@@ -10,10 +10,6 @@ class ManualPayoutReportPublisherIncluder < BaseService
 
     uphold_connection = @publisher.uphold_connection
 
-    uphold_status = uphold_connection.status
-    uphold_member = uphold_connection.is_member?
-    reauthorization_needed = uphold_connection.uphold_access_parameters.present?
-    uphold_id = uphold_connection.uphold_id
     suspended = @publisher.suspended?
 
     invoices = Invoice.where(partner_id: @publisher.id, status: Invoice::IN_PROGRESS)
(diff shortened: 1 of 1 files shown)


app/services/manual_payout_report_publisher_includer.rb
```diff
@@ -26,11 +22,11 @@ def perform
                               publisher_id: @publisher.id,
                               kind: PotentialPayment::MANUAL,
                               address: uphold_connection.address.to_s,
-                              uphold_status: uphold_status,
-                              reauthorization_needed: reauthorization_needed,
-                              uphold_member: uphold_member,
+                              uphold_status: uphold_connection.status,
+                              reauthorization_needed: uphold_connection.uphold_access_parameters.blank?,
```
yachtcaptain23: are we sure we want to invert the condition? This was the original code: `reauthorization_needed = uphold_connection.uphold_access_parameters.present?`

corymcdonald: Yup, we want to because that's actually the error preventing the manual payout from downloading.

The uphold_access_parameters will be blank when reauthorization happens due to this line
https://github.com/brave-intl/publishers/blob/staging/app/models/uphold_connection.rb#L103

PR #2135, opened 2019-08-09
Fix Publisher Uphold Cards

corymcdonald:
#### Features

Some of the publishers that have existing cards linked are having their cards link automatically to their wallet. This adds some additional logic to try and prevent that and add deterministic lookups when `CreateUpholdCardsJob` is called.

**The strategy is:**
We look up the existing address on the connection If it’s the same currency as the uphold_connection then we don’t modify anything.

If it’s not, then we get a list of the uphold’s users cards, then iterate on each one of them to ensure that
   1. they aren’t for a channel (we haven’t stored them in the Uphold
   2. they don’t have a private address (This one requires an API card per card to /card/:id/addresses

If there are no results that match that then we simply create the card

### DIFF
--- a/app/jobs/create_uphold_cards_job.rb
+++ b/app/jobs/create_uphold_cards_job.rb
@@ -9,16 +9,37 @@ class CreateUpholdCardsJob < ApplicationJob
       return
     end
 
-    card = nil
+    # We create the cards, so let's first check to see if the publisher has an address so we might not have to change anything
+    if uphold_connection.address.present?
+      begin
+        card = uphold_connection.uphold_client.card.find(uphold_connection: uphold_connection, id: uphold_connection.address)
+
+        return if card&.currency.eql?(uphold_connection.default_currency)
+      rescue Faraday::ResourceNotFound
+        # This most likely won't ever happen because it's not possible to delete a card. Better safe than sorry.
+      end
+    end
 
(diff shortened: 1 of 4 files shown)


app/jobs/create_uphold_cards_job.rb
```diff
@@ -9,16 +9,37 @@ def perform(uphold_connection_id:)
...
     # Search for an existing card
     cards = uphold_connection.uphold_client.card.where(uphold_connection: uphold_connection)
     existing_private_cards = UpholdConnectionForChannel.select(:card_id).where(uphold_connection: uphold_connection).to_a
 
+    # This is the default label that we apply to a card. If we find it then it's safe to assume that this was one we've created prior.
+    card = cards.detect { |c| c.label.eql?("Brave Rewards") }
+
+    # User's can change the label's on their cards so if we couldn't find it, we'll have to iterate until we find a card that matches are criteria
+    # 1. Isn't the browser's wallet card
+    # 2. Isn't a channel card
```
yachtcaptain23: Might be better to extract this block to its own function. Also only call `if card.present?`

corymcdonald: Fixed things up - I keep in the `break if card.present?` because keeping it as is was a little bit simpler than rewriting the loop

PR #2176, opened 2019-08-20
Move create_uphold_cards to be after_commit

corymcdonald:
#### Features

It seems like when a user adds a new channel then the job fails on the first time. I googled this and found that this is a fairly common "gotcha" 
https://www.justinweiss.com/articles/a-couple-callback-gotchas-and-a-rails-5-fix/

This fix should allow users to have their channels be created successfully

### DIFF
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -77,7 +77,9 @@ class Channel < ApplicationRecord
   validate :verified_duplicate_channels_must_be_contested, if: -> { verified? }
 
   after_save :register_channel_for_promo, if: :should_register_channel_for_promo
-  after_save :create_channel_card, :notify_slack, if: -> { :saved_change_to_verified? && verified? }
+  after_save :notify_slack, if: -> { :saved_change_to_verified? && verified? }
+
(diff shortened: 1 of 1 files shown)


app/models/channel.rb
```diff
@@ -77,7 +77,9 @@ class Channel < ApplicationRecord
   validate :verified_duplicate_channels_must_be_contested, if: -> { verified? }
 
   after_save :register_channel_for_promo, if: :should_register_channel_for_promo
-  after_save :create_channel_card, :notify_slack, if: -> { :saved_change_to_verified? && verified? }
+  after_save :notify_slack, if: -> { :saved_change_to_verified? && verified? }
+
+  after_commit :create_channel_card, if: -> { :saved_change_to_verified? && verified? }
```
yachtcaptain23: I think this possibly could result in a double card create. Might be worth investigating in a redis lock or a sql lock.

corymcdonald: This method should be able to called multiple times. `UpholdConnectionForChannel` has indexes built in and there is logic to handle duplicates in the `CreateUpholdChannelCardJob`

PR #2200, opened 2019-08-28
Fix error on homepage load

corymcdonald:
#### Features

Because we weren't tracking Uphold IDs on the card connections we ran into an issue where if someone tries and create new cards then there will be an error.

### DIFF
--- a/app/jobs/create_uphold_channel_card_job.rb
+++ b/app/jobs/create_uphold_channel_card_job.rb
@@ -61,7 +61,7 @@ class CreateUpholdChannelCardJob < ApplicationJob
   end
 
   def card_exists?(uphold_connection, card_id)
-    return unless card_id.present?
+    return if card_id.blank?
 
     uphold_connection.uphold_client.card.find(uphold_connection: uphold_connection, id: card_id).present?
   rescue Faraday::ResourceNotFound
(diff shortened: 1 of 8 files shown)


app/jobs/create_uphold_channel_card_job.rb
```diff
@@ -57,12 +60,15 @@ def find_or_create_card(uphold_connection, channel)
     [card_id, upfc]
   end
 
+  def card_exists?(uphold_connection, card_id)
+    return if card_id.blank?
```
yachtcaptain23: `return false` is preferable

app/jobs/create_uphold_channel_card_job.rb
```diff
@@ -57,12 +60,15 @@ def find_or_create_card(uphold_connection, channel)
...
+
+    uphold_connection.uphold_client.card.find(uphold_connection: uphold_connection, id: card_id).present?
+  rescue Faraday::ResourceNotFound
```
yachtcaptain23: I think it should return `false`

app/jobs/create_uphold_channel_card_job.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2200/commits/80a3639f7c23cf850d130406ef3db91bc8eb733d

app/jobs/create_uphold_channel_card_job.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2200/commits/80a3639f7c23cf850d130406ef3db91bc8eb733d

PR #2202, opened 2019-08-29
Feature/thirty day login

yachtcaptain23:


corymcdonald:
I'm having some difficulties seeing this is a toggle. Can Jenn maybe propose something that looks a little different? I imagine people with the vision problems will also have difficulties

Screen Shot 2019-09-05 at 10 57 17 AM

app/views/publishers/settings/index.html.slim
```diff
@@ -1,20 +1,33 @@
+.single-panel--wrapper.single-panel--wrapper--large.single-panel--wrapper--short
+  .single-panel--padded-content--short-padding
+    h5= t ".header"
+    = form_for(current_publisher, url: publisher_path(current_publisher.id), method: :patch) do |f|
+      .row
+        .col-left-aligned--content
```
corymcdonald: I was thinking it make more sense to use the flex model in the Bootstrap based on here - https://getbootstrap.com/docs/4.0/utilities/flex/#wrap

It's pretty minor but it might make it easier in the future

app/views/publishers/settings/index.html.slim
```diff
@@ -1,20 +1,33 @@
+.single-panel--wrapper.single-panel--wrapper--large.single-panel--wrapper--short
+  .single-panel--padded-content--short-padding
+    h5= t ".header"
+    = form_for(current_publisher, url: publisher_path(current_publisher.id), method: :patch) do |f|
```
corymcdonald: It's interesting to me that we have this form linking to the publishers_controller but this view is in the settings_controller. Seems like this might be a separation of concerns problem?

yachtcaptain23: To me it makes most sense since it's an update to the `publishers` table/model. `settings_controller` in retrospect should be `#edit`. Thoughts?

corymcdonald: I'm good with that

corymcdonald:
Most of the time we see these things on login screens, wondering did you talk to Jenn about that?

yachtcaptain23:
@corymcdonald the problem with our flow is that there isn't a convenient place to show. We can drive users to the settings page which users have discovered.

PR #2209, opened 2019-09-04
Better Statements

corymcdonald:
#### Features

Improve the statements for publishers. This is most likely going to be closed and re-open when it's finalized


app/javascript/packs/StatementDetails.jsx
```diff
@@ -0,0 +1,14 @@
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+import StatementDetails from "../views/statements/statements/StatementDetails";
+
+document.addEventListener("DOMContentLoaded", () => {
+  const element = document.querySelector("#statement_details");
+  console.log(element.dataset);
```
yachtcaptain23: Debug statement

app/controllers/publishers/statements_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      else
+        @statement_period = publisher_statement_period(@transactions)
+        statement_file_name = publishers_statement_file_name(@statement_period)
+        statement_string = render_to_string :layout => "statement"
+        send_data statement_string, filename: statement_file_name, type: "application/html"
+      end
+    end
+
+
+    def publisher
```
yachtcaptain23: Might be worth `def protect` to application_controller and adding `before_action :protect`

corymcdonald: There's a before_action that ensures a user is currently logged in

Does that satisfy the intended purpose of the `protect` method?

app/javascript/views/statements/statements/StatementDetails.tsx
```diff
@@ -0,0 +1,160 @@
...
+          {this.props.statement.details.map((detail, index) => (
+            <StatemenDetail detail={detail} index={index} />
+          ))}
+        </div>
+      </Details>
+    );
+  }
+}
+
+const StatemenDetail = props => (
```
yachtcaptain23: Misspelling

app/models/views/user/statements.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+
+      def as_json(*)
+        {
+          overviews: overviews,
+        }
+      end
+
+      def build_statements(publisher, details_date)
+        @overviews = []
```
yachtcaptain23: I think you can keep this as a local variable

app/javascript/views/statements/statements/StatementDetails.tsx
corymcdonald: https://github.com/brave-intl/publishers/pull/2209/commits/1b300b11587f843bc178734ac0de4a8492c8e14e

corymcdonald:
Discussed offline, going to develop a process for localizing rest of application

app/controllers/publishers/statements_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    private
+
+    def publisher
+      return Publisher.find(params[:id]) if current_publisher.admin?
+      current_publisher
+    end
+
+    def protect
```
yachtcaptain23: I think we should move `def protect` from PublisherController to ApplicationController

app/models/views/user/statements.rb
```diff
@@ -0,0 +1,64 @@
...
+
+        # Then we can group the payouts into monthly periods for the statement
+        periods = payouts.group_by { |x| x.created_at.at_beginning_of_month }
+
+        periods.each do |payment_month, payout_entries|
+          # We find the previous month that was paid, or the first statement that was generated
+          period_start = periods.keys.reverse.detect { |k| k < payment_month } || @statements.first.created_at.at_beginning_of_month
+
+          # Sum all the BAT amounts
+          total_amount = payout_entries.sum { |x| x.amount.abs }
```
yachtcaptain23: I think it should be `payout_entries.sum { |x| x.amount > 0 ? x.amount : 0 }`

app/services/publisher_statement_getter.rb
```diff
@@ -1,53 +1,100 @@
...
+
+    # Replicate the similar behavior of Eyeshade. Aggregates all the contribution amounts for the month, and puts them into one entry
+    uphold.group_by { |u| u.created_at.at_end_of_month }.each do |date, entries|
+      entries.group_by { |e| e.channel }.each do |channel, channel_entries|
+        amount = channel_entries.sum { |x| x.amount }
+
+        uphold << Statement.new(
+          channel: channel,
+          transaction_type: Statement::UPHOLD_CONTRIBUTION_SETTLEMENT,
+          amount: -amount,
```
yachtcaptain23: Why is this negative? (Preferable you add a comment in-line to explain)

app/services/publisher_statement_getter.rb
```diff
@@ -1,53 +1,100 @@
...
+          amount: transaction.origin.dig("amount")&.to_d,
+          settlement_currency: transaction.destination.dig("currency"),
+          settlement_amount: transaction.destination.dig("amount")&.to_d,
+          created_at: transaction.createdAt.to_date,
+        )
+      end
+    end
+
+    # Replicate the similar behavior of Eyeshade. Aggregates all the contribution amounts for the month, and puts them into one entry
+    uphold.group_by { |u| u.created_at.at_end_of_month }.each do |date, entries|
```
yachtcaptain23: I'm very confused by this block. So we're going through a list of uphold contributions and somehow the settlement statements gets appended in this block?

yachtcaptain23:
Where does `publishers/statements/statement.html.slim` get called to be rendered?

app/services/publisher_statement_getter.rb
corymcdonald: When we mark a transaction as paid out then we subtract from the Publishers current balance on Eyeshade so the transactions appear as negative. 

In order to actually get a valid grouping for periods for which months we paid out I select all the transactions that are negative, then group them by then select the transactions which are within the start and end of that grouping.

This was the easiest way for finding the ranges for when publishers got paid. So for Uphold it’s a little different. Technically we are have technically already paid these transactions out as they happen in real time. However if we did mark them as negative BAT amounts and the positive amount that the user got paid then our monthly groupings would be wrong. 

In order to keep the same groupings and also include the summary for how many Uphold tips there were I summarize them after I append them to all the transactions.

app/controllers/publishers/statements_controller.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2209/commits/152975f8a26ad1f7223146f0a0b7338b3bdfd1de

app/models/views/user/statements.rb
corymcdonald: I don't think so, basically Eyeshade returns negative amounts (and a positive conversion with currency) when we pay users out, so this sums up all the negative amounts that we paid out.

corymcdonald:
>Where does publishers/statements/statement.html.slim get called to be rendered?

Good catch :) It doesn't get rendered anymore.

app/services/publisher_statement_getter.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2209/commits/35058de3041adc5cf544b4d3bce3a4d4594a66da

app/models/views/user/statements.rb
corymcdonald: I don't think so, because this is what is being serialized and sent as json to the view

PR #2212, opened 2019-09-05
Remove keys for default site banner values

### LINKED ISSUE #2113: removing "0.58.x" warning when selecting tipping categories
We should remove the "0.58.x" as this was due to a bug on the client side when we first released tipping. We only had the ability to tip `1 BAT`, `5 BAT` & `10 BAT` even though a user set a custom category. This was been fixed for a long time now. Example of issue:

Screen Shot 2019-07-31 at 8 53 15 PM

corymcdonald:
#### Features

- if the keys are the same as the default we delete them
- removes the warning text, closes #2113

### DIFF
--- a/app/javascript/packs/style.jsx
+++ b/app/javascript/packs/style.jsx
@@ -40,7 +40,6 @@ export const Dialogue = styled.div`
     props.save &&
     `
     width: 550px;
-    height: 350px;
     margin-left: -275px;
     margin-top: -200px;
     `}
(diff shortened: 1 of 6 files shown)


app/javascript/packs/style.jsx
```diff
@@ -40,7 +40,6 @@ export const Dialogue = styled.div`
     props.save &&
     `
     width: 550px;
-    height: 350px;
```
NejcZdovc: is this intentional? seams unrelated

corymcdonald: Has to do with the height of the panel when removing the warning about 0.58
https://github.com/brave-intl/publishers/issues/2113

PR #2214, opened 2019-09-06
Create show and edit page for payout report

corymcdonald:
#### Features

In order to track which payout reports were actually used to pay users out, it would be nice to have the ability to edit these. In addition view some details about what is happening during the payout.

### DIFF
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -70,7 +70,7 @@ module AdminHelper
 
   def percentage_difference(x,y)
     numerator = (x-y).to_f
-    denominator = ((x+y)/2).to_f
+    denominator = y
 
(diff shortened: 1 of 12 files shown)


app/helpers/admin_helper.rb
```diff
@@ -68,6 +68,22 @@ def case_link(text, path, badge= nil)
     )
   end
 
+  def percentage_difference(x,y)
+    numerator = (x-y).to_f
```
yachtcaptain23: Isn't it `value = (x-y / y) * 100`?

corymcdonald: https://github.com/brave-intl/publishers/pull/2214/commits/a22a93349b9c221c0d4c5dd10f2aeb2b6c836df0

PR #2218, opened 2019-09-09
Bugfix/performance dyno caching

### LINKED ISSUE #2205: Move channels JSON to use shared heap size
Right now each thread fetches from Redis. We can use significantly less memory if they all read from the same heap on the process.

yachtcaptain23:
#### Features
Closes #2205 

#### How To Test

1. Pushed to staging-2
2. Check cache hit vs cache miss on the dyno

### DIFF
--- a/app/controllers/api/v1/public/channels_controller.rb
+++ b/app/controllers/api/v1/public/channels_controller.rb
@@ -1,6 +1,7 @@
 class Api::V1::Public::ChannelsController < Api::V1::Public::BaseController
   include BrowserChannelsDynoCaching
   @@cached_payload = nil
+  REDIS_KEY = 'browser_channels_json'.freeze
 
   def totals
     render(json: Channel.statistical_totals, status: 200)
--- a/app/controllers/api/v1/public/channels_controller.rb
+++ b/app/controllers/api/v1/public/channels_controller.rb
@@ -1,14 +1,16 @@
 class Api::V1::Public::ChannelsController < Api::V1::Public::BaseController
-  def channels
-    channels_json = Rails.cache.fetch('browser_channels_json', race_condition_ttl: 30) do
-      require 'sentry-raven'
-      Raven.capture_message("Failed to use redis cache for /api/public/channels, using DB instead.")
-      channels_json = JsonBuilders::ChannelsJsonBuilder.new.build
-    end
-    render(json: channels_json, status: 200)
-  end
-
+  include BrowserChannelsDynoCaching
   def totals
     render(json: Channel.statistical_totals, status: 200)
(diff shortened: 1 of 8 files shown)


app/controllers/concerns/browser_channels_dyno_caching.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def invalid_dyno_cache?
+    cached_dyno_value = self.class.class_variable_get(klass_dyno_cache)
+    cached_dyno_value.nil? || !cached_dyno_value.is_a?(String)
+  end
+
+  def update_dyno_cache
+    redis_value = Rails.cache.fetch(self.class::REDIS_KEY, race_condition_ttl: 30) do
+      require 'sentry-raven'
```
corymcdonald: nit: Let's require this at the top - I feel like it is better to know what libraries are being used in a module as opposed to in a method.

app/controllers/concerns/browser_channels_dyno_caching.rb
```diff
@@ -0,0 +1,41 @@
...
+      Raven.capture_message("Failed to use redis cache for Dyno cache: #{klass_dyno_cache}, continuing to read from cache instead")
+    end
+    if redis_value.present?
+      self.class.class_variable_set(klass_dyno_cache, redis_value)
+      Rails.cache.write(dyno_expiration_key, 1.hour.from_now.to_s, expires_in: 1.hour.from_now )
```
corymcdonald: We're writing the cache and it's set to expire in an hour, but we run these jobs every 2 hours. Will this be a problem?

app/controllers/api/v1/public/channels_controller.rb
```diff
@@ -1,14 +1,15 @@
...
-  def channels
-    channels_json = Rails.cache.fetch('browser_channels_json', race_condition_ttl: 30) do
-      require 'sentry-raven'
-      Raven.capture_message("Failed to use redis cache for /api/public/channels, using DB instead.")
-      channels_json = JsonBuilders::ChannelsJsonBuilder.new.build
-    end
-    render(json: channels_json, status: 200)
-  end
+  include BrowserChannelsDynoCaching
+  @@cached_payload = nil
```
corymcdonald: Do we need to set this to nil in every class? Won't it be nil be default?

app/controllers/concerns/browser_channels_dyno_caching.rb
yachtcaptain23: The expiration is for the dyno, so worst case scenario is about 3 hours.

PR #2219, opened 2019-09-10
Create Uphold Report

### LINKED ISSUE #2215: Track and Generate a report the KYC Date for a publisher
Business Development should be able to generate a report which contains the publisher_id, uphold_id, kyc_at information. 

edgecase: if they disconnect, the creation dates at our end stick so should not matter

corymcdonald:
## Create Uphold Report

For Business Development Purposes we will have a report that they can download via the admin dashboard to view all the uphold ids that were connected.

Because Uphold Connections can change owners and be disconnected it made more sense to create a separate db table that would track anytime a new connection was created. 

Closes #2215


db/migrate/20190910163930_create_uphold_report.rb
```diff
@@ -0,0 +1,13 @@
+class CreateUpholdReport < ActiveRecord::Migration[5.2]
+  def change
+    add_column :uphold_connections, :member_at, :datetime
+
+    create_table :uphold_reports, id: :uuid, default: -> { "uuid_generate_v4()"}, force: :cascade do |t|
+      t.belongs_to :publisher, index: true, type: :uuid
+      t.uuid :uphold_id
```
yachtcaptain23: Let's create an index here too to improve the above `find_by`

app/models/uphold_report.rb
```diff
@@ -0,0 +1,3 @@
+class UpholdReport < ApplicationRecord
```
yachtcaptain23: Nitpicking, but internally the convention that would be used is `UpholdStatusReport`

corymcdonald: https://github.com/brave-intl/publishers/pull/2219/commits/c35cb3d2f276a0c1b81ee111a38d9f4b56e8744e

db/migrate/20190910163930_create_uphold_report.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2219/commits/c35cb3d2f276a0c1b81ee111a38d9f4b56e8744e

app/controllers/admin/uphold_reports_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+        group('(EXTRACT(MONTH FROM created_at))::integer').
+        order('2 DESC, 3 DESC').count
+    end
+
+    def show
+      date = DateTime.strptime(params[:id], "%Y-%m")
+      start_date = date.at_beginning_of_month
+      end_date = date.at_end_of_month
+
+      @uphold_report = UpholdStatusReport.where("created_at >= :start AND created_at <= :finish", start: start_date, finish: end_date)
```
yachtcaptain23: This shouldn't be using `@`. Prefer renaming from `@uphold_report` to `uphold_status_report`.

app/controllers/admin/uphold_reports_controller.rb
```diff
@@ -0,0 +1,30 @@
+require 'csv'
+
+module Admin
+  class UpholdStatusReportsController < AdminController
+    def index
+      @uphold_report = UpholdStatusReport.
```
yachtcaptain23: Prefer renaming to `@uphold_status_reports`

app/views/admin/uphold_reports/index.html.slim
```diff
@@ -0,0 +1,23 @@
...
+  p Anytime a publisher connects their uphold account we create an entry in our database that logs this. If an uphold id has already been reported we do not create an entry. This data is then aggregated and grouped by month and displayed here.
+
+table.table
+  thead
+    tr
+      td Period
+      td Number of users
+      td
+  tbody
+    - @uphold_report.each do |report|
```
yachtcaptain23:
```suggestion
    - @uphold_reports.each do |report|
```

app/controllers/admin/uphold_reports_controller.rb
yachtcaptain23:
```suggestion
      uphold_status_report = UpholdStatusReport.where("created_at >= :start AND created_at <= :finish", start: start_date, finish: end_date)
```

app/controllers/admin/uphold_reports_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+      date = DateTime.strptime(params[:id], "%Y-%m")
+      start_date = date.at_beginning_of_month
+      end_date = date.at_end_of_month
+
+      @uphold_report = UpholdStatusReport.where("created_at >= :start AND created_at <= :finish", start: start_date, finish: end_date)
+
+      generated = []
+      generated << ["publisher id", "publisher created at", "uphold id", "uphold connected time"].to_csv
+
+      @uphold_report.each do |report|
```
yachtcaptain23:
```suggestion
      uphold_status_reports.each do |report|
```

app/controllers/admin/uphold_reports_controller.rb
```diff
@@ -0,0 +1,30 @@
+require 'csv'
+
+module Admin
+  class UpholdStatusReportsController < AdminController
+    def index
+      @uphold_status_reports = UpholdStatusReport.
+        group('(EXTRACT(YEAR FROM created_at))::integer').
+        group('(EXTRACT(MONTH FROM created_at))::integer').
+        order('2 DESC, 3 DESC').count
```
yachtcaptain23:
```suggestion
        order('2 DESC, 3 DESC')
```

corymcdonald: Why this change? Since we're doing a `group` I want get a count of the amount in each report

PR #2222, opened 2019-09-12
Updated Dockerfile to not consume excess entries

yachtcaptain23:
## Pull Request Name

#### Features

#### How To Test

1. Step one
2. Step two
3. Step three

#### Pull Request Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,12 @@
-FROM ruby:2.4.5
+FROM ruby:2.5.5
 
 # Install node
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
+RUN curl -sL https://deb.nodesource.com/setup_11.x | bash -
 
-RUN apt-get update -qq && apt-get install -y build-essential libpq-dev nodejs libxml2 iceweasel xvfb fonts-liberation \
- libappindicator3-1 libnspr4 libnss3 libxss1 xdg-utils gdb chromium chromium-l10n
+RUN apt-get update
+RUN apt install nodejs
+RUN node --version
+RUN npm --version
 
 # Install Chrome
 RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
--- a/README.md
+++ b/README.md
@@ -335,9 +335,6 @@ top of the repo. For example you can expose ports on your system for the databas
 version: "2.1"
 
 services:
-  mongo:
-    ports:
-      - "27017:27017"
   redis:
     ports:
       - "6379:6379"
@@ -346,7 +343,7 @@ services:
       - "5432:5432"
 ```
 
-to start with docker build the app and eyeshade images
+to start with docker build the app
 
 ```sh
 docker-compose build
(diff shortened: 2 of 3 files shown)


corymcdonald:
Can you update the README with the most up to date information?

corymcdonald:
### Dockerfile

If Docker is not intended to be used for local development of Publishers then we should consider making our `Dockerfile` optimized for production mode. (Or having a Dockerfile-production [similar to this article](https://anonoz.github.io/tech/2018/05/01/rails-dockerfile.html))

We should do this because

1. When we deploy the Docker image then want it to be production-ready.
2. We want the image to be the smallest possible so it is quicker to transfer to build, deploy, store, and maintain.
3. Asset minification and production optimization techniques improves quality of life for our users.

Following this line of thought we should make the following modifications

#### 1. Remove the Chrome installation section, this is exclusively used for tests and explodeds the Docker image file

```Docker
# Install Chrome
RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
RUN dpkg -i google-chrome-stable_current_amd64.deb; apt-get -fy install

# Symlink google-chrome to chrome so our tests can find it
RUN ln -sf /usr/bin/google-chrome /usr/bin/chrome
```

#### 2. Run in production mode.

This is essential, if we don't run in production mode then this can potentially expose API tokens, debug information, access to environment, and other really bad. So we need to have this be configured on the image.

```Docker
ENV RAILS_ENV production
ENV RAILS_LOG_TO_STDOUT true
ENV NODE_ENV production
```

#### 3. Only put into the image what we really need to put into the image.

**We don't specify bundle install to only install items in production mode**

We should not install development or test dependencies when running in production. Also let's remove the cache for those gems because bundle install will not be run again on this image. We should also precompile the assets similar to what is done during Heroku's build process.

```Docker
bundle install --without development test \
&& rm -rf /usr/local/bundle/bundler/gems/*/.git \
  /usr/local/bundle/cache/


bundle exec rails assets:precompile
```

**We should create a .dockerignore so we don't include node_modules and other things into our images.**

Docker is including node_modules, logs, etc unless we choose to ignore them in dockerignore. This is fine if our image was being built on Jenkins, or another CI system that pulled our image fresh but if we're building locally it will mess up some of the testing.

I simply copy/pasted our .gitignore into a new folder called .dockerignore, and that seemed to work pretty well.


####  4. Consider switching to a slimmer base image.

If we're considered about image size then we can consider switching over to alpine linux, or `ruby-slim` from the same base image. It seems like quite a few companies use Alpine linux due to sizing constraints but that doesn't mean that we need to adopt.

### General README Updates

We should have instructions on how to build for Production mode locally. Right now Docker-compose is used for a multi-container system. We should take into consideration that we will not be bundling all these other containers into our image, and instead be injecting these via environment variables like we have been doing with Heroku.

Typically I've built docker images for production with the simple command,

```sh
docker build --no-cache -t publishers .
```

#### Environment Variables

We should also have a minimum number of environment variables required for building the image locally. Things like the Sentry initializer raising an exception when there's not an environment variable is frustrating.

For example I had to set both `SENTRY_DSN`, and `UPHOLD_AUTHORIZATION_ENDPOINT` in my test Dockerfile just to run `assets:precompile`.

This is pretty annoying and probably going to be an issue later.


### We should submit a security review for this Dockerfile

We're going to be running this code in production, they should review all the dependencies and ensure everything is secure from the base images to the way we're compiling assets and removing unused packages.j

### Consider moving our Heroku builds over to Docker too

Once we have a production instance running we should also consider building docker images with our heroku box so that everything is using the same build and image system.

We can use the following article for reference for how to do this.

https://devcenter.heroku.com/articles/build-docker-images-heroku-yml



These notes and research led me with this form this Dockerfile, it might need some tweaking once you start deploying to AWS.

```Docker
FROM ruby:2.5.5

RUN apt-get update

# Install node
RUN curl -sL https://deb.nodesource.com/setup_11.x | bash -

RUN apt install nodejs
RUN node --version
RUN npm --version

RUN gem install bundler foreman mailcatcher
RUN npm install -g yarn

ENV RAILS_ENV production
ENV RAILS_LOG_TO_STDOUT true
ENV NODE_ENV=production

# Enabling app reloading based off of https://stackoverflow.com/questions/37699573/rails-app-in-docker-container-doesnt-reload-in-development
# Sets the path where the app is going to be installed
ENV RAILS_ROOT /app

# Creates the directory and all the parents (if they don't exist)
RUN mkdir -p $RAILS_ROOT

# This will be the de-facto directory where all the contents are going to be stored.
WORKDIR $RAILS_ROOT

COPY Gemfile ./
COPY Gemfile.lock ./

# Install the gems.
RUN bundle install --without development test && rm -rf /usr/local/bundle/bundler/gems/*/.git \
  /usr/local/bundle/cache/

COPY package.json yarn.lock ./

COPY . ./

RUN yarn install && yarn build
RUN SECRET_KEY_BASE=foo bundle exec rails assets:precompile
RUN SECRET_KEY_BASE=foo bundle exec rails assets:clean

EXPOSE 3000
CMD ["bundle", "exec", "rails", "server", "-b", "0.0.0.0"]
```

PR #2230, opened 2019-09-16
Make payout process more transparent

corymcdonald:
## Make payout process more transparent

image
image
image
image
image

#### Features

Frequently users complain that: 
-they don't know how much money they're going to receive
- how that amount was determined
- why they did not get paid
- why payout takes so long

Hence I spent sometime creating a payout progress bar. This has great features such as
- Letting users know where they're at in the process
- How much they will earn
- If they need to re-verify, KYC or something else to receive funds
- Increases progress per day 🎉 

#### How To Test

1. Generate final payout report
2. Increase through the days by looking at `payout_helper.rb`

### DIFF
--- a/app/helpers/payout_helper.rb
+++ b/app/helpers/payout_helper.rb
@@ -1,4 +1,10 @@
 module PayoutHelper
+
+  PREPARING = I18n.t('.publishers.payout_status.statuses.preparing')
+  REVIEWING = I18n.t('.publishers.payout_status.statuses.reviewing')
+  IN_PROGRESS = I18n.t('.publishers.payout_status.statuses.in_progress')
+  DONE = I18n.t('.publishers.payout_status.statuses.done')
+
   def icon_class(statuses, current, index)
     selected = index <= statuses.find_index(current)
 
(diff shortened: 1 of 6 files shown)


app/helpers/payout_helper.rb
```diff
@@ -0,0 +1,73 @@
...
+      transition_index = transitions.keys.find_index(current) - 1
+      base = transitions.keys[transition_index]
+      base = 0 if transition_index < 0
+
+      percent = ((difference - base).to_f / (current - base).to_f).to_f
+    end
+
+    [
+      I18n.t(".publishers.payout_status.statuses.#{status}"),
+      percent
```
yachtcaptain23: You need to instantiate this variable where `status != 'done'`

app/helpers/payout_helper.rb
```diff
@@ -0,0 +1,73 @@
...
+    selected = index <= statuses.find_index(current)
+
+    return 'inactive' unless selected
+    return "animated" if  statuses[index] == current && current != I18n.t(".publishers.payout_status.statuses.done")
+
+    "active"
+  end
+
+  def current_status_and_percent(report_created_at)
+    today = DateTime.now
```
yachtcaptain23: I prefer refactoring to make it more readable

e.g.
```
status = nil
status = DONE if !Rails.cache.fetch('payout_in_progress')
if status.blank
days_ago = (Date.today - report_created_at.to_date).to_i
if days_ago < 3
  status = PREPARING
  progress_percentage = ((days_ago / 3) * 100).to_i
elsif days_ago < 7
  status = REVIEWING
  progress_percentage = (((days_ago - 3) / 4) * 100).to_i
elsif days_ago < 11
  status = IN_PROGRESS
  progress_percentage = (((days_ago - 7) / 4) * 100).to_i
else
  status = DONE
end

return [I18n.t(".publishers.payout_status.statuses.done"), nil] if status == DONE
```

corymcdonald: I like this, my original approach was very similar, but I thought having a bunch of if statements might make it confusing - however looks better than the current implementation. 

One of the things I'd eventually want is to be able to display the payout progress tab for a few days after the payout is complete. If the payout in progress was removed on a certain day it would be helpful to automatically switch over to "Done" and show the end users.

yachtcaptain23: Probably something in Redis is easiest or adding status to `payout_reports`

PR #2243, opened 2019-09-20
Add japanese locale to creators landing page

yachtcaptain23:


corymcdonald:
I'm seeing some issue with the links not working, and it being referred to as [object Object] this might be an issue with the react `<FormatMessage>` component. We might need to use the underlying function `intl.formatMessage({ id: 'home.menu.link' })` or something similar. 

The sign-up / log-in pages don't look like they're translated so we need to tackle that too

PR #2258, opened 2019-09-26
Japanese localization via param headers

yachtcaptain23:
Related to https://github.com/brave-intl/publishers/issues/2186

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -44,6 +44,8 @@ class ApplicationController < ActionController::Base
   end
 
   def set_locale
+    p "albert #{request.headers}"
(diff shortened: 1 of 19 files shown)


app/controllers/application_controller.rb
```diff
@@ -20,12 +19,21 @@ class ApplicationController < ActionController::Base
...
+      params[:locale]
+    elsif extract_locale_from_accept_language_header == 'ja'
+      redirect_to(request.original_url + "?locale=ja") and return
+    else
+      I18n.default_locale
+    end
     I18n.with_locale(locale, &action)
   end
 
+  def default_url_options
```
corymcdonald: Does this mean that locale will be passed through to every single link? I'm not sure we want that and it probably would be better to read from the header

PR #2259, opened 2019-09-26
Check balance for no_grants users

### LINKED ISSUE #2194: Some users in no grants can't load home page
Sentry Issue: [PUBLISHERS-PRODUCTION-T2-9ED](https://sentry.io/organizations/brave-software/issues/1108409690/?referrer=github_integration)

```
NoMethodError: undefined method `-' for nil:NilClass
  app/helpers/publishers_helper.rb:69:in `publisher_converted_overall_balance'
    amount =  publisher.wallet&.overall_balance&.amount_default_currency - publisher.wallet&.contribution_balance&.amount_default_currency
  app/views/publishers/home.html.slim:52:in `_app_views_publishers_home_html_slim___3494719539947285593_47049423333600'
    = publisher_converted_overall_balance(@publisher)
  config/middlewares/http_header_middleware.rb:8:in `call'
    @app.call(env)
...
(151 additional frame(s) were not displayed)

ActionView::Template::Error: undefined method `-' for nil:NilClass

ActionView::Template::Error: undefined method `-' for nil:NilClass
```

yachtcaptain23:
Closes #2194

### DIFF
--- a/app/helpers/publishers_helper.rb
+++ b/app/helpers/publishers_helper.rb
@@ -59,25 +59,27 @@ module PublishersHelper
   def publisher_converted_overall_balance(publisher)
     return if publisher.uphold_connection.default_currency == "BAT" || publisher.uphold_connection.default_currency.blank?
 
-    publisher = publisher&.become_subclass
+    result = I18n.t("helpers.publisher.conversion_unavailable", code: publisher.uphold_connection.default_currency)
+    sentry_catcher do
+      publisher = publisher&.become_subclass
 
-    if publisher.partner?
-      balance = publisher.balance_in_currency
-    elsif publisher.only_user_funds?
-      balance = publisher.wallet&.contribution_balance&.amount_default_currency
-    elsif publisher.no_grants?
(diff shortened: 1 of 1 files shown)


app/helpers/publishers_helper.rb
```diff
@@ -59,25 +59,27 @@ def publisher_overall_bat_balance(publisher)
...
-             amount: '%.2f' % balance,
-             code: publisher&.uphold_connection.default_currency)
-    else
-      I18n.t("helpers.publisher.conversion_unavailable", code: publisher.uphold_connection.default_currency)
+      if publisher.partner?
+        balance = publisher.balance_in_currency
+      elsif publisher.only_user_funds?
+        balance = publisher.wallet&.contribution_balance&.amount_default_currency
+      elsif publisher.no_grants?
+        balance =  publisher.wallet&.overall_balance&.amount_default_currency - publisher.wallet&.contribution_balance&.amount_default_currency
```
corymcdonald: It seems like this add a lot of extra additions to Sentry when we know that the error is that something is going on with the Balance class.

Perhaps it would be better to return `0` if the `amount_default_currency` is nil

corymcdonald:
Let's leave the issue open, but solve the 500 error by doing this fix.

PR #2281, opened 2019-10-03
Don't include partners on the payout report

corymcdonald:
## Don't include partners on the payout report

Partners who we manually create invoices for should not appear in the payout report


corymcdonald:
Thinking about this more, I'm not quite sure this is actually what we want. This would prevent Partners from getting paid out for contributions.

PR #2285, opened 2019-10-03
Update Channel ID when a channel is being transferred

corymcdonald:
## Update Channel ID when a channel is being transferred

If a channel is deleted or re-added like in the case of Uphold.com we need to update the reference to the new channel_id


app/jobs/create_uphold_channel_card_job.rb
```diff
@@ -23,6 +23,7 @@ def perform(uphold_connection_id:, channel_id:)
     upfc.update(
       address: get_address(uphold_connection, card_id),
       card_id: card_id,
+      channel_id: channel_id,
```
yachtcaptain23:
```suggestion
      # It's possible a channel can be removed, so this covers re-linking an existing UCFC to the re-added channel.
      channel_id: channel_id,
```

PR #2330, opened 2019-10-18
Fix Statements

### LINKED ISSUE #2329: Fix Statement Bugs
There are a couple of things that are wrong with the statement workflow.

A good portion of these are related to the Uphold Transactions which are a part of statement. 

- Uphold Transactions are paginated and we've only been requesting 50 or so
- The total earned is a bit confusing, it’s the amount before the fees. I’m propose either we just show the amount BAT that is going to actually be deposited.
- The amount displayed is the settlement amount is in different currencies

corymcdonald:
## Fix statements

Closes #2329 

#### Features

- Fixes problem where earning period wasn't accurate due to Eyeshade's created_at time being a month ahead
- Fixes pagination of transactions being missing
- Fixes uncertainty about total earned

One thing I'm thinking about adding is ensuring that there isn't different settlement amounts, like USD and BAT being changed in the middle of the month. This is an edge case i'm not quite sure we should handle yet. 

#### How To Test

Message me for transactions and stuff :)

### DIFF
--- a/app/services/publisher_statement_getter.rb
+++ b/app/services/publisher_statement_getter.rb
@@ -7,6 +7,23 @@ class PublisherStatementGetter < BaseApiClient
 
     UPHOLD_CONTRIBUTION = "uphold_contribution".freeze
     UPHOLD_CONTRIBUTION_SETTLEMENT = "uphold_contribution_settlement".freeze
+
+    def fee?
+      transaction_type == 'fees'
+    end
+
+    def eyeshade?
+      transaction_type == 'contribution_settlement' || transaction_type == 'referral_settlement' || fee?
+    end
+
+    def earning_period
+      # If the transaction_type is from Eyeshade this means the period was for the previous month
+      if eyeshade?
+        (created_at - 1.month).at_beginning_of_month
+      else
+        created_at.at_beginning_of_month
+      end
+    end
   end
 
   def initialize(publisher:)
@@ -70,19 +87,23 @@ class PublisherStatementGetter < BaseApiClient
 
     # Replicate the similar behavior of Eyeshade. Aggregates all the contribution amounts for the month, and puts them into one entry
     uphold.group_by { |u| u.created_at.at_end_of_month }.each do |date, entries|
+      # Group by channels so we can show tips per channel
       entries.group_by { |e| e.channel }.each do |channel, channel_entries|
-        amount = channel_entries.sum { |x| x.amount }
+        # Finally group by currency, the currency can change in the middle of the month for direct tips but likely it will just be 1.
+        channel_entries.group_by { |c| c.settlement_currency }.each do |currency, currency_entries|
+          amount = channel_entries.sum { |x| x.amount }
 
-        # We're specifying a negative amount because we group by transactions already paid out.
-        # This gives us the ability to aggregate and show one Uphold transaction, rather than 300 or so tips that might have been sent.
-        uphold << Statement.new(
-          channel: channel,
-          transaction_type: Statement::UPHOLD_CONTRIBUTION_SETTLEMENT,
-          amount: -amount,
-          settlement_currency: "BAT",
-          settlement_amount: amount,
-          created_at: date,
-        )
+          # We're specifying a negative amount because we group by transactions already paid out.
+          # This gives us the ability to aggregate and show one Uphold transaction, rather than 300 or so tips that might have been sent.
+          uphold << Statement.new(
+            channel: channel,
+            transaction_type: Statement::UPHOLD_CONTRIBUTION_SETTLEMENT,
+            amount: -amount,
+            settlement_currency: currency,
+            settlement_amount: amount,
+            created_at: date,
+          )
+        end
       end
     end
 
--- a/app/services/publisher_statement_getter.rb
+++ b/app/services/publisher_statement_getter.rb
@@ -19,7 +19,7 @@ class PublisherStatementGetter < BaseApiClient
     def earning_period
       # If the transaction_type is from Eyeshade this means the period was for the previous month
       if eyeshade?
-        (created_at - 1.month).at_beginning_of_month
+        created_at.prev_month.at_beginning_of_month
       else
         created_at.at_beginning_of_month
       end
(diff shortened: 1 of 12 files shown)


app/services/uphold/models/transaction.rb
```diff
@@ -28,9 +28,28 @@ def initialize(params = {})
...
+          ending = start + batch_size
+
+          response = get(PATH.expand(id: id), {}, authorization(@uphold_connection), { "Range" => "items=#{start}-#{ending}" })
+
+          transactions << parse_response(response).map { |a| Uphold::Models::Transaction.new(a) }
+
+          range = response.headers["content-range"]
+          total_pages = range.split('/').second.to_i
+
+          break if total_pages < ending
```
yachtcaptain23: This doesn't seem right at first glance, given that `ending` measures based on entries and `total_pages` is based on `page_index`

app/services/uphold/models/transaction.rb
```diff
@@ -28,9 +28,28 @@ def initialize(params = {})
       def all(id:)
         Rails.logger.info("Connection #{@uphold_connection.id} is missing uphold_access_parameters") and return if @uphold_connection.uphold_access_parameters.blank?
 
-        response = get(PATH.expand(id: id), {}, authorization(@uphold_connection))
+        index = 1
```
yachtcaptain23: Prefer to rename this to `page_index`

app/services/uphold/models/transaction.rb
```diff
@@ -28,9 +28,28 @@ def initialize(params = {})
...
+        batch_size = 50
 
-        parse_response(response).map { |a| Transaction.new(a) }
+        transactions = []
+
+        loop do
+          start = (batch_size * index) - batch_size
+          ending = start + batch_size
+
+          response = get(PATH.expand(id: id), {}, authorization(@uphold_connection), { "Range" => "items=#{start}-#{ending}" })
```
yachtcaptain23: Did you check to make sure this is exclusive of ending, i.e. it seems to query from 0 to 50 and then 50 to 100.

corymcdonald: Yes, per the Uphold API documentation -
```
The endpoints that support pagination returns a Content-Range header. For instance, if you make a request with Range: items=0-4 header the response will contain the following header: Content-Range: 0-4/* where * will be the total number of items that this endpoint can return.```

File: app/services/uphold/models/transaction.rb (hunk 1 above)
Comment (corymcdonald) [reply app/services/uphold/models/transaction.rb]: I agree, just looking at this doesn't make a lot of sense, but breaking out the loop 

```ruby
# 1st loop

start = 0
ending = 50
total_pages = 137
break if total_pages < ending

# 2nd loop
start = 50
ending = 100
total_pages = 137
break if 137 < 100


# 3rd loop
start = 100
ending = 150
total_pages = 137
break if 137 < 150
```

So I think the variables should be renamed maybe to something like `total_items` and then rename `ending` to be  `ending_index`

app/services/uphold/models/transaction.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2330/commits/a6d89d6d16ca96edacdbb20bbfc71019dd3f54bd

app/services/uphold/models/transaction.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2330/commits/a6d89d6d16ca96edacdbb20bbfc71019dd3f54bd

app/services/publisher_statement_getter.rb
```diff
@@ -7,6 +7,23 @@ class Statement
...
+    end
+
+    def eyeshade?
+      transaction_type == 'contribution_settlement' || transaction_type == 'referral_settlement' || fee?
+    end
+
+    def earning_period
+      # If the transaction_type is from Eyeshade this means the period was for the previous month
+      if eyeshade?
+        (created_at - 1.month).at_beginning_of_month
```
yachtcaptain23: Would you double check to prove this works all the time? Is 1.month automagically the previous month? Would `1.month.ago` provide a different answer?

yachtcaptain23:
```suggestion
        Date.today.prev_month.at_beginning_of_month
```

app/services/publisher_statement_getter.rb
```diff
@@ -7,6 +7,23 @@ class Statement
...
+      else
+        created_at.at_beginning_of_month
```
yachtcaptain23: We don't need the time and timezone right?

```suggestion
        created_at.to_date.at_beginning_of_month
```

app/services/publisher_statement_getter.rb
corymcdonald: For future reference - 
1.month is automagically the correct month before, [here's the implementation](https://github.com/rails/rails/blob/9ae511f55047b2380224ad993a7d0ed97dbcf944/activesupport/lib/active_support/core_ext/integer/time.rb#L20), from what I can recall it basically subtracts 28-29ish days from the date? 

Doing `.ago` converts it from `Date.today` and looks to [be implemented here](https://api.rubyonrails.org/v5.1/classes/Time.html#method-i-ago)

app/services/publisher_statement_getter.rb
corymcdonald: Going to leave this as-is because`to_date` was already called on `created_at` as shown in the code below

#### In the normal Eyeshade based transactions
https://github.com/brave-intl/publishers/blob/bbe717855ddb41627cdc8ded5921589b133d6d37/app/services/publisher_statement_getter.rb#L64

#### In the Uphold-based transactions
https://github.com/brave-intl/publishers/blob/bbe717855ddb41627cdc8ded5921589b133d6d37/app/services/publisher_statement_getter.rb#L83

app/services/publisher_statement_getter.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2330/commits/341abda2eda9fe5bb65df58dd1dbefa2e33bf312

PR #2337, opened 2019-10-23
Bugfix/promo registration errors rebase

yachtcaptain23:
## Pull Request Name

#### Features

#### How To Test

1. Step one
2. Step two
3. Step three

#### Pull Request Checklist


app/controllers/promo_registrations_controller.rb
```diff
@@ -14,20 +14,11 @@ def create
...
-        Promo::PublisherChannelsRegistrar.new(publisher: @publisher).perform
-        @promo_enabled_channels = @publisher.channels.joins(:promo_registration)
-        PromoMailer.promo_activated_2018q1_verified(@publisher, @promo_enabled_channels).deliver
-      end
-    else
-      PromoMailer.promo_activated_2018q1_unverified(@publisher).deliver
+    current_publisher.channels.left_joins(:promo_registration).where(promo_registrations: {id: nil}).find_each do |channel|
+      channel.register_channel_for_promo # Callee does a check
     end
+    @promo_enabled_channels = @publisher.channels.joins(:promo_registration)
```
corymcdonald: When a user adds channels and then activates the promo this code doesn't behave as expected. 
There's a race condition where the jobs haven't completed to show the referral codes on the screen so the user gets a list of blank referral codes.


Then if you refresh the screen you get this

app/controllers/promo_registrations_controller.rb
```diff
@@ -14,20 +14,11 @@ def create
...
-        Promo::RegisterPublisherForPromoJob.perform_later(publisher: @publisher)
-        redirect_to home_publishers_path, notice: t("promo.activated.please_wait")
-      else
-        Promo::PublisherChannelsRegistrar.new(publisher: @publisher).perform
-        @promo_enabled_channels = @publisher.channels.joins(:promo_registration)
-        PromoMailer.promo_activated_2018q1_verified(@publisher, @promo_enabled_channels).deliver
-      end
-    else
-      PromoMailer.promo_activated_2018q1_unverified(@publisher).deliver
+    current_publisher.channels.left_joins(:promo_registration).where(promo_registrations: {id: nil}).find_each do |channel|
```
corymcdonald: It wasn't exactly obvious what this query was doing to me. Since the `AssignPromoToChannelService` job is checking to see if the channel is verified and there isn't a promo_registration already on the channel can we simplify this and call `register_channel_for_promo` on all channels so this code is easier to understand?

app/controllers/promo_registrations_controller.rb
yachtcaptain23: Will fix to do promo creation as a background job

app/models/channel.rb
```diff
@@ -267,6 +267,10 @@ def uphold_connection
     @uphold_connection ||= uphold_connection_for_channel.detect { |connection| connection.currency == publisher.uphold_connection.default_currency }
   end
 
+  def register_channel_for_promo
+    Promo::RegisterChannelForPromoJob.perform_now(channel_id: id)
```
corymcdonald: I believe moving this simply pass the `id` is going to lead to a race condition due to the `after_save` callback on this method.

Before we were passing `self` into the registration job which wasn't reliant on the database. 

Now that we're passing the `id` the `Channel.new.save` transaction may not yet be committed when we call `perform_now` and so when we call `Channel.find` in the `RegisterChannelForPromoJob` the lookup will fail. 

Since this is done synchronously the job never enters the queue and so this will actually fail to let people create new channels. 

So I think the best approach is to move this to an `after_commit` callback

PR #2340, opened 2019-10-23
Don't run certain jobs in dev

yachtcaptain23:


corymcdonald:
I would prefer if we didn't do it in the actual code and rather chang `sidekiq.yml` only include jobs we want in production to be in production - something like this


```yml
---
:concurrency: 5
:timeout: 8
:max_retries: 3
:queues:
  - mailers
  - default
  - scheduler
  - low
:schedule: &scheduled_jobs
  CleanStaleUpholdDataJob:
    cron: "0 4 * * *"
    queue: scheduler
  CleanAbandonedSiteChannelsJob:
    cron: "0 3 * * *"
    description: "Remove abandoned site channels."
  TransferChannelsJob:
    cron: "0 8 * * *"
    description: "Complete transfer of channels that have exceeded their timeout time without being rejected"

# MORE JOBS HERE


production:
  :schedule:
    <<: *scheduled_jobs
    EnqueuePublishersForPayoutNotificationJob:
      cron: "0 0 23 * *"
      description: "Notifies publishers their wallet is disconnected on the 23rd of every month."
      queue: scheduler
    EnqueuePublishersForPayoutJob:
      cron: "0 0 1 * *"
      description: "Generates the list of unsettled transactions on the first of every month"
      queue: scheduler
    EnqueueSiteChannelVerifications:
      cron: "42 * * * *"
      decription: "For Site Channels created within the past week, enqueue jobs to verify the domain of each unique brave_publisher_id."
      queue: scheduler

```

corymcdonald:
Glad this is separated into environments however, this isn't really how I envisioned it working

It looks like with the changes you proposed we won't be running any jobs in dev at all. However, I want some of these jobs to run in dev, especially things related to payout like `EnqueuePublishersForPayoutJob`, `SetPayoutInProgressJob`, etc. 


In my original comment I specified a name for the standard configuration like so: `:schedule: &scheduled_jobs`. And then later reference that configuration like so
```yaml
production:
  :schedule:
    <<: *scheduled_jobs
    # Production-specific jobs go below this line
```

This was to prevent the same configuration being put into 2 places, as if we have it in two places it will get out of sync very quickly.

Can you ensure 
  - We have a list of default jobs that are run in all environments
  - Production inherits from the default jobs list
  - Staging inherits from production (We should work on removing staging environment [it's best practice to use RAILS_ENV=production even for staging environments](https://devcenter.heroku.com/articles/multiple-environments#managing-staging-and-production-configurations))

yachtcaptain23:
Moving this to a separate PR

PR #2348, opened 2019-10-28
Prevent Duplicate Channels for same publishers

### LINKED ISSUE #2117: Prevent duplicate channels
Let's add an unique index on both the publisher id and the channel identifier on the Channels table to prevent channels from being duplicate.

In addition let's go through and delete the existing duplicated channels

corymcdonald:
## Prevent Duplicate Channels for same publishers

Closes #2117

#### Features

- Adds ability for administrators to delete Publisher's channel (with record)
- Add additional validation to prevent duplicate channels

### DIFF
--- a/app/views/admin/channels/destroy.js.erb
+++ b/app/views/admin/channels/destroy.js.erb
@@ -1,2 +1,1 @@
-console.log("<%= @channel_id %>")
 document.getElementById("<%= @channel_id %>").remove();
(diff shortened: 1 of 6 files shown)


app/views/admin/channels/destroy.js.erb
```diff
@@ -0,0 +1,2 @@
+console.log("<%= @channel_id %>")
```
yachtcaptain23: Debug

PR #2349, opened 2019-10-28
Show total amount during KYC emails

### LINKED ISSUE #2346: Put the amount the publisher will be paid out in the monthly KYC email
When a publisher receives the email that they have to KYC with Uphold in order to be paid out we should specify the amount BAT that they are to be paid out if they KYC in time.

corymcdonald:
## Show total amount during KYC emails
Closes #2346

#### Features

- Put the total amount of BAT that a user will be paid if they verify their Uphold account

### DIFF
--- a/app/services/payout_report_publisher_includer.rb
+++ b/app/services/payout_report_publisher_includer.rb
@@ -73,16 +73,16 @@ class PayoutReportPublisherIncluder < BaseService
 
     # Notify publishers that have money waiting, but will not will not receive funds
     if total_probi > PROBI_THRESHOLD && @should_send_notifications
-      send_emails(uphold_connection)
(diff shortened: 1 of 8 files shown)


app/services/payout_report_publisher_includer.rb
```diff
@@ -73,16 +73,16 @@ def perform
 
     # Notify publishers that have money waiting, but will not will not receive funds
     if total_probi > PROBI_THRESHOLD && @should_send_notifications
-      send_emails(uphold_connection)
+      send_emails(uphold_connection, probi_to_bat(total_amount))
```
yachtcaptain23: Where's `total_amount` coming from?

corymcdonald: good catch, should have been total_probi!

PR #2362, opened 2019-11-05
Upgrade handlebars version

yachtcaptain23:


yachtcaptain23:
FYI I ran `yarn-deduplicate` on `yarn.lock` due to a possible circular dependency.

PR #2363, opened 2019-11-06
Feature: Paypal UI

yachtcaptain23:


app/services/paypal/refresh_identity.rb
```diff
@@ -0,0 +1,53 @@
...
+    raise "Access token is nil for publisher: #{@publisher.id}"
+  end
+
+  def fetch_access_token
+    result = Faraday.post("#{Rails.application.secrets[:paypal_api_uri]}/v1/oauth2/token") do |req|
+      req.headers['Authorization'] = "Basic " + Base64.encode64("#{Rails.application.secrets[:paypal_client_id]}:#{Rails.application.secrets[:paypal_client_secret]}").rstrip.tr("\n", "")
+      req.headers['Accept'] = 'application/json'
+      req.headers['Accept-Language'] = "en_US"
+      req.params['grant_type'] = "refresh_token"
+      req.params['refresh_token'] = @paypal_connection.refresh_token
```
corymcdonald: Do you need to update the current `refresh_token` in the model once you get a new refresh token?

app/controllers/publishers/paypal_accounts_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      user_info_response = Faraday.get("#{Rails.application.secrets[:paypal_api_uri]}/v1/identity/oauth2/userinfo") do |req|
+        req.headers['Authorization'] = "Bearer #{access_token}"
+        req.headers['Accept'] = 'application/json'
+        req.params['schema'] = "paypalv1.1"
+      end
+
+      user_info = JSON.parse(user_info_response.body)
+      paypal_connection = PaypalConnection.find_or_initialize_by(
+        user_id: current_publisher.id,
+        email: user_info['emails'][0]['value']
```
corymcdonald: Let's use `.dig` here for a more safe way to access these values.

(Be sure to update the values below to use dig as well)

app/services/paypal/refresh_identity.rb
```diff
@@ -0,0 +1,53 @@
...
+    @paypal_connection = @publisher.paypal_connection
+  end
+
+  def perform
+    return if @paypal_connection.nil? || @paypal_connection.refresh_token.nil?
+    begin
+      access_token = fetch_access_token
+      raise_token_exception if access_token.nil?
+    rescue => e
+      require 'sentry-raven'
```
corymcdonald: You shouldn't need to require this because it's included during initialization of Rails 🌈 
```suggestion

```

app/services/paypal/refresh_identity.rb
```diff
@@ -0,0 +1,53 @@
...
+      req.params['schema'] = "paypalv1.1"
+    end
+
+    user_info = JSON.parse(user_info_response.body)
+    paypal_connection = PaypalConnection.find_by(
+      user_id: @publisher.id,
+      hidden: false
+    )
+    paypal_connection.update(
+      email: user_info['emails'][0]['value'],
```
corymcdonald: Let's use `.dig` for accessing values in this class similarly to the controller.

app/views/publishers/_paypal_account_panel.html.slim
```diff
@@ -0,0 +1,59 @@
...
+          = @publisher.paypal_connection.email
+        .payout-account-disconnect-button
+          = link_to t(".disconnect"), paypal_account_disconnect_path(@publisher.paypal_connection.id)
+      br
+  - if @publisher.paypal_connection.present? 
+    - if @publisher.paypal_connection.verified_account?
+      .dashboard-panel--content#last_settlement class=last_settlement_class(@publisher)
+        .last-deposit-v2.d-flex.justify-content-between
+          .last-deposit-v2--label
+            = t ".last_deposit"
```
corymcdonald: nit: let's use similar style for all these translations
```suggestion
            = t(".last_deposit")
```

app/views/publishers/_paypal_account_panel.html.slim
```diff
@@ -0,0 +1,59 @@
...
+              div.last-deposit-v2--date
+                label = t(".last_deposit_date") + publisher_last_settlement_date(@publisher, params[:locale])
+    - else
+      .dashboard-panel--content#last_settlement
+        .last-deposit-v2.last-deposit-v2--no-payout-account
+          = t ".need_bank_account"
+  .wide-action
+    - if @publisher.paypal_connection.nil?
+      = link_to t(".connect_with_paypal"),
+        "#{Rails.application.secrets[:paypal_api_uri]}/connect?flowEntry=static&client_id=#{Rails.application.secrets[:paypal_client_id]}&scope=openid email address https%3A%2F%2Furi.paypal.com%2Fservices%2Fpaypalattributes&redirect_uri=https%3A%2F%2Fbat-publishers-dev.herokuapp.com%2Fpublishers%2Fconnect_callback",
```
corymcdonald: I think this should be in a helper, it looks very strange in a view.

Also concerned that I see `bat-publishers-dev.herokuapp.com` the callback URL as this value will change between environments

app/views/publishers/home.html.slim
```diff
@@ -102,63 +102,11 @@ script type="text/html" id="confirm_default_currency_modal_wrapper"
...
-                    - unless last_deposit_balance == I18n.t("helpers.publisher.no_deposit")
-                      span.bat-code= " BAT"
-                  .converted#last_deposit_converted_amount
-                    = publisher_converted_last_settlement_balance(@publisher)
-                = link_to(t(".uphold.check_balance"), uphold_dashboard_url, class: "btn btn-primary",
-                        id: "uphold_check_balance", :"data-piwik-action" => "CheckUpholdBalanceClicked",
-                        :"data-piwik-name" => "Clicked", :"data-piwik-value" => "Dashboard")
-
+      .col-md.mb-4
+        - if !(@publisher.excluded_from_payout? || @publisher.paypal_locale?(params[:locale]))
```
corymcdonald: This logic confused me at first - I thought If a publisher is excluded from payout they will be shown the PayPal information. Can we nest the `if` statement so it's easier to read?

```slim
- unless @publisher.excluded_from_payout?
  - if @publisher.paypal_locale?(params[:locale]
     = render partial: "paypal_account_panel"
  - else 
    = render partial: "uphold_wallet_panel"
```

app/controllers/publishers/paypal_accounts_controller.rb
```diff
@@ -0,0 +1,55 @@
+require "faraday"
+
+module Publishers
+  class PaypalAccountsController < ApplicationController
```
corymcdonald: This should be name the same as the model `PaypalConnectionsController` to follow Rails conventions for mail/view/controllers.

app/controllers/publishers/paypal_accounts_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      # (Albert Wang): I don't think Paypal will go down that often, but we can change this into perform_later
+      # if this is a long running call.
+      Paypal::RefreshIdentity.new(publisher_id: current_publisher.id).perform
+      flash[:alert] = I18n.t('publishers.paypal_account_panel.still_not_verified') unless current_publisher.paypal_connection.verified_account?
+      redirect_to home_publishers_path
+    end
+
+    def connect_callback
+      authorization_code = params[:code]
+      result = Faraday.post("#{Rails.application.secrets[:paypal_api_uri]}/v1/oauth2/token") do |req|
```
corymcdonald: In the future I would like us to extract this into a bare-bones API class that gets created in an initializer so that we can do the following.

```ruby
access_token, refresh_token = PaypalClient.request_token(authorization_code: params[:code])
user_info = PaypalClient.user(access_token: access_token)

paypal_connection = PaypalConnection.find_or_initialize_by(
  user_id: current_publisher.id,
  email: user_info.email
)
paypal_connection.update(
  refresh_token: refresh_token,
  email: user_info.email,
  country: user_info.country,
  verified_account: user_info.verified_account == 'true',
  paypal_account_id: user_info.id,
  hidden: false
)
```

This is fine for now, but in the future let's keep this in mind.

app/helpers/publishers_helper.rb
```diff
@@ -154,10 +155,14 @@ def publisher_converted_last_settlement_balance(publisher)
     I18n.t("helpers.publisher.conversion_unavailable", code: settlement_currency)
   end
 
-  def publisher_last_settlement_date(publisher)
+  def publisher_last_settlement_date(publisher, locale)
     last_settlement_balance = publisher.wallet&.last_settlement_balance
     if last_settlement_balance&.timestamp.present?
-      Time.at(last_settlement_balance.timestamp).to_datetime.strftime("%B %d, %Y")
+      if locale.to_sym == :ja
+        I18n.l(Time.at(last_settlement_balance.timestamp).to_date, format: :long, locale: :'ja')
```
corymcdonald: To prevent the locale-specific if statement - can we simplify by passing in the locale into the `I18n.l`?

diracdeltas:
fyi this is blocked - see sec review issue for details

app/services/paypal/refresh_identity.rb
yachtcaptain23: No, the refresh token is used to obtain a new access token.

app/controllers/publishers/paypal_connections_controller.rb
```diff
@@ -0,0 +1,55 @@
+require "faraday"
```
corymcdonald: Do you need to require this? I was under the impression that autoloading took care of this.

app/services/paypal/refresh_identity.rb
```diff
@@ -0,0 +1,51 @@
...
+    paypal_connection.update(
+      country: user_info.dig('address', 'country'),
+      verified_account: user_info.dig('verified_account'),
+      paypal_account_id: user_info.dig('user_id')
+    )
+  end
+
+  private
+
+  def raise_token_exception
```
corymcdonald: Nit methods that raise exceptions should have exclamation marks 
```suggestion
  def raise_token_exception!
```

corymcdonald:
If the user has connected to PayPal they should not be able to go back to an english locale because at that point the account is in "BAT Points" mode.

PR #2379, opened 2019-11-13
Bugfix/disable uc jp

yachtcaptain23:


app/views/publishers/home.html.slim
```diff
@@ -102,7 +102,7 @@ script type="text/html" id="confirm_default_currency_modal_wrapper"
        .col-md.mb-4
          = render partial: 'payout_status', locals: { payout_report: @payout_report }
     .row
-      - unless @publisher.excluded_from_payout?
+      - unless @publisher.excluded_from_payout? || @publisher.uphold_connection.japanese_account? || @publisher.japanese_locale?(params[:locale])
```
corymcdonald:
```suggestion
      - unless @publisher.excluded_from_payout? || @publisher.uphold_connection&.japanese_account? || @publisher.japanese_locale?(params[:locale])
```

While `uphold_connection` shouldn't be nil it certainly can be, so let's do a check here.

PR #2383, opened 2019-11-13
BAT Points

yachtcaptain23:


corymcdonald:
We should remove the built with BAT
image

corymcdonald:
The statements page is not translated for BAT

mattmcalister:
good question. should the statements page be empty and/or inaccessible until we can do payouts and have payouts to send in BAP regions?

Issue #1719, opened 2019-03-15
Harden publishers#home for 503 errors from Ledger/Eyeshade


yachtcaptain23:
It seems we'll need to address anything that uses Faraday. Sample trace can be found https://sentry.io/organizations/brave-software/issues/909303464/?project=225116&query=is%3Aunresolved&statsPeriod=14d&utc=false

corymcdonald:
We've discussed 2 approaches 
- Completely make this synchronous with only one call to Eyeshade
- Detach the eyeshade wallet from blocking the loading of the homepage. AKA Everything asychronous

As a temporary solution the following options are being considered
- WalletGetter returns an empty hash `def wallet` while tracking if the first call failed
- One quick hack would be to have an instance variable set in the application controller for `@wallet`

tsmartt:
Closing as stale.

PR #2386, opened 2019-11-15
Cache wallet endpoint

corymcdonald:
#### Features

I want this caching to eventually go down from an hour to something more reasonable like 5-10 minutes. For now due to the instability of the servers we're going to set it for an hour.

- When Eyeshade is having problem throw an error instead of returning `:unavailable` 
- Cache service calls for an hour which make up the `Wallet` object.

### DIFF
--- a/app/helpers/publishers_helper.rb
+++ b/app/helpers/publishers_helper.rb
@@ -22,6 +22,15 @@ module PublishersHelper
     }
   end
 
+  def publisher_balance_last_update(publisher)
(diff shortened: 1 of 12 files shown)


yachtcaptain23:
Closes https://github.com/brave-intl/publishers/issues/1719

yachtcaptain23:
@corymcdonald will expand this PR to cover more transactions and store in a separate table

PR #2391, opened 2019-11-19
Additional translation from token on landing page

yachtcaptain23:


corymcdonald:
Still this in the footer
image

corymcdonald:
Ensure in the `_channels.html.slim` partial for the homepage that Uphold is not shown for Japanese creators
image

PR #2398, opened 2019-11-22
Sendgrid with marketing opt-in to allow Re-permissioning

yachtcaptain23:


app/models/publisher.rb
```diff
@@ -44,6 +44,7 @@ class Publisher < ApplicationRecord
 
   attr_encrypted :authentication_token, key: :encryption_key
 
+  attribute :visible, :boolean, default: false # (Albert Wang): We will use this as a flag for whether or not marketing emails are on for the user.
```
corymcdonald: We can't use the same database field that has already been set as true for all publishers. Let's create a new field named `subscribed_to_marketing_emails` and drop this field on publishers.

PR #2406, opened 2019-11-25
[Admin] Refactor Referral Reports

### LINKED ISSUE #2277: Timeouts when downloading Referral Reports
Business Development is encountering timeouts when downloading referral reports with geo-information included. This is because the server is taking longer than 30 seconds to generate the report.

There are a few different approaches to solving this

1. Refactor to reduce the complexity of [RegistrationStatsReportGenerator](https://github.com/brave-intl/publishers/blob/d61fb4175b827e40384f59b694284d8b210bc28a/app/services/promo/registration_stats_report_generator.rb#L22). Upon initial inspection is looks to be `O(n^3)`
1. Convert the stats generation to be an [enumerable and return a stream](https://coderwall.com/p/kad56a/streaming-large-data-responses-with-rails) to the browser to the request doesn't time out. [Heroku's 30-second timeout doesn't apply to streaming responses](https://devcenter.heroku.com/articles/request-timeout#long-polling-and-streaming-responses)
1. Return an alert that the report is being generated. Create a job which will generate and email the compiled report as an attachment. 
 

Upon further investigation even sending as many referral codes as we do actually leads to a 503 from Promo Server as well. We will also likely need to separate out the API calls into batches of 10 referral codes per API call.

### LINKED ISSUE #2395: Download by campaign

### LINKED ISSUE #2396: Add search to admin/unattached_promo_registrations by referral_code

corymcdonald:
## [Admin] Refactor Referral Reports

Closes #2277
Closes #2396
Closes #2395 

#### Features

- Sends emails of completed reports to user requesting
- Refactor and optimize the report generation to be easier to understand
- Add retry logic for retrieving geographic based information for referrals
- Improve searching / filtering on campaigns, including multi-select
- Unify creating campaign and code creation so we reduce orphaned referral codes 
- Add Select All / Select None checkbox for easily selecting which codes you want to filter on
- Add "Copy" button to allow admins to copy all the referral codes they've generated so they can share with partners.

Misc

- Move Pry to test and development mode so we can easily do `binding.pry` in test.

#### How To Test

**Main**
1. Download the referral report from the admin dashboard
2. Ensure that your localhost is connected to the promo server
3. Download the referral report locally
4. Ensure parity between 2 versions

**Copy Button**
Press the copy button, ensure all the selected referral codes are copied.

**Filtering**
Filter on multiple campaigns and ensure they show up in the result
Filter on Not assigned and ensure only unassigned referral codes appear
Filter on one campaign to ensure only one campaign appears in the results


app/jobs/generate_referral_report_job.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    InternalMailer.email_report(
+      email: email,
+      filename: filename,
+      subject: "#{interval.titleize} Referral Report  for #{referral_codes.length} codes between #{start_date} and #{end_date}",
+      body: "This report contains information for the following referral codes \n\n" \
+        "#{referral_codes.join("\n")} \n\n" \
+        "Reporting Interval: #{interval} \n" \
+        "Range: #{start_date} through #{end_date} "
+    ).deliver_now
```
yachtcaptain23: Would you delete the file after this runs?

app/jobs/generate_referral_report_job.rb
```diff
@@ -0,0 +1,28 @@
...
+    report_csv = Promo::RegistrationStatsReportGenerator.new(
+      referral_codes: referral_codes,
+      start_date: start_date,
+      end_date: end_date,
+      reporting_interval: interval,
+      is_geo: break_down_by_country
+    ).perform
+
+    filename = "tmp/#{Time.now}_referral_report_#{start_date}-#{end_date}.csv"
+    File.open(filename, "w") { |f| f.write(report_csv) }
```
yachtcaptain23: I don't really follow how this file gets stored. It gets stored on worker dyno which isn't shared with the web dynos. It would make sense to store the CSV in S3.

app/services/promo/registrations_geo_stats_fetcher.rb
```diff
@@ -1,29 +1,71 @@
...
+      request.options.params_encoder = Faraday::FlatParamsEncoder
+      request.headers["Authorization"] = api_authorization_header
+      request.headers["Content-Type"] = "application/json"
+      request.params = params.compact
+      request.url("/api/2/promo/geoStatsByReferralCode")
+    end
   rescue Faraday::ClientError => e
     Rails.logger.info("Error receiving referral stats #{e.message}")
-    []
+    nil
```
yachtcaptain23: Would it be useful to log this in Sentry?

app/jobs/generate_referral_report_job.rb
corymcdonald: I'll make this change. Fortunately there should never be a conflict because the file has a unique identifier and files and files are removed after the dyno restarts.

app/services/promo/registrations_geo_stats_fetcher.rb
corymcdonald: I don't think so, only because these errors are fairly consistent. The StandardError will be shown in Sentry if all the HTTP responses fail and then the engineer responsible can search the logs for the corresponding message.

PR #2407, opened 2019-11-25
Warn when settlement report can't be uploaded

corymcdonald:
#### Features

Last payout cycle when the final settlement report was uploaded the file wasn't able to be uploaded for some reason and so we need to warn when this happens

### DIFF
--- a/app/controllers/admin/payout_reports_controller.rb
+++ b/app/controllers/admin/payout_reports_controller.rb
@@ -78,8 +78,10 @@ class Admin::PayoutReportsController < AdminController
     redirect_to admin_payout_reports_path, flash: { notice: "Successfully uploaded settlement report" }
   rescue JSON::ParserError => e
     redirect_to admin_payout_reports_path, flash: { alert: "Could not parse JSON. #{e.message}" }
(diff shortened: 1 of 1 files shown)


app/controllers/admin/payout_reports_controller.rb
```diff
@@ -78,8 +78,10 @@ def upload_settlement_report
     redirect_to admin_payout_reports_path, flash: { notice: "Successfully uploaded settlement report" }
   rescue JSON::ParserError => e
     redirect_to admin_payout_reports_path, flash: { alert: "Could not parse JSON. #{e.message}" }
-  rescue Faraday::ClientError => eyeshade_error
+  rescue Faraday::ClientError
     redirect_to admin_payout_reports_path, flash: { alert: "Eyeshade responded with a 400 🤷‍️" }
+  rescue StandardError
+    redirect_to admin_payout_reports_path, flash: { alert: "Something bad happened!" }
```
yachtcaptain23:
```suggestion
    redirect_to admin_payout_reports_path, flash: { alert: "Something bad happened! Please check Sentry for more details" }
    require "sentry-raven"
    Raven.capture_exception(e)
```

PR #2408, opened 2019-11-25
Add index to promo registration order

yachtcaptain23:
Used. by https://github.com/brave-intl/publishers/pull/2406/files


yachtcaptain23:
> Not sure how necessary this is but doesn't hurt to add it.

It's due to https://github.com/brave-intl/publishers/pull/2406/files#diff-d11d53c09c2e1d4c10028e5ad03c848eR22

PR #2417, opened 2019-12-03
BAT Points ever on. Closes #2403

### LINKED ISSUE #2403: Blurb for showing payouts will be coming for Japanese Publishers

yachtcaptain23:


db/migrate/20191202190956_add_bat_points_ever_on_to_publishers.rb
```diff
@@ -0,0 +1,5 @@
+class AddBatPointsEverOnToPublishers < ActiveRecord::Migration[6.0]
+  def change
+    add_column :publishers, :bat_points_ever_on, :boolean, default: false
```
corymcdonald: I think I prefer the name `bat_points_enabled`, what do you think?

app/controllers/publishers_controller.rb
```diff
@@ -153,7 +153,10 @@ def home
       uphold_connection.create_uphold_cards if uphold_connection.missing_card?
     end
 
-    flash[:notice] = I18n.t("publishers.home.disabled_payouts") if current_publisher.japanese_locale?(params[:locale])
+    current_publisher.update(bat_points_ever_on: true) if current_publisher.japanese_locale?(params[:locale]) || current_publisher.uphold_connection&.japanese_account?
```
corymcdonald: I think we should be careful here. I don't think locale is enough to enforce BAT Points everywhere. I would think it should be geo based?

db/migrate/20191202190956_add_bat_points_ever_on_to_publishers.rb
yachtcaptain23: My impression is that `bat_points_enabled` is a user-defined setting. `ever_on` or `ever_enabled`  is a setting that can only go from false to true but never true to false

app/controllers/publishers_controller.rb
yachtcaptain23: I'm discussing with Matt

yachtcaptain23:
Spoke with Matt, closing.

PR #2421, opened 2019-12-03
Fix labels for payout in progress. Closes #2415

### LINKED ISSUE #2415: Payout Tracker Updates
The community and design have mentioned a few items that need to change on the payout tracker. 

- If `payout_warning` is present do not put the amount to be paid
The amount will be zero if there's a payout_warning, so Design suggest that should hide the amount.

- Rename “pending payout” to “balance
Design suggestion for clearing up confusion

- Add "Payout Report Generating" message in [`payout_warning`](https://github.com/brave-intl/publishers/blob/40f0ab2b4c5175bfa1da5daf0cd3c0cf2634f729/app/helpers/payout_helper.rb#L51) helper method so users see that the payout report is currently being built.

A lot of community members have reported this and are concerned they aren't being paid this cycle because the report hadn't generated any information for them. We should stop panic by letting users know that the report is still being built

yachtcaptain23:
Screen Shot 2019-12-03 at 2 07 25 PM
Screen Shot 2019-12-03 at 2 09 44 PM

### DIFF
--- a/app/jobs/set_payout_in_progress_job.rb
+++ b/app/jobs/set_payout_in_progress_job.rb
@@ -4,5 +4,6 @@ class SetPayoutInProgressJob < ApplicationJob
   def perform(set_to: true)
     Rails.cache.write('payout_in_progress', set_to)
     Rails.logger.info("Set payout in progress to #{set_to}")
+    Rails.cache.write('payout_report_generating', set_to)
   end
 end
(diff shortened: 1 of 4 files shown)


corymcdonald:
Will you update the `SetPayoutInProgressJob` to additionally write to the cache `Rails.cache.write('payout_report_generating', set_to)`. 

Then we can manually remove this when the report has completely finished

PR #2422, opened 2019-12-04
Sign in as the user. Closes #2419

### LINKED ISSUE #2419: Sign-in as user from Admin panel

yachtcaptain23:


app/controllers/admin/publishers_controller.rb
```diff
@@ -107,6 +107,16 @@ def refresh_uphold
     redirect_to admin_publisher_path(@publisher.id)
   end
 
+  def sign_in_as_user
```
corymcdonald: I don't like this, because it takes you out of the context of being an admin. Admins could login and then be under the user's account for up to 30 days. 

I would much rather have the home page use a singular instance variable for `@publisher`, and then be able to access the `publishers_controller` as an administrator and view the publishers account that way.

yachtcaptain23: Taking this offline, as I'll ask support what their thoughts are. The standard convention is found here: https://github.com/plataformatec/devise/wiki/How-To:-Sign-in-as-another-user-if-you-are-an-admin

app/javascript/views/admin/components/userNavbar/components/TopNav/TopNav.tsx
```diff
@@ -87,6 +96,24 @@ export default class Referrals extends React.Component<ITopNavProps, {}> {
...
     );
   }
+
+  private modalClick = async () => {
+    const url = "/admin/publishers/" + this.props.userID + "/sign_in_as_user";
+    const options = {
+      credentials: "same-origin",
+      headers: {
+        Accept: "text/html",
+        "X-CSRF-Token": document.head.querySelector("[name=csrf-token]")['content'] as string,
```
corymcdonald: .getAttribute("content")

yachtcaptain23:
Closed in favor of https://github.com/brave-intl/publishers/pull/2441

Issue #2387, opened 2019-11-15
Pending payouts is mislabeled on #home


yachtcaptain23:
If it's less than 5 BAT, it should not say `Pending Payout`. It should say payout floor which is 5 BAT.

PR #2433, opened 2019-12-09
Change pending label to 'Minimum balance for payout is X currency'. C…

yachtcaptain23:
…loses #2387

### DIFF
--- a/app/models/payout_report.rb
+++ b/app/models/payout_report.rb
@@ -4,6 +4,9 @@ class PayoutReport < ApplicationRecord
 
   LEGACY_PAYOUT_REPORT_TRANSITION_DATE = "2018-12-01 09:14:58 -0800".freeze
 
+  MINIMUM_BALANCE_AMOUNT = 5
+  BAT = 'bat'.freeze
+
   attr_encrypted :contents, key: :encryption_key, marshal: true
 
   has_many :potential_payments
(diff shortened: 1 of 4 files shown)


config/locales/ja.yml
```diff
@@ -554,6 +554,7 @@ ja:
       balance_pending: 保留中の支払い
       next_deposit_date: "次の入金日："
       payout_in_progress: "進行中の支払い"
+      payout_minimum_balance: "Minimum balance for payout is %{amount} %{currency}"
```
corymcdonald: Reminder to get the japanese locale before merge. :)

app/models/payout_report.rb
```diff
@@ -4,6 +4,9 @@ class PayoutReport < ApplicationRecord
 
   LEGACY_PAYOUT_REPORT_TRANSITION_DATE = "2018-12-01 09:14:58 -0800".freeze
 
+  MINIMUM_BALANCE_AMOUNT = 5
+  BAT = 'bat'.freeze
```
corymcdonald: Is this being used anywhere?

yachtcaptain23:
```suggestion
```

config/locales/ja.yml
yachtcaptain23: I'm aware :)

yachtcaptain23:
```suggestion
      payout_minimum_balance: "最低出金金額は %{amount} %{currency}です"
```

corymcdonald: Do you have the locales?

yachtcaptain23:
```suggestion
      payout_minimum_balance: "最低出金金額は %{amount} %{currency}です"
```

PR #2455, opened 2019-12-27
Continue making calls to eyeshade for wallet balance. Closes #2454.

### LINKED ISSUE #2454: Bug: 0 balance is there's no channel attached.
If a user removes all their channels, their reported balance gets set to 0. This ticket tracks a fix against https://github.com/brave-intl/publishers/blob/staging/app/services/publisher_balance_getter.rb#L10

yachtcaptain23:


corymcdonald:
Can you fix the tests?

PR #2465, opened 2020-01-10
Report generation for paypal

yachtcaptain23:


app/models/json_builders/paypal_payout_report_json_builder.rb
```diff
@@ -0,0 +1,42 @@
...
+            "type" => PotentialPayment::CONTRIBUTION,
+            "URL" => "#{Channel.find(potential_payment.channel_id).details.url}",
+            "address" => "#{potential_payment.derivied_paypal_account_id}",
+          })
+        end
+      end
+    end
+    contents
+  end
+end
```
nvonpentz: We probably can remove the payout report JSON builders since we aren't using them anymore.  Instead of downloading the JSON file from the admin dashboard, the antifraud system pulls the potential_payments table directly from the publishers database.

app/models/json_builders/paypal_payout_report_json_builder.rb
```diff
@@ -0,0 +1,42 @@
...
+        contents.push({
+          "name" => "#{potential_payment.name}",
+          "altcurrency" => "BAT",
+          "probi" => "#{potential_payment.amount}",
+          "fees" => "#{potential_payment.fees}",
+          "authority" => "",
+          "transactionId" => "#{potential_payment.payout_report_id}",
+          "owner" => "#{Publisher.find(potential_payment.publisher_id).owner_identifier}",
+          "type" => PotentialPayment::REFERRAL,
+          "address" => "#{potential_payment.derivied_paypal_account_id}",
```
nvonpentz: derived

yachtcaptain23:
> Could you link me to the migration files? I can't seem to find them in this PR.
> 
> I think it would make things easier on antifraud and settlement if we could add this bit of abstraction to the potential_payment table:
> 
> Instead of having two columns `uphold_id` and `paypal_account_id`, could we do something like `wallet_provider_id` and `wallet_provider`?

Yep I'll do so

yachtcaptain23:
> > I think it would make things easier on antifraud and settlement if we could add this bit of abstraction to the potential_payment table:
> > Instead of having two columns `uphold_id` and `paypal_account_id`, could we do something like `wallet_provider_id` and `wallet_provider`?
> 
> Yep I'll do so

Hm so looking back at it, it's a fairly risky refactor as there will be 20 lines to rename for just `uphold_id`. If you're adament, it's do-able, though the table states `potential_paypal_payments`, so your script will have to make that read.

yachtcaptain23:
Merged in https://github.com/brave-intl/publishers/pull/2494

PR #2471, opened 2020-01-14
Create Payment Message for reporting

corymcdonald:
#### Features

It's currently very hard to debug why a publisher might not have been included in a payout. We introduce a new table to help diagnose and debug reasons why a publisher might not have been included. 

- Introduce PayoutMessage
- Introduce a field for `include_transactions` on PublisherWalletGetter. Transactions are not needed during payout report.

### DIFF
--- a/app/controllers/admin/publishers_controller.rb
+++ b/app/controllers/admin/publishers_controller.rb
@@ -50,6 +50,12 @@ class Admin::PublishersController < AdminController
     @potential_referral_payment = @publisher.most_recent_potential_referral_payment
     @referral_owner_status = PromoClient.owner_state.find(id: params[:id])
     @current_user = current_user
+
+    if true || payout_in_progress? || Date.today.day < 12 # Let's display the payout for 5 days after it should complete (on the 8th)
+      @payout_report = PayoutReport.where(final: true, manual: false).order(created_at: :desc).first
+      @payout_message = PayoutMessage.find_by(payout_report: @payout_report, publisher: @publisher)
+      @payout_message = PayoutMessage.first
+    end
   end
 
   def edit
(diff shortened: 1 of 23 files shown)


app/controllers/admin/publishers_controller.rb
```diff
@@ -50,6 +50,12 @@ def show
     @potential_referral_payment = @publisher.most_recent_potential_referral_payment
     @referral_owner_status = PromoClient.owner_state.find(id: params[:id])
     @current_user = current_user
+
+    if payout_in_progress? || Date.today.day < 12 # Let's display the payout for 5 days after it should complete (on the 8th)
+      @payout_report = PayoutReport.where(final: true, manual: false).order(created_at: :desc).first
+      @payout_message = PayoutMessage.find_by(payout_report: @payout_report, publisher: @publisher)
```
yachtcaptain23: Duplicate call?

yachtcaptain23:
```suggestion
```

app/services/publisher_wallet_getter.rb
```diff
@@ -30,6 +33,18 @@ def perform
 
   private
 
+  def rates
+    # Cache the ratios every minute. Rates are used for display purposes only.
+    Rails.cache.fetch("RATES_CACHE_KEY", expires_in: 1.minute) do
```
yachtcaptain23:
```suggestion
    Rails.cache.fetch(RATES_CACHE_KEY, expires_in: 1.minute) do
```

app/controllers/admin/publishers_controller.rb
corymcdonald: One if PayoutReport the other is PayoutMessage 😄

app/controllers/admin/publishers_controller.rb
```diff
@@ -50,6 +50,12 @@ def show
...
+      @payout_message = PayoutMessage.first
```
corymcdonald:
```suggestion

```

PR #2501, opened 2020-01-26
Perf/reader read

yachtcaptain23:


yachtcaptain23:
Split to #2502 and #2503

PR #2502, opened 2020-01-27
Use one transaction for save

yachtcaptain23:


app/services/promo/registrations_stats_fetcher.rb
```diff
@@ -4,42 +4,63 @@ class Promo::RegistrationsStatsFetcher < BaseApiClient
...
         request.headers["Authorization"] = api_authorization_header
         request.headers["Content-Type"] = "application/json"
         request.url("/api/2/promo/statsByReferralCode#{query_string}")
       end
       referral_code_events_by_date = JSON.parse(response.body)
-      referral_code_batch.each do |referral_code|
-        promo_registration = PromoRegistration.find_by_referral_code(referral_code)
-        next if promo_registration.nil?
+      values = []
+      promo_registrations_batch.each do |promo_registration|
```
dlipeles: Is it okay to run a sql query for each of these promo registrations?

app/services/promo/registrations_stats_fetcher.rb
```diff
@@ -4,42 +4,63 @@ class Promo::RegistrationsStatsFetcher < BaseApiClient
...
+    @promo_registrations = promo_registrations
     @update_only = update_only
   end
 
   def perform
     return perform_offline if Rails.application.secrets[:api_promo_base_uri].blank?
     stats = []
-    @referral_codes.each_slice(BATCH_SIZE).to_a.each do |referral_code_batch|
-      query_string = query_string(referral_code_batch)
+    @promo_registrations.in_batches(of: BATCH_SIZE) do |promo_registrations_batch|
```
dlipeles: nice use of batching here IMO , looks good.

app/services/promo/registrations_stats_fetcher.rb
yachtcaptain23: It's already loaded in a previous query, so we're good :)

PR #2505, opened 2020-01-28
Sync only at 3 AM

yachtcaptain23:


yachtcaptain23:
No longer needed based on performance mentioned by Newrelic

PR #2508, opened 2020-01-28
Use redis-objects and ConnectionPool

yachtcaptain23:
Closes https://github.com/brave-intl/publishers/issues/2328


yachtcaptain23:
Alternate solution found. No longer needed.

PR #2516, opened 2020-01-30
Release 2020-01-31

yachtcaptain23:


app/jobs/cache_browser_channels_json_job.rb
```diff
@@ -2,8 +2,11 @@ class CacheBrowserChannelsJsonJob < ApplicationJob
   queue_as :heavy
 
   MAX_RETRY = 10
+  LAST_WRITTEN_AT_KEY = "CacheBrowserChannelsJsonJob_last_written_at".freeze
 
   def perform
+    last_written_at = Rails.cache.fetch(LAST_WRITTEN_AT_KEY)
+    return if last_written_at.present? && last_written_at > 2.hours.ago
```
corymcdonald: Reading this logic wouldn't we want to only return if the `last_written_at` was less than 2 hours ago?

app/jobs/cache_browser_channels_json_job_v2.rb
```diff
@@ -17,6 +20,7 @@ def perform
     end
 
     if result
+      Rails.cache.write(LAST_WRITTEN_AT_KEY, Time.now)
```
dlipeles: Will this key / value get overwritten by the other jobs calling it? Maybe a unique key for each job?

app/jobs/cache_browser_channels_json_job.rb
yachtcaptain23: Unless I'm mistaken, that's what it says?

app/jobs/cache_browser_channels_json_job_v2.rb
yachtcaptain23: Single worker, so it's fine

app/jobs/cache_browser_channels_json_job.rb
corymcdonald: `last_written_at` is greater than 2 hours ago, in which case we should not return?

```
return if last_written_at.present? && last_written_at < 2.hours.ago
```

See the below for example
```
irb(main):009:0> Time.now
=> 2020-01-31 12:10:08 -0600
irb(main):010:0> last_written_at = 3.hours.ago
=> 2020-01-31 09:10:17 -0600
irb(main):011:0> last_written_at > 2.hours.ago
=> false
irb(main):012:0> last_written_at < 2.hours.ago
=> true
```

yachtcaptain23: That's correct, we only run the job if the cache is no longer fresh

PR #2543, opened 2020-02-13
Make search a little better

corymcdonald:
#### Features

For far too long has search been slow when searching for emails. This is an improvement to search email if the user is searching for an email, by referral_codes if the user is searching for a referral code, or fallback to a generic search if they aren't

### DIFF
--- a/app/controllers/concerns/search.rb
+++ b/app/controllers/concerns/search.rb
@@ -18,6 +18,23 @@ module Search
     query.strip
   end
 
+  # Returns an ActiveRecord::Relation of publishers for pagination
+  def publishers_search(publishers, query)
+    search_query = remove_prefix_if_necessary(params[:q])
+
+    # Simple optimization to only search for the things that the admins search for the most
+    if is_email?(search_query)
+      results = publishers.where('email ILIKE ?', "%#{search_query}%")
+    elsif is_promo_code?(search_query)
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/search.rb
```diff
@@ -18,6 +18,23 @@ def remove_prefix_if_necessary(query)
...
+  def publishers_search(publishers, query)
+    search_query = remove_prefix_if_necessary(params[:q])
+
+    # Simple optimization to only search for the things that the admins search for the most
+    if is_email?(search_query)
+      results = publishers.where('email ILIKE ?', "%#{search_query}%")
+    elsif is_promo_code?(search_query)
+      results = publishers.left_joins(:channels).joins(channels: :promo_registration).where(promo_registrations: { referral_code: search_query.upcase })
+    else
+      search_query = "%#{search_query}%" unless is_a_uuid?(search_query)
```
yachtcaptain23: Would you move `is_a_uuid?` from `app/controllers/admin/publishers_controller.rb` to this file? Otherwise this looks great

yachtcaptain23:
Looks good to me. There's one caveat but I think we're fine

```
select *
from promo_registrations
where publisher_id is not null
and channel_id is null;
```
yields 91 results, but I think they're old enough that it's fine to ignore this group.

PR #2559, opened 2020-02-24
Feature flags and special payments

### LINKED ISSUE #2540: Create feature flag system
We need to create a feature flag system to enable certain features for certain publishers

An API similar to this would be implemented 

```ruby
current_user.feature_name_enabled?
# true / false

current_user.features 
# [{ name: feature_name1, enabled: false }, { name: feature_name2, enabled: false }]
```

This will allow us to enable certain features for a specific subset of users and will allow us to trial features.

yachtcaptain23:
Closes #2540


app/models/publisher.rb
```diff
@@ -15,6 +15,10 @@ class Publisher < ApplicationRecord
 
   OWNER_PREFIX = "publishers#uuid:".freeze
 
+  # **************************************************
+  # Feature flags:
+  WIRE_ONLY = "wire_only".freeze
```
corymcdonald: Can we separate this constant and relevant function out into a concern?

db/schema.rb
```diff
@@ -340,6 +341,29 @@
     t.index ["publisher_id"], name: "index_potential_payments_on_publisher_id"
   end
 
+  create_table "potential_paypal_payments", id: :uuid, default: -> { "uuid_generate_v4()" }, force: :cascade do |t|
```
corymcdonald: I don't see a migration for this. Can you reset db/schema.rb and re-run your migrations?

corymcdonald:
Would you be able to make a simple interface in the `admin/publishers/edit` page to toggle on the wire feature flag?

app/views/admin/publishers/edit.html.slim
```diff
@@ -14,6 +14,10 @@
         label
           = f.check_box :excluded_from_payout, {}, checked: @publisher.excluded_from_payout?
           = " Uphold widget disabled"
+          br
```
corymcdonald: When you click the text it checks the wrong box, so if we change this to be like this it should work.
```suggestion
          br
        label
```

PR #2569, opened 2020-03-02
Feature/login with uphold

yachtcaptain23:


yachtcaptain23:
Please add index on `card_id`

app/controllers/uphold_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+      redirect_to root_path
+    end
+  end
+
+  private
+
+  def signup_user_if_necessary_or_signin(uphold_model_card)
+    uphold_connection = UpholdConnection.find_by(card_id: uphold_model_card.id)
+    if uphold_connection.nil?
+      user = BrowserUserSignUpService.new(uphold_card_id: uphold_card_id).perform
```
corymcdonald:
```suggestion
      user = BrowserUserSignUpService.new(uphold_card_id: uphold_model_card.id).perform
```

app/services/browser_user_sign_up_service.rb
```diff
@@ -0,0 +1,13 @@
+class BrowserUserSignUpService < BaseService
+  def initialize
```
corymcdonald: Getting an error when I call this
```
wrong number of arguments (given 1, expected 0)
```

app/services/promo/peer_to_peer_registration.rb
```diff
@@ -0,0 +1,49 @@
+class Promo::PeerToPeerRegistration < BaseApiClient
```
corymcdonald: I think this might look better in something that's similar to the new API client structure we have

Like this
https://github.com/brave-intl/publishers/blob/07702bfbe7df53448496b1f0f5bd0851ceac75da/app/services/promo/models/owner_state.rb#L37-L43

Or 

https://github.com/brave-intl/publishers/blob/staging/app/services/promo/models/reporting.rb#L9-L28

app/controllers/uphold_controller.rb
```diff
@@ -0,0 +1,57 @@
+class UpholdController < ApplicationController
```
corymcdonald: I would prefer to not have two controllers named the same, can we rename one of these `UpholdConnectionsController`?

db/schema.rb
```diff
@@ -573,6 +574,8 @@
     t.datetime "default_currency_confirmed_at"
     t.datetime "member_at"
     t.datetime "send_emails", default: -> { "CURRENT_TIMESTAMP" }
+    t.text "card_id"
+    t.index ["card_id"], name: "index_uphold_connections_on_card_id"
```
corymcdonald: I don't necessarily think we need create a `card_id` field because it's equivalent to the existing `address` field.

yachtcaptain23: `address` isn't the same as `card_id` from what I can tell

app/controllers/uphold_connections_controller.rb
```diff
@@ -0,0 +1,59 @@
+class UpholdConnectionsController < ApplicationController
+  def login
+    # TODO: Make sure these params are safe
```
diracdeltas: what do you mean by safe?

app/views/layouts/application.html.slim
```diff
@@ -54,7 +54,7 @@ html
 
 
     #main
-      - if !request.original_url.include? "/partners/"
+      - if request.path.starts_with?("/publisher")
```
diracdeltas: this would match something like `/publisher_foobar`, which i assume you don't want.

yachtcaptain23: This is just UI that we only want to show for publishers exclusively. I can't foresee any risk.

yachtcaptain23:
Closing in favor of #2598

app/controllers/uphold_connections_controller.rb
yachtcaptain23: Just want to make sure the values won't consume Redis. Check is done in https://github.com/brave-intl/publishers/pull/2598/files#diff-5f2a026ba7311cca218cea5b67ee9cd2R5

PR #2580, opened 2020-03-13
Set a Content-Security-Policy

### LINKED ISSUE #1565: security: site should set security headers
according to https://www.hardenize.com/report/publishers.basicattentiontoken.org/1548890217#www_csp, publishers does not set CSP headers. see https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP for more details.

yachtcaptain23:
Closes #1565


corymcdonald:
Can you deploy this on staging so we can test?

PR #2586, opened 2020-03-18
Handle future default donation amount changes

yachtcaptain23:


db/migrate/20200317211000_drop_nullable_requirement_from_site_banners.rb
```diff
@@ -0,0 +1,6 @@
+class DropNullableRequirementFromSiteBanners < ActiveRecord::Migration[6.0]
+  def change
+    change_column_null :site_banners, :donation_amounts, true
```
corymcdonald: Have you checked the Site Banner code to see what happens when someone creates a new account / channel with a nil site banner? I think I remember that any new site banner creates with a tip set

yachtcaptain23: Yeah it works just fine.

yachtcaptain23:
I don't think we should let https://github.com/wycats/handlebars.js/issues/1658 block this

PR #2588, opened 2020-03-18
Limit annoying emails

yachtcaptain23:


config/initializers/rack-attack.rb
```diff
@@ -107,12 +107,8 @@ class Rack::Attack
...
   # end
 
-  # In PublishersController we'll check the annotated request object
-  # to apply additional Recaptcha.
-  throttle("registrations/ip", limit: 60, period: 1.hour) do |req|
-    # In JS we specify that the format is application/json by writing /publishers/registrations.json
-    path = req.path.sub('.json', '')
-    if (path == "/publishers" || path == "/publishers/registrations") && (req.post? || req.patch?)
+  throttle("registrations/create", limit: 10, period: 1.hour) do |req|
+    if (req.path.starts_with?("/publishers/registrations") && (req.post? || req.patch? || req.put?))
```
corymcdonald: The `req.path.starts_with?` won't take into account the old routes that we have if users have javascript disabled. 

https://publishers.basicattentiontoken.org/publishers/sign_up
https://publishers.basicattentiontoken.org/publishers/log_in

yachtcaptain23: That should already be covered by:
```

  # Throttle POST requests to /login by IP address
  #
  # Key: "rack::attack:#{Time.now.to_i/:period}:logins/ip:#{req.ip}"
  throttle("logins/ip", limit: 5, period: 120.seconds) do |req|
    if req.path.start_with?("/publishers/") && req.params["token"]
      req.ip
    end
  end
```

corymcdonald: That's only attempting to log into the application, not for sending emails to our end users

diracdeltas:
any update on whether we're doing anything about this?

PR #2589, opened 2020-03-20
Protect against thundering herd problem by only allowing 1 request at…

yachtcaptain23:
… a time

### DIFF
--- a/app/controllers/concerns/browser_channels_dyno_caching.rb
+++ b/app/controllers/concerns/browser_channels_dyno_caching.rb
@@ -3,17 +3,29 @@ module BrowserChannelsDynoCaching
   require 'sentry-raven'
 
   def channels
-    value = "#{Time.now.to_i}"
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/browser_channels_dyno_caching.rb
```diff
@@ -3,14 +3,30 @@ module BrowserChannelsDynoCaching
...
       update_dyno_cache
     end
     render(json: self.class.class_variable_get(klass_dyno_cache), status: 200)
+    Rails.cache.delete(self.class::REDIS_THUNDERING_HERD_KEY)
   end
 
   private
 
+  def clear_if_old_lock
+    past_time = Time.at(Rails.cache.fetch(self.class::REDIS_THUNDERING_HERD_KEY))
```
corymcdonald: Can this be nil? Can you add a check?

yachtcaptain23:
Tested on dev by setting the lock.

PR #2603, opened 2020-04-06
Update count of content creator accounts to 500,000

AlanBreck:
Screen Shot 2020-04-06 at 11 43 58 AM

#### Features

Updates count of verified content creators to 500,000 as per https://github.com/brave/brave-com-resources/issues/96

Changes text to:
> Join over 500,000 content creators and publishers earning more from Brave Rewards

_**NOTE**_: not modified in Japanese.

### DIFF
--- a/public/creators-landing/src/locale/en.js
+++ b/public/creators-landing/src/locale/en.js
@@ -62,7 +62,7 @@ export default {
   signoff: {
     btn: "Sign up now",
     btnHref: "/sign-up",
-    headlineOne: "Join ",
+    headlineOne: "Join over ",
     headlineTwo:
       " content creators and publishers earning more from Brave Rewards"
   },
--- a/public/creators-landing/src/locale/en.js
+++ b/public/creators-landing/src/locale/en.js
@@ -62,9 +62,7 @@ export default {
   signoff: {
     btn: "Sign up now",
     btnHref: "/sign-up",
-    headlineOne: "Join over ",
-    headlineTwo:
-      " content creators and publishers earning more from Brave Rewards"
+    headline: "Join over {count} content creators and publishers earning more from Brave Rewards"
   },
   sign: {
     sentAgain:
--- a/public/creators-landing/src/locale/ja.js
+++ b/public/creators-landing/src/locale/ja.js
@@ -62,7 +62,7 @@ export default {
   signoff: {
     btn: "新規登録",
     btnHref: "/sign-up",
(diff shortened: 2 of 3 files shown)


AlanBreck:
The automated translations I attempted do not fit with the current localization strategy of concatenating two portions of the string to either side of the number (as seen below).

*Automated translation*
> 50万人以上のコンテンツクリエイターや出版社に参加して、Brave Rewardsからより多くの収入を得ましょう

*Current localization strategy*
```js
{<FormattedMessage id="signoff.headlineOne"/>}
<strong>500,000</strong>
{<FormattedMessage id="signoff.headlineTwo"/>}
```

public/creators-landing/src/locale/en.js
```diff
@@ -62,7 +62,7 @@ export default {
   signoff: {
     btn: "Sign up now",
     btnHref: "/sign-up",
-    headlineOne: "Join ",
+    headlineOne: "Join over ",
```
corymcdonald, line 65: Can we remove headlineTwo and change this to be 

```yaml
headline: "Join over {count} content creators and publishers earning more from Brave Rewards",
```

And then in the code located in the `signoff/index.js` page we can have something similar to what the [example in the react-intl](https://github.com/formatjs/react-intl#example) contains

```jsx
<Heading alignSelf='center' level='4' textAlign='center'>
  <FormattedMessage id="signoff.headline" values={{count: <b>500,000</b>}}/>
</Heading>
```

AlanBreck: One possible hiccup is that the translated text (at least from AWS) doesn't even have the count formatted as we would in English:

> 50万人以上のコンテンツクリエイターや出版社に参加して、Brave Rewardsからより多くの収入を得ましょう

Wondering if there's value in not even interpolating the count at all.

corymcdonald: Possibly! I would like to introduce a PR that use calculates the number based on [totals endpoint](https://publishers-distro.basicattentiontoken.org/api/v3/public/channels/totals) so we no longer have to manually update it. 

I'm going to start a chat on Slack with you and Ryotaro and see what he thinks. 😄

AlanBreck: I've updated as suggested (with a few slight changes).

Below is what it looks like in Japanese. Confirming with Ryotoro in Slack.

Screen Shot 2020-04-09 at 3 41 21 PM

PR #2618, opened 2020-04-10
Better lookup of invoices

corymcdonald:
## Better lookup of invoices

The amount provide in the JSON is in the currency that will be deposited into the publishers uphold account. To counteract this we should lookup by the `documentId` provided or the `probi`  amount

### DIFF
--- a/app/controllers/admin/payout_reports_controller.rb
+++ b/app/controllers/admin/payout_reports_controller.rb
@@ -56,18 +56,26 @@ class Admin::PayoutReportsController < AdminController
 
     EyeshadeClient.publishers.create_settlement(body: json)
 
+    not_found = []
+
     json.each do |entry|
       next unless entry["type"] == MANUAL && entry["owner"] && entry["amount"]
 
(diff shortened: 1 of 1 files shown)


app/controllers/admin/payout_reports_controller.rb
```diff
@@ -56,18 +56,26 @@ def upload_settlement_report
...
-      ).order(:created_at).first
-
-      next unless invoice.present?
+      begin
+        invoice = Invoice.find(entry.dig('documentId'))
+      rescue ActiveRecord::RecordNotFound
+        publisher_id = entry["owner"].sub(Publisher::OWNER_PREFIX, "")
+        invoice = Invoice.where(
+          publisher_id: publisher_id,
+          finalized_amount: (entry["probi"].to_i / 1E18).round(2),
```
yachtcaptain23, line 70: Would it be reasonable to eliminate this part of the `where` clause? It looks like it might return 0 records when in fact there is a record.

corymcdonald: If we take it out and there's multiple invoices which are being settled as part of the process this might mark the wrong one as paid

PR #2622, opened 2020-04-13
Create Merchant Feature

corymcdonald:
#### Features

* Adds Merchant Feature Flag
* Add Ability to create keys

#### How To Test

1. Add the merchant feature flag through admin dashboard
1. Click on the top bar navigation that says "Merchant keys"
1. Create / roll / delete keys


app/javascript/packs/application.js
```diff
@@ -9,6 +9,7 @@ import "publishers/home";
 import "site_channels/siteChannels";
 import Rails from "@rails/ujs";
 
+console.log('I cant believe this isntl oading')
```
yachtcaptain23, line 12: Lol

app/services/payment/models/key.rb
```diff
@@ -0,0 +1,41 @@
...
+module Payment
+  module Models
+    class Key < Client
+      include Initializable
+
+      attr_accessor :id, :name, :merchant, :secret_key, :created_at, :expiry
+
+      # For more information about how these URI templates are structured read the explaination in the RFC
+      # https://www.rfc-editor.org/rfc/rfc6570.txt
+      PATH = Addressable::Template.new("v1/merchants/{publisher_id}/keys{/id}")
```
yachtcaptain23, line 13: Extra slash in the parameter?

yachtcaptain23:
Would you provide or point to documentation to launch the Merchant side for testing (and my own curiosity lol)

app/views/publishers/keys/new.html.slim
```diff
@@ -0,0 +1,6 @@
+.dashboard
+  .row
+    .col.mb-4
+      .rounded.bg-white.p-4
+        .d-flex.align-items-center.justify-content-between.mb-3
+          = render "new"
```
yachtcaptain23, line 6: This is slightly confusing. Is it possible to change this file or this rendering?

yachtcaptain23: Nevermind.

app/javascript/modal.js
```diff
@@ -205,4 +202,14 @@ document.addEventListener("DOMContentLoaded", function() {
...
     );
   }
-});
+}
+
+
+/*
+ * Setup the DOM event listeners on links requesting confirmation.
+ */
+document.addEventListener("DOMContentLoaded", detectModals);
```
yachtcaptain23, line 211: How does this work?

corymcdonald: Before we had a function that was being called when the `DOMContentLoaded` event occurred. I extracted that function and gave it name `detectModels` and added it on the window so we can call it again in the `create.js.erb`

app/services/payment/models/key.rb
corymcdonald: Don't think so, Faraday should trim extra slashes 😄

app/javascript/packs/application.js
corymcdonald: https://github.com/brave-intl/publishers/pull/2622/commits/07c7fe562bf184c456e40ab36f32a76d3cff062d

PR #2625, opened 2020-04-13
Eliminate erroneous devise flash messages. Closes #2620

### LINKED ISSUE #2620: Fix `session has expired` on a new session

yachtcaptain23:

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -188,7 +188,9 @@ class PublishersController < ApplicationController
   private
 
   def authenticate_via_token
-    # For some odd reason, devise flash alerts get displayed during auth
+    # For some odd reason, devise flash alerts get displayed during auth.
+    # Deeper details can be chased in:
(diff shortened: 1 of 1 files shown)


app/controllers/publishers_controller.rb
```diff
@@ -188,6 +188,8 @@ def get_site_banner_data
   private
 
   def authenticate_via_token
+    # For some odd reason, devise flash alerts get displayed during auth
```
corymcdonald, line 191: Can we link to the CustomFailureApp configured on Devise that sets this flash? https://github.com/heartcombo/devise/blob/83a32e6d2118b0535cf54b48df4f9853d85b55fd/lib/devise/failure_app.rb

yachtcaptain23: I got rid of the CustomFailureApp since it was causing significant problems related to sign-in.

corymcdonald: I think that the FailureApp in Devise is called by default, and our CustomFailureApp overrode that functionality.

https://github.com/heartcombo/devise/blob/715192a7709a4c02127afb067e66230061b82cf2/lib/devise/mapping.rb#L123

PR #2626, opened 2020-04-14
Disallow accounts that try signin with the same uphold id

### LINKED ISSUE #2610: Restrict users from signing-in with an existing uphold connection

yachtcaptain23:
Closes #2610

### DIFF
--- a/app/controllers/publishers/uphold_controller.rb
+++ b/app/controllers/publishers/uphold_controller.rb
@@ -77,7 +77,21 @@ module Publishers
       ExchangeUpholdCodeForAccessTokenJob.perform_now(uphold_connection_id: uphold_connection.id)
 
       uphold_connection.reload
-      create_uphold_report!(uphold_connection)
+      # TODO: Should we drop this call from publishers_controller?
+      uphold_connection.sync_from_uphold!
+      if uphold_connection.has_duplicate_publisher_account?
+        create_uphold_report!(connection: uphold_connection, duplicate: true)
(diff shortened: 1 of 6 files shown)


app/controllers/publishers/uphold_controller.rb
```diff
@@ -77,7 +77,21 @@ def create
...
+      uphold_connection.sync_from_uphold!
+      if uphold_connection.has_duplicate_publisher_account?
+        create_uphold_report!(connection: uphold_connection, duplicate: true)
+        admin = Publisher.find_by(email: Rails.application.secrets[:zendesk_admin_email])
+        PublisherNote.create(
+          publisher_id: uphold_connection.publisher_id,
+          note: "The user has a duplicate uphold address at uphold address: #{uphold_connection.address_id}",
+          created_by_id: admin.id
+        )
+        uphold_connection.disconnect_uphold
```
corymcdonald, line 90: This is a pretty big change for our users. Especially since Brave has previously told them that it's okay to have multiple uphold accounts connected. I think we should add an easy path to migrate to one account, I'm not too sure how important it is to add to UpholdStatusReport with a `duplicate`.

yachtcaptain23: This will only apply to new users.

corymcdonald: What happens with users who use the peer to peer referral system? Will they be able to have a creator account?

yachtcaptain23:
From a business perspective, this code isn't useful anymore.

PR #2627, opened 2020-04-14
Case System improvements

### LINKED ISSUE #2068: When you close or accept a case it takes you to the next one
image

### LINKED ISSUE #2302: Add button to reject an already resolved Case
We should add a button to move a case back into reject once it's been accepted in case a user accidentally clicks the wrong button.

### LINKED ISSUE #2303: Next in queue for cases
Asad, Steven and Mattches would like the ability to go to the next case they have assigned automatically when they accept or reject a case.

image

corymcdonald:
## Case System improvements

Closes #2303
Closes #2302
Closes #2068

#### Features

* Rename some general translation strings
* Increase the pagination limit from 20 to 50
* Introduce ability to `OR` together search queries. Ex: Searching for `status:open status:in_progress` shows cases which are open or in progress
* Add ability to re-open case

### DIFF
--- a/app/controllers/admin/cases_controller.rb
+++ b/app/controllers/admin/cases_controller.rb
@@ -32,18 +32,7 @@ module Admin
         search_case = Case.where("case_number = ?", case_id)
       elsif query.include?(":")
         # Search for each type on the db model, like assignee = ""
-        query.split(' ').each do |term|
-          type, value = term.split(':')
-          case type
-          when 'assigned'
-            @assigned = Publisher.where('email like ? AND role = ?', "%#{value}%", 'admin').first
-            search_case = search_case.joins(:assignee).where('publishers.email LIKE ?', "%#{value}%")
-          when 'status'
-            # Remove all non-alphanumeric values from the status field
-            value = value.gsub(/[^0-9a-z_]/i, '')
-            search_case = search_case.where('status LIKE ?', "%#{value}%") if value.present?
-          end
-        end
+        search_case = parse_search(search_case, query.split(' '))
       else
         search_query = "%#{query}%"
         search_case = Case.joins(:publisher).where('publishers.name LIKE ?', search_query).or(
@@ -88,7 +77,8 @@ module Admin
     def update
       @case = Case.find(params[:id])
 
-      if @case.update(status: params[:status]) && params[:status] == Case::RESOLVED
+      case_updated = @case.update(status: params[:status])
+      if case_updated && params[:status] == Case::RESOLVED
         note = PublisherNote.create(
           publisher: @case.publisher,
           created_by: current_user,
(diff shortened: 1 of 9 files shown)


app/controllers/admin/cases_controller.rb
```diff
@@ -97,7 +87,43 @@ def update
...
+    private
+
+    def parse_search(search_case, queries)
+      statuses = queries.select { |x| x.include?("status") }.map { |x| x.split(':').last&.gsub(/[^0-9a-z_]/i, '') }
+      assigned = queries.select { |x| x.include?("assigned") }.map { |x| x.split(':').last&.gsub(/[^0-9a-z_]/i, '') }
+
+      first_status = statuses[0]
+      if first_status.present?
+        value = first_status
+        search_case = search_case.where('status LIKE ?', "%#{value}%")
```
yachtcaptain23, line 109: Instead of `status LIKE %value%`, would it be possible to use `where(status: value)` or `status LIKE value%`? Otherwise it'll be an O(n) search operation. 

From https://www.postgresql.org/docs/9.3/indexes-types.html
```
The optimizer can also use a B-tree index for queries involving the pattern matching operators LIKE and ~ if the pattern is a constant and is anchored to the beginning of the string — for example, col LIKE 'foo%' or col ~ '^foo', but not col LIKE '%bar'. 
```

corymcdonald: Yep! https://github.com/brave-intl/publishers/pull/2627/commits/0bb03e74fb8d1cbfbb6fb6ac5e649d264defde29

app/views/admin/shared/_sidebar.html.slim
```diff
@@ -27,8 +27,6 @@ aside
           = nav_link "Approvals", admin_channel_approvals_path
           = nav_link "Duplicates", duplicates_admin_channels_path
 
-      = nav_link "Organizations", admin_organizations_path
```
yachtcaptain23, line 30: Can we retain this? It's likely in near future we'll want to expand Organization for all publishers.

app/views/admin/cases/show.html.slim
```diff
@@ -97,8 +97,10 @@
       - if @case.in_progress?
         .text-dark.justify-content-between.mb-2.align-items-center
           strong Actions
-        = link_to "Resolve", admin_case_path(@case, status: Case::RESOLVED), method: :patch, class: 'btn btn-success mx-1 mb-2'
-        = link_to "Close", admin_case_path(@case, status: Case::CLOSED), method: :patch, class: 'btn btn-dark mx-1 mb-2'
+        = link_to "Approve", admin_case_path(@case, status: Case::RESOLVED), method: :patch, class: 'btn btn-success mx-1 mb-2'
```
yachtcaptain23, line 100: I originally named these `RESOLVED` and `CLOSED` to copy JIRA, in which the definitions are better defined here: https://community.atlassian.com/t5/Jira-questions/What-is-the-difference-between-closed-and-resolved-status-in/qaq-p/378505

I'm open to re-labeling, but I think @atsyed87 and his team should be aware.

app/controllers/admin/cases_controller.rb
```diff
@@ -97,7 +87,43 @@ def update
...
+
+        statuses[1..-1].each do |value|
+          search_case = search_case.or(Case.where('status LIKE ?', "%#{value}%"))
+        end
+      end
+
+      first_assigned = assigned[0]
+      if first_assigned.present?
+        value = first_assigned
+        search_case = search_case.joins(:assignee).where('publishers.email LIKE ?', "%#{value}%")
```
yachtcaptain23, line 119: Would it be okay to use `publishers.email LIKE ?, "#{value}%` instead to make use of the index? Otherwise I'd prefer we'd read the read replica.

Same request applies to line 122

app/controllers/admin/cases_controller.rb
```diff
@@ -97,7 +87,43 @@ def update
...
+    def parse_search(search_case, queries)
+      statuses = queries.select { |x| x.include?("status") }.map { |x| x.split(':').last&.gsub(/[^0-9a-z_]/i, '') }
+      assigned = queries.select { |x| x.include?("assigned") }.map { |x| x.split(':').last&.gsub(/[^0-9a-z_]/i, '') }
+
+      first_status = statuses[0]
+      if first_status.present?
+        value = first_status
+        search_case = search_case.where(status: value)
+
+        statuses[1..-1].each do |value|
```
yachtcaptain23, line 111: Block 111 to 114 can be refactored to 

```
search_case = search_case.where(status: statuses)
```

app/views/admin/shared/_sidebar.html.slim
corymcdonald: View code has already been deleted, I think we'd want a slight refactor, and expectations around how this would work 😄

app/views/admin/cases/show.html.slim
corymcdonald: This was a request from Asad to assist with training 👍

app/controllers/admin/cases_controller.rb
corymcdonald: That would be an `AND` statement, rather than an `OR` which would result in no results. 😅

app/controllers/admin/cases_controller.rb
corymcdonald: Sure thing! https://github.com/brave-intl/publishers/pull/2627/commits/4c5a3f976848852c38f029bd429474a475fb26e3

app/controllers/admin/cases_controller.rb
yachtcaptain23: `statuses` is an array of `Case.status` right?

```
2.5.7 :006 > Case.where(status: ["closed", "open"]).to_sql
 => "SELECT \"cases\".* FROM \"cases\" WHERE \"cases\".\"status\" IN ('closed', 'open')"
Case.where(status: ["closed", "open"]).count
=> 548 // proves that it's not an AND
```

corymcdonald: Wow! that's awesome, had no idea.

https://github.com/brave-intl/publishers/pull/2627/commits/b4dec27a671494c732b67884ec1db5aa1767ef27

PR #2631, opened 2020-04-14
Change dev dependencies to save 167 MB in memory

yachtcaptain23:


yachtcaptain23:
Going to revisit this later. Perhaps it's best to remove `YARN_PRODUCTION` so we get the devDependencies but it'll prune afterwards

PR #2638, opened 2020-04-17
Add more clarity to Statements

### LINKED ISSUE #2481: Labeling for Statements
Currently the labeling for statements makes it confusing to know exactly where the money is coming from, and the separation between Brave Funds and User Funds. 

We should update the statements and ensure that 

1. Labels need to make a lot more sense. Design should be involved :) 
2. The values in different currencies never change

### LINKED ISSUE #2533: Fix the balance view for the user funded vs Brave settled BAT
Issue: As we have moved to user funds and tips from user funds go directly to the creator wallet, we don't reflect what is Brave settled vs. user funded settlement to the creator
This was noted by Wikipedia who was able to see direct deposits in their account not via Brave and caused confusion

See image -- 
image

In this image, the total number is fine as we pull all user funded tips from Uphold so the creator sees the right total; however the split between what was settled via settlement and what was received via user tips (aka settled directly) is not split.

We need to be able to split this to show what is Brave settled vs. already deposited with a note that there is a 5% platform fee on both.

cc @jenn-rhim @corymcdonald 

Assigning to @jenn-rhim for design suggestions. @jenn-rhim please assign back to me when ready.

corymcdonald:
## Add more clarity to Statements

Closes #2533
Closes #2481

#### Screenshots
localhost_3000_publishers_statements_locale=en

localhost_3000_publishers_statements_Mar%202020%20-%20Apr%202020

Screen Shot 2020-04-17 at 10 33 09 AM

#### Features

✨ Support for different currencies paid out in one month
✨ Show breakdowns for where currencies come from 
🌐 Convert to using React-Intl
🐛 When we haven't paid out for a month but the user has gotten uphold tips do not show that we've paid them out.
♻️ Refactor the invoice generation code so that it's easier to read / follow
 

#### How To Test

1. Connect to production
1. Call `PublisherStatementGetter.new(publisher: publisher).perform.to_json`
1. Copy JSON locally to own StatementGetter, build out the statements
1. Run this branch


app/models/views/user/statement_overview.rb
```diff
@@ -6,49 +6,75 @@ class StatementOverview
...
         details = []
         grouped_transactions = settled_transactions.group_by { |x| x.transaction_type }
+        # Fees are only associated with Contribution Settlement, so let's group the fees explicitly with contributions.
+        fees = grouped_transactions.delete("fees")
 
         grouped_transactions.each do |type, results|
           total_amount = 0
 
+          # Add previously removed fees to the contribution settlement details
+          results += fees if type == PublisherStatementGetter::Statement::CONTRIBUTION_SETTLEMENT
```
yachtcaptain23, line 70: This seems like it's getting added multiple times?

app/models/views/user/statement_overview.rb
```diff
@@ -6,49 +6,75 @@ class StatementOverview
       # include PublishersHelper
 
       GROUP_START_DATE = Date.new(2019, 10, 1)
+      MONTH_FORMAT = "%b %Y".freeze
+      YEAR_FORMAT = "%b %e, %Y".freeze
```
yachtcaptain23, line 10: I think we should use a standard date format since we've never discussed this. I'm a fan of ISO 8601 which is YYYY-MM-DD or `%Y-%m-%d`

corymcdonald: This class has been kind of treated as a decorator for the view, but I'd be happy to not do any formatting and put the formatting in the React code if that would make more sense?

app/models/views/user/statement_overview.rb
corymcdonald: Agreed! It's confusing, I added a comment here with a lot more code samples.
https://github.com/brave-intl/publishers/pull/2638/commits/742c1d37bf672bbf6905e4d8561c89b0e12f9cda

Since we call `group_by` and iterate through each transaction type we only add this once

app/models/views/user/statement_overview.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2638/commits/4ca571bca6e8fc5ec8c8a9bd9694bde98804bce1

PR #2645, opened 2020-04-21
Feature/snapshot balances

yachtcaptain23:


app/services/eyeshade/create_snapshot.rb
```diff
@@ -0,0 +1,22 @@
...
+    else
+      begin
+        response = connection.post do |request|
+          request.headers["Authorization"] = api_authorization_header
+          request.url("/v1/snapshots/#{payout_report_id}")
+        end
+        result = JSON.parse(response.body)
+      end
+    end
+    result
```
corymcdonald, line 14: I think this code could be refactored to be only a couple of lines.

```suggestion
    return perform_offline if Rails.application.secrets[:api_eyeshade_offline]
    response = post("/v1/snapshots/#{payout_report_id}")
    result = JSON.parse(response.body)
```

PR #2690, opened 2020-05-11
Fix Statement Bugs

corymcdonald:
#### Features

🐛 UpholdConnectionForChannel was `BigInt` rather than `UUID`
🐛 Fix issue where only the first page would be returned if there weren't any previously cached values
🐛 Fix grouping of Statements due to DateTime vs Date mismatch
🐛 Only return and display transactions that were made from the browser to the uphold card
✨ Provide a link to the Uphold card that the deposits were made on. 
✨ Fix overview page to show more specifically the total brave deposited and Uphold tips

### DIFF
--- a/app/javascript/views/statements/Statements.tsx
+++ b/app/javascript/views/statements/Statements.tsx
@@ -1,6 +1,10 @@
 import * as moment from "moment";
 import * as React from "react";
-import { FormattedMessage, FormattedNumber, injectIntl } from "react-intl";
+import {
+  FormattedMessage,
+  FormattedNumber,
+  injectIntl,
+} from "react-intl";
 import ReactTooltip from "react-tooltip";
 
 import { DownloadIcon } from "brave-ui/components/icons";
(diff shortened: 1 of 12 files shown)


app/services/publisher_statement_getter.rb
```diff
@@ -27,7 +27,7 @@ def earning_period
       if eyeshade_settlement? || fee?
         created_at.prev_month.at_beginning_of_month
```
yachtcaptain23, line 28: Should we also include `to_date` here?

yachtcaptain23, line 28:
```suggestion
        created_at.prev_month.at_beginning_of_month.to_date
```

corymcdonald: https://github.com/brave-intl/publishers/pull/2690/commits/ec797dc64d5f339470d54e9ede1bb264c81f2c52

corymcdonald: https://github.com/brave-intl/publishers/pull/2690/commits/ec797dc64d5f339470d54e9ede1bb264c81f2c52

app/javascript/views/statements/Statements.tsx
```diff
@@ -65,6 +71,49 @@ export const DisplayPaymentDate = (paymentDate: string) => {
...
 };
 
+export const GetParameterCaseInsensitive = (object, key) => {
+  return object[
+    Object.keys(object).find((k) => k.toLowerCase() === key.toLowerCase())
+  ];
+};
+
+export const DepositBreakdown = (props) => {
+  const id = props.name + "-" + Math.random();
```
yachtcaptain23, line 81: If there's an identical id, nothing should happen right?

corymcdonald: Correct, there would only be 2 things hovering over each other, so not that serious of a problem just a visual bug that `Math.random()` helps assist with

app/models/views/user/statements.rb
```diff
@@ -63,6 +65,8 @@ def build_statement_overviews(earning_periods)
             overview.total_earned += payout_entry.amount.abs
           end
 
+          overview.settlement_destination = settlement_destination
```
yachtcaptain23, line 68:
```suggestion
          overview.settlement_destination = settlement_destination # Assume that settlement_destination doesn't change
```

Issue #2364, opened 2019-11-06
Channels list V4 API


yachtcaptain23:
Will update performance info with benchmarks, but slimming down the channels list as much as possible is ideal.

Benchmarks (Newer iPhones on 130 mbps download):
~ 5s download, 30s - 40s database insert

yachtcaptain23:
Proposed solution 1:
Create the following endpoints:
`api/v4/public/channels/youtube`
`api/v4/public/channels/sites`
`api/v4/public/channels/excluded_sites`
`api/v4/public/channels/vimeo`
...

```
# V4 Version of Channels list
#
# Each channel is an array:
# [
#   channel_id (string),
#   connected state (enum where 0 is PUBLISHER_NOT_VERIFIED, 1 is CONNECTED, 2 is VERIFIED)
#   excluded (numerator where 0 is false, 1 is true),
#   wallet address (string)
#   site_banner details (hash)
# ]
```

Examples:
For `api/v4/public/channels/youtube`
`["UCr_USjgn4PQhVpqOT6RcAtQ",1,0,"6131cad3-fe35-4c75-b5a0-bd1febe74b35",{}]`

NejcZdovc:
I think that we shouldn't try to improve current list approach, but go with api approach. We could mimic cache on client like we do on desktop, so to only hit api after 24h when publisher cache on client expires and user visits the site again. We would have same mechanism that we have now to force fetch publisher via the panel

NejcZdovc:
Also based on what I just read from second comment this is a huge change on the client side

corymcdonald:
I'm wondering what the privacy implications of having an API for this. Based on this paper it looks like SafeBrowsing can track users websites, Yandex being the worse offender of this. https://hal.inria.fr/hal-01120186/file/RR8686.pdf

However, if we proxy through Fastly, similar to the way we currently are doing, then we wouldn't be tracking users.

NejcZdovc:
@corymcdonald @evq has a proposal how to do this with privacy in mind. Why I am pushing for alternative approach (not need to be API, could be x) and not improve list approach is because it would take a lot of time on both sides and it's just a temp solution as list is just not scaling

yachtcaptain23:
[...]

PR #2697, opened 2020-05-13
Feature: V4 Channels prefix list. Major performance improvements focus

### LINKED ISSUE #2660: Prefix Date Diffing
Fetching the diff:
`prefixes-Y-M-D.br`
e.g. `prefixes-2020-03-30.br`

Fetching whole file:
`prefixes.br`

yachtcaptain23:
for both server side costs and client side computation.

* Padding tests and file cleanup tests
* Handle after_save for changes for uphold_connection and paypal_connection changes
* Protobuf file for prefix list
* Differential data

Closes #2660, #2364


app/jobs/cache/browser_channels/main.rb
```diff
@@ -0,0 +1,21 @@
...
+  RESPONSES_PREFIX_LENGTH = 3
+  def perform
+    previous_run = Rails.cache.fetch(LAST_RAN_AT_KEY)
+    return if previous_run.present? && previous_run.to_time >= 2.hours.ago
+    Cache::BrowserChannels::PrefixList.perform_async
+    sql = "SELECT DISTINCT SUBSTRING(sha2_base16, 1, #{RESPONSES_PREFIX_LENGTH}) as prefix FROM site_banner_lookups"
+    sql += " WHERE wallet_status != 0"
+    sql += " AND updated_at >= '#{(previous_run.to_time - 3.minutes).to_s}'" if previous_run.present?
+    sql += " ORDER BY prefix desc"
+    result = ActiveRecord::Base.connection.execute(sql)
```
corymcdonald, line 15: One thing that I noticed while [reading the docs](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/DatabaseStatements.html#method-i-execute) on this function call is memory might need to be cleared after this call.

>Note: depending on your database connector, the result returned by this method may be manually memory managed. Consider using the exec_query wrapper instead.

It's fine to keep this as is but a nit I have is that this isn't using SQL parameters. 

I think I'd prefer to see something like this
```ruby
if previous_run.present?
  previous_run.to_time - 3.minutes)
else 
  previous_run = Time.at(0)
end

ActiveRecord::Base.connection.execute("
  SELECT DISTINCT SUBSTRING(sha2_base16, 1, :response_prefix_length) as prefix FROM site_banner_lookups
  WHERE wallet_status != 0
  AND updated_at >= :updated_at
  ORDER BY prefix desc",
  response_prefixe_length: RESPONSES_PREFIX_LENGTH,
  updated_at: previous_run
)
```

yachtcaptain23: I checked the documentation, since we get a `PG::Result`, the documentation says it'll get added to Ruby's garbage collector: https://www.rubydoc.info/gems/pg/PG/Result 

My impression is that Ruby uses the mark & sweep garbage collection algorithm, so I think we'll be okay.

protos/publisher_list.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+option optimize_for = LITE_RUNTIME;
+
+package publishers_pb;
+
+message PublisherList {
```
zenparsing, line 7: Please use the current version of the proto definition file [here](https://docs.google.com/document/d/19zp-GSqPzp6mqFBcy482RUrVM0gEi7HySxSWlmM1Vmg/edit). Note that `publisher_type` was removed (it can be added in a future version if we have a use for it).

Also, note that `BROTLI_COMPRESSION` is not supported yet. Let's get raw encoding working first. Also, I have another compression technique I'm working on that will compress the list better than Brotli and much faster.

protos/channel_responses.proto
```diff
@@ -0,0 +1,40 @@
...
+  repeated int32 donation_amounts = 5;
+  SocialLinks social_links = 6;
+}
+
+enum WalletConnectedState {
+  NO_VERIFICATION = 0;
+  UPHOLD_ACCOUNT_NO_KYC = 1;
+  UPHOLD_ACCOUNT_KYC = 2;
+  /* (Albert Wang) These might be used */
+  PAYPAL_ACCOUNT_NO_KYC = 3;
```
zenparsing, line 26: Will these be included in the list? We don't currently support "paypal" as such on the client side.

protos/channel_responses.proto
```diff
@@ -0,0 +1,40 @@
...
+  PAYPAL_ACCOUNT_KYC = 4;
+}
+
+message ChannelResponse {
```
zenparsing, line 30: Question: why do we call them "channels" here but "publishers" elsewhere? Should we call them just one thing?

protos/channel_responses.proto
```diff
@@ -0,0 +1,40 @@
...
+  string channel_identifier = 1;
+  WalletConnectedState wallet_connected_state = 2;
+  string wallet_address = 3;
+  bool publisher_ads_opted_in = 4;
+  SiteBannerDetails site_banner_details= 5;
+}
+
+message ChannelResponses {
+  repeated ChannelResponse channel_response = 1;
```
zenparsing, line 39: Repeated fields should have a plural name, for better code readability. It would be awkward to have `channel_responses` as the only field in `ChannelResponses` though. Can we do some name changes here?

Perhaps:

```
message ChannelResponse {
  repeated Channel channels = 1;
}
```

Any other ideas?

app/jobs/cache/browser_channels/prefix_list.rb
```diff
@@ -0,0 +1,85 @@
...
+    s3 = Aws::S3::Resource.new(region: Rails.application.secrets[:s3_rewards_bucket_region])
+    obj = s3.bucket(Rails.application.secrets[:s3_rewards_bucket_name]).object(save_to_filename)
+    obj.upload_file(temp_file_path)
+  end
+
+  private
+
+  def to_protobuf_file(result)
+    publisher_list_pb = PublishersPb::PublisherList.new
+    publisher_list_pb.compression_type = PublishersPb::PublisherList::CompressionType::BROTLI_COMPRESSION
```
zenparsing, line 75: Let's get raw working before we add support for compression - and I think we can do better than Brotli anyway (I'll write this up today).

app/jobs/cache/browser_channels/prefix_list.rb
```diff
@@ -0,0 +1,85 @@
...
+    result = ActiveRecord::Base.connection.execute("
+        SELECT SUBSTRING(sha2_base16, 1, #{PREFIX_LENGTH})
+        FROM site_banner_lookups
+        WHERE wallet_status != #{PublishersPb::WalletConnectedState::NO_VERIFICATION}"
+        ).map { |r| r['substring'] }.to_json
+
+    to_protobuf_file(result)
+  end
+
+  def save_differential_file!
```
zenparsing, line 43: Differential lists will be a follow-up on the client side. We should probably hold off on this until we can discuss what would work best on the client.

app/jobs/cache/browser_channels/prefix_list.rb
```diff
@@ -0,0 +1,85 @@
...
+    save_to_s3!(temp_file_path: temp_file.path, save_to_filename: "prefixes-#{date}")
+    cleanup!(temp_file_path: temp_file)
+  end
+
+  def save_main_file!
+    result = ActiveRecord::Base.connection.execute("
+        SELECT SUBSTRING(sha2_base16, 1, #{PREFIX_LENGTH})
+        FROM site_banner_lookups
+        WHERE wallet_status != #{PublishersPb::WalletConnectedState::NO_VERIFICATION}"
+        ).map { |r| r['substring'] }.to_json
```
zenparsing, line 38: Instead of JSON, we need to sort the prefixes and then hex decode them so that we have a raw byte string. I think this can be done in Ruby like so:

```rb
# For an array of hex prefixes |hex_values|
raw_values = hex_values.sort().map { |item| [item].pack('H*') }

# Joining the array elements into one long raw string
raw_data = raw_values.join()
```

app/jobs/cache/browser_channels/main.rb
```diff
@@ -0,0 +1,30 @@
+class Cache::BrowserChannels::Main
+  include Sidekiq::Worker
+  sidekiq_options queue: :scheduler, retry: false
+
+  LAST_RAN_AT_KEY = "cache_browser_channels_main_last_ran_at".freeze
+  RESPONSES_PREFIX_LENGTH = 3
```
zenparsing, line 6: Why is this 3? Three hex chars will result in around ~170 channels per response, whereas 4 hex chars will result in about 11. Are we trying to limit the number of files generated?

app/jobs/cache/browser_channels/prefix_list.rb
```diff
@@ -0,0 +1,85 @@
...
+  require './protos/channel_responses'
+rescue
+end
+
+class Cache::BrowserChannels::PrefixList
+  include Sidekiq::Worker
+  sidekiq_options queue: :low, retry: false
+
+  # Might need to adjust the value based on 
+  PREFIX_LENGTH = 9
```
zenparsing, line 18: Please make this 8, so that prefixes are 4 bytes as in Safe Browsing and can be efficiently compressed as numbers. With 4 bytes we can expect a collision after around 92,000 distinct lookups.

protos/channel_responses.proto
yachtcaptain23: We likely will include Stripe. I personally don't mind since these fields are optional, but I'll remove to keep it clean and add once it is necessary.

app/jobs/cache/browser_channels/prefix_list.rb
yachtcaptain23: Vast majority of the space savings comes from differential data. It's a core piece and the implementation that we do client side should use differential data.

app/jobs/cache/browser_channels/prefix_list.rb
yachtcaptain23: I came up with a different answer, where there will be a 50% chance of a collision after 2976 lookups on a 1 million in our publisher list and 297 once we hit 10 million

I used the birthday problem formula: 0.5 = (1 - P)^x

https://www.wolframalpha.com/input/?i=ln%280.5%29+%2F+ln%281+-+%281000000+%2F+16%5E8%29%29

I'm reluctant to move down to 8 nibbles if you're proposing a certain compression that will only work for 8 nibbles. My impression is that things will change quite a bit if we need to move from 8 to 9. The numbers we'll see will vastly outpace Safe Browsing due to the number of individual channels and accounts on Youtube, Reddit, Github, etc.

protos/channel_responses.proto
yachtcaptain23: So on the server-side, publisher refers to the person or organization. Channel is the domain representation. I realize it's confusing, so I can make the change from PublisherResponse to ChannelResponse if you desire.

protos/channel_responses.proto
yachtcaptain23: I think it should be 
```
message ChannelResponses {
  repeated ChannelResponse channel_responses = 1;
}
```

app/jobs/cache/browser_channels/prefix_list.rb
yachtcaptain23: Sorting is resolved.

protos/channel_responses.proto
zenparsing: No, I think `ChannelResponse` is fine.

protos/channel_responses.proto
zenparsing: To avoid ChannelResponses::channel_responses, how about this:

```
message ChannelResponseList {
  repeated ChannelResponse channel_responses = 1;
}
```

yachtcaptain23: Fine with me, I'll merge the change.

app/jobs/cache/browser_channels/prefix_list.rb
zenparsing: Regardless of the prefix size we choose, it needs be a whole number of bytes (an even number of nibbles). Can we set `PREFIX_LENGTH` to 8 and adjust upward if needed based on testing? You might also want to set these constants as a number of bytes and then multiply by 2 in the query in order to avoid confusion down the road.

app/jobs/cache/browser_channels/prefix_list.rb
```diff
@@ -0,0 +1,93 @@
...
+    save_to_s3!(temp_file_path: temp_file.path, save_to_filename: "prefixes-#{date}") # 2020-05-17
+    cleanup!(temp_file_path: temp_file)
+  end
+
+  def save_main_file!
+    result = ActiveRecord::Base.connection.execute("
+        SELECT SUBSTRING(sha2_base16, 1, #{PREFIX_LENGTH})
+        FROM site_banner_lookups
+        WHERE wallet_status != #{PublishersPb::WalletConnectedState::NO_VERIFICATION}"
+    ).map { |r| r['substring'] }.sort!
```
zenparsing, line 43: We need to decode the hex. Does the following work?

```rb
).map { |r| r['substring'] }.sort!.map { |item| [item].pack('H*') }
```

app/jobs/cache/browser_channels/prefix_list.rb
```diff
@@ -0,0 +1,93 @@
...
+  def to_protobuf_file(result)
+    publisher_list_pb = PublishersPb::PublisherList.new
+    publisher_list_pb.compression_type = @compression_type
+    if @compression_type == PublishersPb::PublisherList::CompressionType::NO_COMPRESSION
+      publisher_list_pb.prefixes = result.join("")
+    elsif @compression_type == PublishersPb::PublisherList::CompressionType::BROTLI_COMPRESSION
+      publisher_list_pb.prefixes = Brotli.deflate(result.to_json)
+    end
+    publisher_list_pb.prefix_size = PREFIX_LENGTH
+    publisher_list_pb.uncompressed_size = result.length
```
zenparsing, line 87: This should be the uncompressed length of the joined prefix string, instead of the number of prefixes. So `result.join("").length`? I suppose you'll want a temporary to hold the `result.join("")` value.

app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -0,0 +1,109 @@
...
+    end
+  end
+
+  # We want to hide which file is being downloaded by making all requests be the same size
+  def pad_file!
+    path = @temp_file.path
+    # Round up to nearest KB
+    file_size = File.size(path)
+    # Converts size from like 326 to 1000
+    new_size = (((file_size + 1000) / 1000)) * 1000
```
zenparsing, line 63: Nit: 1024 for KB?

protos/channel_responses.proto
zenparsing: Since the client isn't going to know how to deal with these values, it would be fine to take them out (and would leave more options open for later).

app/jobs/cache/browser_channels/prefix_list.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  def to_protobuf_file(result)
+    publisher_list_pb = PublishersPb::PublisherList.new
+    publisher_list_pb.compression_type = @compression_type
+    if @compression_type == PublishersPb::PublisherList::CompressionType::NO_COMPRESSION
+      publisher_list_pb.prefixes = result.join("")
+    elsif @compression_type == PublishersPb::PublisherList::CompressionType::BROTLI_COMPRESSION
+      publisher_list_pb.prefixes = Brotli.deflate(result.to_json)
+    end
+    publisher_list_pb.prefix_size = PREFIX_LENGTH
```
zenparsing, line 86: This should be in bytes instead of nibbles.

app/jobs/cache/browser_channels/prefix_list.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  private
+
+  def to_protobuf_file(result)
+    publisher_list_pb = PublishersPb::PublisherList.new
+    publisher_list_pb.compression_type = @compression_type
+    if @compression_type == PublishersPb::PublisherList::CompressionType::NO_COMPRESSION
+      publisher_list_pb.prefixes = result.join("")
+    elsif @compression_type == PublishersPb::PublisherList::CompressionType::BROTLI_COMPRESSION
+      publisher_list_pb.prefixes = Brotli.deflate(result.to_json)
```
zenparsing, line 84: If you want to try out delta encoding, I think this function will work to create a delta list:


```rb
def to_delta_list(list)
  last = 0
  return list.map do |item|
    # Read prefix as big-endian uint32
    value = item.unpack('N').first
    if (value < last)
      raise 'Input list out of order'
    end
    # Get delta from last value
    delta = value - last
    last = value
    delta
  end
end
```

and then instead of setting `publisher_list_pb.prefixes`, just do:

```rb
publisher_list_pb.deltas = to_delta_list(result)
```

app/jobs/cache/browser_channels/responses_for_prefix.rb
yachtcaptain23: Let's keep it to base_10 for simplicity sake? It's not cut-and-dry: https://en.wikipedia.org/wiki/Kilobyte

corymcdonald:
What kind of monitoring can we introduce to know if there is a problem? Is there any smoke tests that we can introduce?

yachtcaptain23:
> What kind of monitoring can we introduce to know if there is a problem? Are there any smoke tests that we can introduce?

Upload job will be monitored in Sentry/NewRelic.

File lands in S3 and through Fastly. We can make a timeliness check by running scripts to make sure the file gets uploaded via a cronjob in AWS Lambda. I don't think Pingdom supports this functionality, but it'll be fairly easy for us to check timeliness with AWS Lambda cron jobs.

app/models/channel.rb
```diff
@@ -271,6 +274,25 @@ def register_channel_for_promo
...
+    )
+    site_banner_lookup.set_sha2_base16
+    site_banner_lookup.set_wallet_status(publisher: publisher)
+    site_banner_lookup.derived_site_banner_info =
+      if skip_site_banner_info_lookup
+        {}
+      else
+        site_banner&.non_default_properties || publisher&.default_site_banner&.non_default_properties || {}
+      end
+    site_banner_lookup.update(
```
corymcdonald, line 289:
```suggestion
    site_banner_lookup.update!(
```

We probably want this to raise an exception so the rest of the data doesn't save if this fails?

app/models/channel.rb
```diff
@@ -271,6 +274,25 @@ def register_channel_for_promo
     Promo::RegisterChannelForPromoJob.perform_now(channel_id: id, attempt_count: 0)
   end
 
+  def update_site_banner_lookup!(skip_site_banner_info_lookup: false)
+    site_banner_lookup = SiteBannerLookup.find_or_initialize_by(
+      channel_identifier: details&.channel_identifier || details.brave_publisher_id,
```
corymcdonald, line 279: I think we want this strictly to be 
```suggestion
      channel_identifier: details&.channel_identifier,
```

if there's a problem with the  channel_identifier not working for `SiteChannelDetails` then we should probably fix that rather than have this `||`?

config/brakeman.ignore
```diff
@@ -1,5 +1,25 @@
 {
   "ignored_warnings": [
+    {
```
corymcdonald, line 3: Instead of ignoring these warnings can you pass in the parameters into the execute commands? Something like this:

```ruby
    result = ActiveRecord::Base.connection.execute("
      SELECT DISTINCT SUBSTRING(sha2_base16, 1, #{RESPONSES_PREFIX_LENGTH * 2}) as prefix
      FROM site_banner_lookups
      WHERE wallet_status != 0
      AND updated_at >= :previous_run'
      ORDER BY prefix desc",
      previous_run: previous_run
    )
```

If that doesn't work [this article has more information](https://blog.saeloun.com/2019/10/28/bind-parameters-in-activerecord-sql-queries.html) about binding parameters.

test/jobs/cache/browser_channels/prefix_list_test.rb
```diff
@@ -0,0 +1,35 @@
+require 'test_helper'
+
+class Cache::BrowserChannels::PrefixListTest < ActiveSupport::TestCase
```
corymcdonald, line 3: It seems we're checking to make sure the prefix list can compress and decompress. Can we have a test that checks to make sure that content is correct in the prefix list? Like generating the right hashes?

app/jobs/cache/browser_channels/prefix_list.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def perform
+    # We only care about first PREFIX_LENGTH number of nibbles of the prefix.
+    # Don't waste memory reading excess
+    temp_file = save_main_file!
+    save_to_s3!(temp_file_path: temp_file.path, save_to_filename: "prefixes")
+    cleanup!(temp_file_path: temp_file)
+
+    temp_file = save_differential_file!
```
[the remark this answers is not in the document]
yachtcaptain23: That shouldn't happen unless sidekiq didn't get an exception.

At the moment, browser doesn't request the differential file.

test/jobs/cache/browser_channels/prefix_list_test.rb
yachtcaptain23: Already covered by https://github.com/brave-intl/publishers/pull/2697/files#diff-69172adae94b07bb140d87297dfd3052R24

PR #2703, opened 2020-05-13
Backfill Publisher Statuses

corymcdonald:
#### Features

Publisher Statuses that were never created are causing some issues

### DIFF
--- a/lib/tasks/database_updates/backfill_publisher_statuses.rake
+++ b/lib/tasks/database_updates/backfill_publisher_statuses.rake
@@ -11,7 +11,7 @@ namespace :database_updates do
     publishers.find_each do |publisher|
       note = publisher.notes.create(note: "Backfilling active status", created_by_id: admin.id)
       publisher.status_updates.create(status: PublisherStatusUpdate::ACTIVE, publisher_note: note)
-      puts 'Created'
+      puts "Created #{publisher.id}"
     end
 
     puts 'Done!'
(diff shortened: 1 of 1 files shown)


lib/tasks/database_updates/backfill_publisher_statuses.rake
```diff
@@ -0,0 +1,19 @@
...
+
+    publishers = Publisher.all.
+        joins('LEFT OUTER JOIN publisher_status_updates ON publisher_status_updates.publisher_id = publishers.id').
+        where('publisher_status_updates.publisher_id is NULL').
+        order(created_at: :desc)
+
+    publishers.find_each do |publisher|
+      note = publisher.notes.create(note: "Backfilling active status", created_by_id: admin.id)
+      publisher.status_updates.create(status: PublisherStatusUpdate::ACTIVE, publisher_note: note)
+      puts 'Created'
```
yachtcaptain23, line 14: Would you put in the publisher_id so we know where it stopped if it does get stop in the middle?

Also would be good for the script to be retried with parameters.

corymcdonald: https://github.com/brave-intl/publishers/pull/2703/commits/637f91ebcdf491404b19814e9cf79aae19b40bf7

I think we'll be okay with re-running the script without passing in any arguments. The query will return publishers who don't currently have a status. So re-running this script will only target the publishers not yet touched by the script.

yachtcaptain23:
Btw, if this ever keeps going down, then it might be worth while to refactor to do bulk insert: https://blog.bigbinary.com/2019/04/15/bulk-insert-support-in-rails-6.html

PR #2705, opened 2020-05-13
Perf: Don't fetch extra content on home

### LINKED ISSUE #1719: Harden publishers#home for 503 errors from Ledger/Eyeshade

yachtcaptain23:
Closes #1719


corymcdonald:
Can you look into the `home.js` file? There's a lot of data that gets fetched and used to update the `lastSettlementBalance`. We likely don't need that anymore?

app/views/publishers/home.html.slim
```diff
@@ -32,61 +32,8 @@ script type="text/html" id="confirm_default_currency_modal_wrapper"
...
-                        span == t('.rejected_funds_description')
-                      -else
-                        span.warning-icon.mr-2
-                          = render 'icon_circled_bang_warning'
-                        span
-                          == t('.held_funds_description')
-
-
+        = render_async home_balances_publishers_path do
+          h3 Loading balances...
```
corymcdonald, line 36: Instead of having this display "Loading balances" without any styling or anything let's do the following

1. Have the white panel still a part of the main page. 
2. Include a spinner icon (I believe we have one somewhere in our project already)
3. Internationalize the text.

app/views/publishers/home.html.slim
```diff
@@ -104,9 +51,9 @@ script type="text/html" id="confirm_default_currency_modal_wrapper"
     .row
       .col-md.mb-4
         - if !(@publisher.excluded_from_payout? || @publisher.paypal_locale?(params[:locale]) || @publisher.uphold_connection&.japanese_account? || @publisher.wire_only?)
-          = render partial: "uphold_wallet_panel"
+          = render_async uphold_wallet_panel_publishers_path
```
corymcdonald, line 54: Let's be sure that we show the white panel so that the page makes sense while we're loading this and additionally have the same behavior as the balance panel. 

For bonus points we could create a loading partial that we render anywhere that we're using `render_async`.

I would imagine that it should contain 

- the spinner
- internationalized loading text

config/locales/en.yml
```diff
@@ -391,6 +391,9 @@ en:
       heading: Reject transfer
       reject: Reject
   publishers:
+    home_balances:
```
corymcdonald, line 394: Do translations need to be moved  for the `ja.yml` as well?

db/schema.rb
```diff
@@ -457,12 +459,28 @@
...
     t.text "data"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
+    t.datetime "created_at"
+    t.datetime "updated_at"
     t.index ["session_id"], name: "index_sessions_on_session_id", unique: true
     t.index ["updated_at"], name: "index_sessions_on_updated_at"
   end
 
+  create_table "site_banner_lookups", id: :uuid, default: -> { "uuid_generate_v4()" }, force: :cascade do |t|
```
corymcdonald, line 468: Can you revert this file? I don't think anything in here should change

yachtcaptain23:
> Can you look into the `home.js` file? There's a lot of data that gets fetched and used to update the `lastSettlementBalance`. We likely don't need that anymore?

It seems to still be used after the user updates their balance.

app/views/browser_users/home.html.slim
```diff
@@ -6,5 +6,5 @@
     .p-2
       = render partial: 'how_it_works', referral_code: @promo_registration
     .p-2
-      = render partial: 'pending_balance'
+      = render_async browser_users_pending_balance_path
```
corymcdonald, line 9: Let's be sure to add a loading section to this `render_async` too.

app/services/ratio/ratio.rb
```diff
@@ -21,6 +22,13 @@ def relative(currency:)
       JSON.parse(response.body)
     end
 
+    def self.relative_cached(currency:)
```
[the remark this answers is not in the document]
yachtcaptain23: I honestly hate the way we setup the PaymentClient services and think it's horrible for a maintenance standpoint. Sure I understand wanting to stay DRY, but having it be set during the initializer makes our code approach more like spaghetti.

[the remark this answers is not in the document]
yachtcaptain23: Anyone else who jumps in on the project and doesn't realize we setup an initializer arbitrarily would have a hard time figuring out why we're making something so seemingly simple so complex, similarly to how we use Addressable for building out urls rather than using URI.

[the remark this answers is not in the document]
corymcdonald: It seems we disagree on the right approach here. I want us to be on the same page and work towards building a code base we both agree is quality.  Can you explain more about your feelings towards maintenance of the services? It would be great to hear some ideas on how you feel like we can make this code less complex, and easier to onboard new developers. If it's easier we can hop on a call and discuss. 😄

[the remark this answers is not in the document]
corymcdonald: I think using this client structure for our services makes things a little easier to maintain. 
Instead of having calls to the same services in multiple different places with names that don't associate with the service `PublisherWalletGetter`, `PublisherTransactionGetter` the service client scopes the code under the service name. One problem that I have under the previous approach is that it's hard to answer the question "Does Publishers call this endpoint?". 

In my experience having a single initializer for a client makes it easy to pass configuration in based on the environment. Because there's a single object representing the service it makes it easier to mock service calls out in tests. It's the same approach set forth by most ruby gems which connect to external services. A few good examples are [Yt](https://github.com/brave-intl/publishers/blob/master/config/initializers/yt.rb), [Stripe-ruby](https://stackoverflow.com/questions/35800053/stripe-on-ruby-on-rails-creating-config-initializers-stripe-rb), and [AWS SDK](https://github.com/thepracticaldev/dev.to/blob/master/config/initializers/aws_sdk.rb) gems. It's also nice to have a single object that you can search for (e.g. `PaymentClient`) to see all the instances where the service is being called. 

I've also found some app methodologies that [making no distinction between local and third party services, treating them as attached resources, and loosely coupling them](https://12factor.net/backing-services) is recommend practice. 

We don't have to use addressable. I thought it made things a little easier when dealing with complicated URIs that require multiple substitution, but when you're implementing these methods we can more simply pass a string into any of the `get(url)`, `post(url)`, etc methods.

corymcdonald:
Most of the code surrounding this was refactored in the [New Wallet Services](https://github.com/brave-intl/publishers/pull/2811/commits/6b37c940de0c35a0e93c473226f6e3e24694d0d5) PR.

PR #2723, opened 2020-05-27
Release 2020-05-27

corymcdonald:
## Release 2020-05-27

🐛 Reduce number of times refresh stats is called.


corymcdonald:
Deployed this as a temporary stopgap, the database duration was causing requests to enqueue for 5-10 seconds and then finally get served. This should stabilize our app.

image

PR #2735, opened 2020-06-03
Change http to https on meta og:url tag

AlanBreck:
Our SEO consultant flagged the `og:url` meta tag on creators.brave.com as being an SEO concern due to using `http` instead of `https`.


AlanBreck:
> Actually, can we use `og:secure_url`? https://ogp.me/ suggests to do so

Done.

yachtcaptain23:
After discussion, it seems like the format is `og:url`

AlanBreck:
Changes made and ready for review.

PR #2760, opened 2020-06-16
Add mock data for staging environment

yachtcaptain23:
We want to simulate fake prefix list encoding and decoding.


app/jobs/cache/browser_channels/prefix_list.rb
```diff
@@ -31,7 +31,17 @@ def save_main_file!
         {
           nibble_length: PREFIX_LENGTH * 2,
         }
-    ]).map { |r| r['substring'] }.sort!
+    ]).map { |r| r['substring'] }.uniq.sort!
+
+    if ENV["RAILS_ENV"].in?(["staging"])
```
corymcdonald, line 36:
```suggestion
    if Rails.env.staging?
```

PR #2761, opened 2020-06-17
Fix channel transfers

### LINKED ISSUE #2758: Refactor Channel Transfer Job so that channels will continue to transfer
`ApproveChannelTransfer` should be a job

corymcdonald:
## Fix channel transfers

Closes #2758

#### Features

- Move Channel Transfers to it's own job queue so there is retry logic


config/sidekiq.yml
```diff
@@ -56,9 +56,6 @@
     cron: "0 4 * * *"
     description: "For Publishers who have stale uphold codes or access parameters, clean out uphold codes or access parameter."
     queue: scheduler
-  TransferChannelsJob:
```
evq, line 59: this was just a duplicate entry - right?

app/jobs/channels/approve_channel_transfer_job.rb
```diff
@@ -1,41 +1,37 @@
 module Channels
-  class ApproveChannelTransfer < BaseService
```
yachtcaptain23, line 2: Would you add `sidekiq_options retry: X` here or to `application_job.rb`? I don't think there isn't a default set. The `max_retries` in sidekiq.yml seems to be fooling us: https://github.com/mperham/sidekiq/wiki/Active-Job#customizing-error-handling

corymcdonald:
We also should remove the SlackMessenger, or at least harden it from 429's.  If the Slack API fails to send then the entire transaction fails to go through. 😩

app/jobs/channels/approve_channel_transfer_job.rb
corymcdonald: Based off the documentation It seems like ActiveJob will automatically retry and then kick it over to sidekiq? 

> The default AJ retry scheme is 3 retries, 5 seconds apart. Once this is done (after 15-30 seconds), AJ will kick the job back to Sidekiq, where Sidekiq's retries with exponential backoff will take over.

Where then Sidekiq will retry up to `max_retries`

config/sidekiq.yml
corymcdonald: Yep!

test/jobs/channels/transfer_channels_job_test.rb
```diff
@@ -28,11 +28,21 @@ class TransferChannelsJobTest < ActiveJob::TestCase
...
+      Channels::TransferChannelsJob.perform_now
+
+      assert_enqueued_jobs 1
+      assert_performed_jobs 0
+
+      perform_enqueued_jobs
+
+      # There are 5 jobs enqueued (4 emails and a slack message) when a channel completes transfer
+      # + 2 for the actual jobs that were completed, so 7 overall
+      assert_performed_jobs 7
```
nvonpentz, line 45: It looks like we removed the slack messenger call so wouldn't the jobs be lower?

corymcdonald: Normally yes, but we converted `Channels::ApproveChannelTransferJob` to be a job and called `perform_later` to enqueue it 😄

evq: @corymcdonald can you adjust the comment above to match?

corymcdonald: We actually still do send a slack message during this process because we have an `after_save` event on `Channel` which triggers when the transferred channel verifies 😩 

https://github.com/brave-intl/publishers/blob/staging/app/models/channel.rb#L58

PR #2763, opened 2020-06-17
Add check for deleted cards

### LINKED ISSUE #2757: Detect publishers with deleted cards
Publishers whose Uphold cards were removed cannot be paid until we create them a new set of channel / owner cards.

corymcdonald:
## Add check for deleted cards

Closes #2757 

#### Features

- ♻️ Refactor "UpholdClient" usage so new API clients aren't being instantiated into memory all the time
- 🐛 Add the check for the missing cards


app/controllers/api/v1/transactions_controller.rb
```diff
@@ -3,9 +3,12 @@ class GetTransactionError < StandardError; end
 
   def show
     channel = get_channel!(params[:channel])
-    uphold_client = Uphold::Client.new(uphold_connection: channel.publisher.uphold_connection)
+    transaction = UpholdClient.transaction.find(
```
yachtcaptain23, line 6: This seems backwards and confusing from a mental model standpoint. `UpholdClient::Transaction.find` would be preferable

evq: I chatted with Cory to get his help understanding what `UpholdClient::Transaction.find` would mean in Ruby. It sounds like this would mean we're using a method on the class rather than an `UpholdClient` instance.

Using a class method would be more confusing to me here - we have a client that is instantiated using the publishers uphold connection parameters and we are using that instantiated client to find a transaction. Perhaps my one request would be to lean even more heavily in the instance method direction and change `UpholdClient` to `uphold_client`. I am also okay with status quo as it is still clear that an instance method is being used :smile:

corymcdonald: Albert and I discussed offline and I created this spike to investigate. Due to the structure and not using `self` methods it is possible for someone implementing a new API client to introduce shared variable state if they manipulate an object. It's rare but the code design could be improved. On the other hand the most popular ruby API clients use the same architecture and structure we have implemented. 

https://github.com/brave-intl/publishers/issues/2779

PR #2770, opened 2020-06-23
Fastly trusted_proxies list

corymcdonald:
#### Features

Support Fastly being fronted by publishers by adding the list of proxy IPs to Rails trusted proxies list 

#### How To Test

1. Deploy to staging
2. Ensure you can access the rails admin console using the brave VPN


config/initializers/fastly.rb
```diff
@@ -0,0 +1,14 @@
+FASTLY_API = "https://api.fastly.com/public-ip-list".freeze
```
nvonpentz, line 1: May want this to be an environment variable

PR #2771, opened 2020-06-23
Wallet state expanded

yachtcaptain23:


protos/channel_responses.proto
```diff
@@ -14,22 +14,41 @@ message SiteBannerDetails {
...
+  UpholdWalletState wallet_state = 2;
+}
+
+enum PaypalWalletState {
+  PAYPAL_ACCOUNT_NO_KYC = 0;
+  PAYPAL_ACCOUNT_KYC = 1;
+}
+
+message PaypalWallet {
+  string address = 1;
```
NejcZdovc, line 37: we don't need this one

zenparsing: If we remove this we could put the wallet state fields before the address fields for consistency.

NejcZdovc: agree as we will always have state for all wallets

PR #2780, opened 2020-06-26
Move Publishers to active status on promo after KYC

corymcdonald:
#### Features

If a user is in `no_ugp` state on Promo, after the user has KYC'd we will call `Promo::UpdateStatus` to update their status and move them back to an active status.


app/models/uphold_connection.rb
```diff
@@ -196,6 +197,13 @@ def update_site_banner_lookup!
     publisher.update_site_banner_lookup!
   end
 
+  # Internal: If the publisher previously had referral codes and then we will re-activate their referral codes.
+  #
+  # Returns nil
+  def update_promo_status
+    publisher.update_promo_status
```
yachtcaptain23, line 204: Nit: Change this to `update_promo_status!`

app/models/concerns/referral_promo.rb
```diff
@@ -0,0 +1,53 @@
...
+    return true unless referral_kyc_required?
+
+    # Otherwise they must be brave payable and from a valid country
+    brave_payable? && valid_promo_country?
+  end
+
+  # Public: Enqueues a job which allows publishers referrals to work if they are payable and in a valid promo_country
+  #
+  # Returns nil
+  def update_promo_status
```
yachtcaptain23, line 48: Nit: Change this to `update_promo_status!`

app/models/concerns/referral_promo.rb
```diff
@@ -0,0 +1,53 @@
...
+    return unless may_register_promo?
+
+    Promo::UpdateStatus.perform_later(id: id, status: PublisherStatusUpdate::ACTIVE)
```
nvonpentz, line 51: Does this just reenable the referral codes or does it also set the publisher's state to active?  It may not be applicable here, but I am worried about cases in which a suspended publisher can set themselves back to active by going through KYC.  Are there any edgecases like that here?

corymcdonald: Ah, yeah this is kind of confusing due to the `PublisherStatusUpdate` method. 😅  Fortunately this doesn't set a publisher's state back to active but rather it deletes the promo `owner_state` status from the promo server if the status is the string "active".

PR #2781, opened 2020-06-29
Only add SiteBannerLookups for verified channels only

yachtcaptain23:


app/models/channel.rb
```diff
@@ -274,7 +274,12 @@ def register_channel_for_promo
     Promo::RegisterChannelForPromoJob.perform_now(channel_id: id, attempt_count: 0)
   end
 
+  def site_channel?
```
corymcdonald, line 277: Where's this check being called?

yachtcaptain23: I'll remove.

PR #2817, opened 2020-07-15
Enable Gemini for all users

corymcdonald:
#### Features

For all new accounts gemini will be enabled


corymcdonald:
This was merged here https://github.com/brave-intl/publishers/commit/1603439581a60a8721dc068cfaea77c36da12783

PR #2821, opened 2020-07-21
Add recipient ID for Gemini Connections

corymcdonald:
#### Features

- Adds recipient id to the Gemini Connections

#### How To Test

Run `rails test`

~ Alternatively ~

You can connect your Publishers account to Gemini and then check the database by running 
`bundle exec rails console` 
One into the Rails console you can check all the recipient_ids
`GeminiConnection.all.map(&:recipient_id)`


app/models/publisher.rb
```diff
@@ -160,6 +160,20 @@ def wallet
...
+  # Public: Checks the different wallet connections and enqueues sync jobs to refresh their data
+  #         If their data wasn't refreshed in the last 2 hours.
+  #
+  # Returns nil
+  def sync_wallet_connections
+    if gemini_connection.present? && gemini_connection.updated_at > 2.hours.ago
+      Sync::Connection::GeminiConnectionSyncJob.perform_later(publisher_id: id)
+    end
+
+    if uphold_connection.present? && uphold_connection.updated_at > 2.hours.ago
```
yachtcaptain23, line 172:
```suggestion
    if uphold_connection.present? && uphold_connection.updated_at < 2.hours.ago
```

app/models/publisher.rb
```diff
@@ -160,6 +160,20 @@ def wallet
     @wallet ||= PublisherWalletGetter.new(publisher: self, include_transactions: false).perform
   end
 
+  # Public: Checks the different wallet connections and enqueues sync jobs to refresh their data
+  #         If their data wasn't refreshed in the last 2 hours.
+  #
+  # Returns nil
+  def sync_wallet_connections
+    if gemini_connection.present? && gemini_connection.updated_at > 2.hours.ago
```
yachtcaptain23, line 168:
```suggestion
    if gemini_connection.present? && gemini_connection.updated_at < 2.hours.ago
```

db/migrate/20200716170339_add_recipient_id_to_gemini_connection.rb
```diff
@@ -0,0 +1,8 @@
+class AddRecipientIdToGeminiConnection < ActiveRecord::Migration[6.0]
+  def change
+    add_column :gemini_connections, :recipient_id, :string, index: true, unique: true
```
yachtcaptain23, line 3: Did you want to set a default and allow or disallow null?

corymcdonald: I don't think so, if there's not a recipient_id created then this field should be null.

PR #2849, opened 2020-08-20
Add Gemini Default Currency

### LINKED ISSUE #2841: Add default currency to Gemini Connections
Add the ability to select default currency for Gemini users

corymcdonald:
## Add Gemini Default Currency

Closes #2841

#### Features

**♻️ Refactor** 
* Different provider connections (Uphold, Gemini, Stripe, Paypal) are now in the `Connections` Module. 
* Rails Routes have been refactored
    * New routes align with [one-level deep nesting standard](https://guides.rubyonrails.org/routing.html#limits-to-nesting).
    * Legacy OAuth routes have been mapped to new routes
    * Custom connection routes have been deleted to match REST conventions  

**🧽 Clean up**
* Deleted unused Uphold connection methods and helper functions
* Deleted CSS that is no longer needed.
* Deleted Javascript that no longer interfaced with homepage. 

✨ New
* Created [CurrencySelection](https://github.com/brave-intl/publishers/blob/add-default-currency/app/javascript/views/walletServices/braveConnection/CurrencySelection.tsx) React component
* Created hints for `WalletProvider` property in Publisher model
* Created [Gemini Default Currency Job](https://github.com/brave-intl/publishers/blob/add-default-currency/app/jobs/update_gemini_default_currency_job.rb) 
* Created [API](https://github.com/brave-intl/publishers/blob/add-default-currency/app/services/gemini/setting.rb) to interface with Gemini


app/controllers/connections/gemini_connections_controller.rb
```diff
@@ -1,28 +1,29 @@
 # frozen_string_literal: true
 
-module Publishers
+module Connections
```
yachtcaptain23, line 3: IMO Connections is borderline a "safeword" since it's related to database connections. Would prefer `PaymentConnections` instead

corymcdonald: https://github.com/brave-intl/publishers/pull/2849/commits/782a65bafbca333dc83ec84a154ae5eaba3d7763

PR #2857, opened 2020-08-27
Create wallet_provider on Publishers 

### LINKED ISSUE #2842: Create the current wallet_provider option on Publisher
We should allow users to only have one selected crypto payment option at the time of payout. 

Edge cases:
- When a publisher changes their default currency between the 1st- 8th.  (Perhaps we can include the payout type in the payout tracker)
- Paypal is considered a crypto payment.

corymcdonald:
## Create wallet_provider on Publishers 

Closes #2842

#### Features

✨ Creates a `wallet_provider_id` and `wallet_provider_kind` entry on the Publishers table.
✨ Now the code can call `publisher.wallet_provider` and get either a `GeminiConnection`, `UpholdConnection`, or `PaypalConnection`

#### How To Test

1. Connect Gemini
2. In the `rails console` type `Publisher.first.wallet_provider`
3. See the `GeminiConnection` is returned.


app/models/uphold_connection.rb
```diff
@@ -95,6 +95,7 @@ def disconnect_uphold
       default_currency_confirmed_at: nil,
       default_currency: nil,
     )
+    publisher.update(wallet_provider: nil)
```
yachtcaptain23, line 98: I think the `wallet_provider` should only be cleared if the uphold_connection is the selected wallet_provider

db/migrate/20200826231131_add_wallet_provider_to_publishers.rb
```diff
@@ -0,0 +1,5 @@
+class AddWalletProviderToPublishers < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :publishers, :wallet_provider, polymorphic: true, type: :uuid
```
yachtcaptain23, line 3: Nit picking: `selected_wallet_provider` or `preferred_wallet_provider` over `wallet_provider`

corymcdonald: https://github.com/brave-intl/publishers/pull/2857/commits/5243340640e464a6d87ed6a8aec1277e870721a5

app/models/uphold_connection.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2857/commits/5243340640e464a6d87ed6a8aec1277e870721a5

app/controllers/payment/connection/currencies_controller.rb
```diff
@@ -6,16 +6,16 @@ class CurrenciesController < ApplicationController
 
       def show
         render json: {
-          supported_currencies: current_publisher.wallet_provider.supported_currencies,
-          default_currency: current_publisher.wallet_provider.default_currency,
+          supported_currencies: current_publisher.selected_wallet_provider.supported_currencies,
```
yachtcaptain23, line 9: Will this crash if the selected_wallet_provider is nil?

corymcdonald: Yes, but that's intended with the changes to the default currency workflow. 👍

evq:
++

PR #2862, opened 2020-09-01
Update Scope and Gemini Verified Status

### LINKED ISSUE #2835: Gemini - Any verified status sets the account to be verified
For the case where there’re multiple users on the account we're taking the first account that’s returned from the endpoint for the name and verification status.

An account will be verified if any user on the account is verified (rather than taking the status of the first in the list). This is safe because if an unverified user tries to connect their account to Brave they will be able to but only read-only APIs will function (this mirrors the behavior of our UI).

### LINKED ISSUE #2836: Gemini - Update scopes
Creator Wallet
Recipient ids: payments:read
Get balances: balances:read
Account status: account:read

These are updated in QA now and will be in production Thurs eve. The prior scopes will still work so you can migrate over anytime after testing / our release. These should be added as a comma separated list, eg `scope=addresses:read,balances:read,orders:create`

corymcdonald:
## Update Scope and Gemini Verified Status

Closes #2836 
Closes #2835

### Features 

✨  De-duplicates some code with in the GeminiConnections controller. 
🐛 Fixes edge case where Gemini returns multiple users and only one `is_verified`

### DIFF
--- a/app/services/gemini.rb
+++ b/app/services/gemini.rb
@@ -1,5 +1,5 @@
 module Gemini
-  @scope = "account:read,balances:read,payments:send,payments:read"
+  @scope = "balances:read,account:read,payments:create"
   @response_type = "code"
 
   class << self
--- a/app/controllers/payment/connection/gemini_connections_controller.rb
+++ b/app/controllers/payment/connection/gemini_connections_controller.rb
@@ -26,25 +26,17 @@ module Payment
           code: params[:code],
           redirect_uri: publishers_gemini_connection_new_url(locale: nil)
         )
-        account = Gemini::Account.find(token: authorization.access_token)
-        user = account.users.first
-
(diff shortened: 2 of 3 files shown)


app/services/gemini.rb
```diff
@@ -1,5 +1,5 @@
 module Gemini
-  @scope = "account:read,balances:read,payments:send,payments:read"
+  @scope = "balances:read,account:read,payments:create"
```
evq, line 2: why is `payments:create` needed? i would think we need read permission to assemble a statement for the user in the future

corymcdonald: We need payments:create in order to generate new recipientIds, as well as make changes to default currency

evq (approved):
LGTM, scopes appear minimized

PR #2866, opened 2020-09-02
Release 2020-09-02

corymcdonald:
#### Features

:sparkles: Add checks for verified users.


corymcdonald:
Good idea, I've created an issue to tackle this. In the meantime we'll be testing this on staging and production so we can proceed with the monthly payouts

https://github.com/brave-intl/publishers/issues/2867

PR #2874, opened 2020-09-03
Do not automatically move publishers out of only_user_funds

corymcdonald:
#### Features

If an API request is made to change a user out of only_user_funds we ignore it.

### DIFF
--- a/app/controllers/api/v1/publishers_controller.rb
+++ b/app/controllers/api/v1/publishers_controller.rb
@@ -37,7 +37,7 @@ class Api::V1::PublishersController < Api::BaseController
 
     # We should not automatically move publishers out of this status.
     if user.only_user_funds?
-      render status: 200 and return
+      render(status: 200, json: { publisher_status_updates_id: user.last_status_update.id }) and return
     end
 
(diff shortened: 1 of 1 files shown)


app/controllers/api/v1/publishers_controller.rb
```diff
@@ -35,6 +35,11 @@ def publisher_status_updates
     status = params[:status]
     note = params[:note]
 
+    # We should not automatically move publishers out of this status.
+    if user.status == PublisherStatusUpdate::ONLY_USER_FUNDS
```
corymcdonald, line 39:
```suggestion
    if user. only_user_funds?
```

app/controllers/api/v1/publishers_controller.rb
```diff
@@ -35,6 +35,11 @@ def publisher_status_updates
     status = params[:status]
     note = params[:note]
 
+    # We should not automatically move publishers out of this status.
+    if user.only_user_funds?
```
evq, line 39: :gem: :raised_hands:

evq (approved):

PR #2882, opened 2020-09-09
Elasticsearch for admin dashboard

### LINKED ISSUE #2877: Improve searching on Publisher's backend
We should introduce Elasticsearch to improve the reliability for searching for Publishers.

corymcdonald:
## Elasticsearch for admin dashbaord

Closes #2877

#### Features

✨ Improves the search reliability

#### How To Test

1. [Download and install Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/7.7/targz.html#install-macos)
2. Run `rails search:setup` 
3. Run `rails indexer:user`


lib/tasks/indexer.rake
```diff
@@ -0,0 +1,13 @@
...
+  desc "create user indexes"
+  task user: :environment do
+    publishers = Publisher.select(:id)
+    count = publishers.size
+    puts "Indexing #{count}"
+    publishers.find_each.with_index do |publisher, index|
+      puts "#{index} / #{count}" if (index % 1000).zero?
+      Search::UserIndexJob.perform_async(publisher.id, queue: 'low')
+    end
+    puts 'Done'
```
nvonpentz, line 11: Single quotes!

app/models/publisher.rb
```diff
@@ -355,6 +358,19 @@ def country
     provider_country.to_s.upcase
   end
 
+  def index_to_elasticsearch
+    Search::UserIndexJob.perform_async(id)
+  end
+
+  def index_to_elasticsearch_now
+    return if ENV["ELASTICSEARCH_URL"].blank?
```
nvonpentz, line 366: Should you use `Rails.application.secrets`?

nvonpentz (commented):
I took a quick pass -- I'm still learning concepts here.

One thing I'm curious about is the strategy by which we keep the elastic search db up to date when new data is added or old data is updated.  I see that there are a few 'after commit' hooks that update elastic search like when a channel is verified.  Is that the primary mechanism? In general when can I be sure data is accurate?

corymcdonald:
@nvonpentz Yes, the ActiveRecord callbacks in the general strategy. My hope is to have as close to real time data in the cluster as possible. 

But I'm also trying to balance that with the number of jobs that will be kicked off. What i'm mostly concerned about is

* Anytime we save a publisher this can cause the `Search::UserIndexJob` to enqueue. (Could be a problem during our bulk jobs). 
* Anytime a channel is verified then this job is called

If you have any suggestions or ideas for maintaining consistency while preventing a large number of retry requests to enqueue I would love to hear.

lib/tasks/indexer.rake
corymcdonald: https://github.com/brave-intl/publishers/pull/2882/commits/048045801fcf1f38b8baebe78f56e9983b66b83a

app/models/publisher.rb
corymcdonald: https://github.com/brave-intl/publishers/pull/2882/commits/93683426af4a2aeae551d2a68c7ecc0bb9b023db

tsmartt:
Not needed since @yachtcaptain23 improved the performance of our search queries. Also a next step would be to use PG's full-text search for operational simplicity.

PR #2886, opened 2020-09-14
Revert "Set a Content-Security-Policy (#2580)"

corymcdonald:
This reverts commit 4fc569e2a741f9cf4c888e599901ccfb19c73ef3.

The CSP policy was breaking due to an outdated `styled-components` in Brave-UI. https://github.com/brave/brave-ui/issues/128

I will be taking this on when I have capacity, but until then we need to revert this commit.


evq (approved):
:cry:

PR #2891, opened 2020-09-23
Simplify deploy by using commit hash as digest

corymcdonald:
#### Features

* Remove the `yarn build` step. 

We don't technically need this right now. This was originally added so that if a change is made on the public site but "yarn build" isn't run then the latest changes will still be deployed.

* Reverse engineer the Rails Webpacker and Sprockets pipeline so that we can use an arbitrary digest.

### DIFF
--- a/public/service-worker.js
+++ b/public/service-worker.js
@@ -35,5 +35,5 @@ workbox.precaching.precacheAndRoute(self.__precacheManifest, {});
 
 workbox.routing.registerNavigationRoute(workbox.precaching.getCacheKeyForURL("/index.html"), {
   
-  blacklist: [/^\/_/,/\/[^/?]+\.[^/]+$/],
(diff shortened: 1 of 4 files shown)


public/service-worker.js
```diff
@@ -35,5 +35,5 @@ workbox.precaching.precacheAndRoute(self.__precacheManifest, {});
 
 workbox.routing.registerNavigationRoute(workbox.precaching.getCacheKeyForURL("/index.html"), {
   
-  blacklist: [/^\/_/,/\/[^/?]+\.[^/]+$/],
+  blacklist: [/^\/_/,/\/[^\/?]+\.[^\/]+$/],
```
corymcdonald, line 38: Not sure why this change, but it happened when I ran `yarn build` before deleting the entry in `package.json`.

evq (approved):

Issue #2921, opened 2020-10-15
Only enqueue 1 job per publisher when generating a payout report corresponding to their active wallet


nvonpentz:
Right now we have it such that we enqueue a new job for each wallet connection that each publisher has, then we abort some of those jobs later on so that we don't create duplicate payments going to different providers.

This has resulted in a bug in which people who have both Gemini and Uphold connections but have Uphold selected as the active wallet, will not be paid because we kill the Uphold jobs early here
https://github.com/brave-intl/publishers/blob/6596bfd9297a7f32463e0ebc5912d71aedafe7d5/app/services/payout/service.rb#L57

and then Gemini jobs fail because the active connection is not an Gemini connection, and thus doesn't have a recipient_id https://github.com/brave-intl/publishers/blob/ecce7ddfe2e783a6c21fff3b60d41c6ad2c1ff97/app/services/payout/gemini_service.rb#L10

Instead we should look at the active wallet provider at the time we enqueue the job and only enqueue a job for that wallet.

tsmartt:
Related to https://github.com/brave-intl/publishers/pull/3238

PR #2932, opened 2020-10-27
Kick off payout jobs based on selected wallet provider

nvonpentz:
Resolves https://github.com/brave-intl/publishers/issues/2921


tsmartt:
This is great, and I'll work on integrating it next week. We need to first close a bug where if a user disconnects a wallet, it sets their selected_waller to nil, even if the wallet _type_ was not the same.

dogweather:
@tsmartt Is there an Issue for that wallet bug?

tsmartt:
There is now https://github.com/brave-intl/publishers/issues/3223

tsmartt:
Closing as duplicate of https://github.com/brave-intl/publishers/pull/3238

PR #2952, opened 2020-11-11
Be more selective about which payout jobs to enqueue

nvonpentz:
For each provider payout job, only enqueue publishers that have that
wallet provider as their selected wallet provider, or have no wallet
provider selected.

### DIFF
--- a/app/jobs/payout/paypal_job.rb
+++ b/app/jobs/payout/paypal_job.rb
@@ -1,8 +1,18 @@
 class Payout::PaypalJob < ApplicationJob
   queue_as :scheduler
 
-  def perform(should_send_notifications: false, payout_report_id: nil)
-    publishers = Publisher.joins(:paypal_connection).with_verified_channel.where(paypal_connections: { country: "JP" })
+  def perform(should_send_notifications: false, payout_report_id: nil, publisher_ids: [])
+    if publisher_ids.present?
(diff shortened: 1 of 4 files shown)


app/jobs/payout/paypal_job.rb
```diff
@@ -1,8 +1,18 @@
 class Payout::PaypalJob < ApplicationJob
   queue_as :scheduler
 
-  def perform(should_send_notifications: false, payout_report_id: nil)
-    publishers = Publisher.joins(:paypal_connection).with_verified_channel.where(paypal_connections: { country: "JP" })
+  def perform(should_send_notifications: false, payout_report_id: nil, publisher_ids: [])
+    if publisher_ids.present?
+      publishers = Publisher.where(selected_wallet_provider_type: [nil, 'PaypalConnection']).
+        joins(:paypal_connection).
+        where(id: publisher_ids)
```
yachtcaptain23, line 8: Nit: Add ` where(paypal_connections: { country: "JP" })`

yachtcaptain23 (approved):
LG2M. Please add country to `Payout::PayPalJob` if you think it should be added

app/jobs/payout/paypal_job.rb
nvonpentz: Good catch

PR #2968, opened 2020-12-04
removed wallet get for payout

maikelmclauflin:
## Remove Eyeshade Account Check

#### Features

removes request to eyeshade for balances, which we will generate in subsequent payout steps

#### Pull Request Checklist

### DIFF
--- a/app/services/payout/uphold_service.rb
+++ b/app/services/payout/uphold_service.rb
@@ -9,10 +9,6 @@ module Payout
       potential_payments = []
       uphold_connection = @publisher.uphold_connection
 
-      wallet = PublisherWalletGetter.new(publisher: @publisher, include_transactions: false).perform
-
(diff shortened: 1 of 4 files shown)


app/services/payout/uphold_service.rb
```diff
@@ -44,15 +37,11 @@ def perform
...
-        fee_probi = wallet.channel_balances[channel.details.channel_identifier].fees_probi # fee = balance - probi
-        total_probi += probi
-
         potential_payments << PotentialPayment.new(
           payout_report_id: @payout_report&.id,
           name: "#{channel.publication_title}",
-          amount: "#{probi}",
-          fees: "#{fee_probi}",
+          amount: "0",
+          fees: "0",
```
nvonpentz, line 44: Maybe a null value here is better than 0.  We can achieve that by just deleting these lines and those columns won't be updated.

yachtcaptain23: There's a db restriction that discourages null values.

PR #2980, opened 2020-12-15
Add support for admins adding and removing whitelists

nvonpentz:
* Add publisher_whitelist_updates table

* Add PublisherWhitelistUpdates model, controller

* Expose UI to enable admins to use API

* Add test for PublisherWhitelistUpdatesController

## Pull Request Name
View status
image

Update status
image

#### Features

#### How To Test

1. Step one
2. Step two
3. Step three

#### Pull Request Checklist

### DIFF
--- a/app/controllers/admin/publishers/publisher_status_updates_controller.rb
+++ b/app/controllers/admin/publishers/publisher_status_updates_controller.rb
@@ -7,7 +7,7 @@ class Admin::Publishers::PublisherStatusUpdatesController < Admin::PublishersCon
 
   def create
     if @publisher.last_whitelist_update&.enabled && [PublisherStatusUpdate::NO_GRANTS, PublisherStatusUpdate::SUSPENDED].include?(status)
-      render(status: 403, json: {"reason": "Cannot suspend whitelisted publisher"}) and return
+      render(status: 403, json: { "reason": "Cannot suspend whitelisted publisher" }) and return
     end
 
     note = @publisher.notes.create(note: params[:note], created_by_id: current_publisher.id)
--- a/app/controllers/admin/publishers/publisher_status_updates_controller.rb
+++ b/app/controllers/admin/publishers/publisher_status_updates_controller.rb
@@ -6,6 +6,10 @@ class Admin::Publishers::PublisherStatusUpdatesController < Admin::PublishersCon
   end
 
   def create
+    if @publisher.last_whitelist_update&.enabled && [PublisherStatusUpdate::NO_GRANTS, PublisherStatusUpdate::SUSPENDED].include?(status)
+      render(status: 403, json: {"reason": "Cannot suspend whitelisted publisher"}) and return
+    end
+
     note = @publisher.notes.create(note: params[:note], created_by_id: current_publisher.id)
     @publisher.status_updates.create(status: params[:publisher_status], publisher_note: note)
     @publisher.reload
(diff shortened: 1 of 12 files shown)


app/controllers/admin/publishers/publisher_whitelist_updates_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Admin::Publishers::PublisherWhitelistUpdatesController < Admin::PublishersController
+  def create
+    note = @publisher.notes.create!(note: params[:note], created_by_id: current_publisher.id)
+    PublisherWhitelistUpdate.create!(publisher: @publisher, enabled: params[:enable], publisher_note: note)
```
1nezer0, line 4:
```suggestion
    PublisherWhitelistUpdate.create!(publisher: @publisher, enabled: params[:enable].present?, publisher_note: note)
```

db/migrate/20201215124546_add_publisher_whitelist_updates.rb
```diff
@@ -0,0 +1,10 @@
+class AddPublisherWhitelistUpdates < ActiveRecord::Migration[6.0]
+  def change
+    create_table :publisher_whitelist_updates, id: :uuid, default: -> { "uuid_generate_v4()"}, force: :cascade do |t|
+      t.belongs_to :publisher, type: :uuid, index: true, unique: true, null: false
+      t.references :publisher_note, type: :uuid
+      t.boolean :enabled, null: false
+      t.timestamps
```
1nezer0, line 7: Would you add indexes to the timestamps?

app/controllers/admin/publishers/publisher_whitelist_updates_controller.rb
nvonpentz: This param is probably poorly named.  If it's `false` then we disable the whitelist, if it's `true` then we enable the whitelist. There is a non null constraint in the database, so this should always be set.

db/migrate/20201215124546_add_publisher_whitelist_updates.rb
nvonpentz: Added

app/controllers/admin/publishers/publisher_status_updates_controller.rb
```diff
@@ -6,6 +6,10 @@ def index
   end
 
   def create
+    if @publisher.last_whitelist_update&.enabled && [PublisherStatusUpdate::NO_GRANTS, PublisherStatusUpdate::SUSPENDED].include?(status)
```
1nezer0, line 9: `status` should be defined

app/controllers/admin/publishers/publisher_whitelist_updates_controller.rb
1nezer0: `params[:enabled].present?` just handles the edge case of `params[:enabled]` being `nil` to be `false`

nvonpentz: I committed this and the test actually started failing.  Because what happens here is if we supply an explicit `false`, then `params[:enable].present?` will be `true` which isn't what we want

app/controllers/admin/publishers/publisher_status_updates_controller.rb
nvonpentz: Used params[:publisher_status] instead

yachtcaptain23 (approved):

PR #3020, opened 2021-02-01
bitFlyer integration

dlipeles:
Screen Shot 2021-01-11 at 7 59 55 PM

#### Adds Publishers infrastructure for bitFlyer wallet connections:

* OAuth integration with bitFlyer staging server
* Database migrations for bitFlyer connections on publishers
* Models/Controllers/Routes for bitFlyer connection creation
* Protobuf definitions for bitFlyer deposit id's on channels

#### brave TODO

 - [x] Add task definitions to Creators Staging
 - [ ] Merge to staging
 - [x] Create new deposit id per channel, rather than one for all channels
 - [ ] Build refresh connection flow using the bitFlyer connection's refresh token
 - [ ] Test channel tipping & new protobuf definitions with the brave browser client team

#### bitFlyer TODO
 - [x] Request new redirect urls on bitFlyer's staging server
 - [ ] Request bitFlyer provide account display name on authorization response
 - [x] Request bitFlyer provide endpoints to select / update default currency
 - [x] Request bitFlyer provide info if account is KYC'd / has 2FA on authorization response ( Unless my understanding is correct, that a bitFlyer account can _only_ be linked if it is already KYC'd and has 2FA )


app/javascript/views/walletServices/braveConnection/NotConnected.tsx
```diff
@@ -59,4 +61,21 @@ const GeminiConnectButton = () => (
   </a>
 );
 
+const BitflyerConnectButton = () => (
+  <a
+    className="btn btn-secondary font-weight-bold mb-2"
+    data-piwik-action="bitflyerConnectClicked"
```
maikelmclauflin, line 67: looks like all others are capitalized. should this be as well?

app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -72,13 +80,15 @@ def get_paypal_wallet_state(paypal_connection:)
...
   def cleanup!
-    begin
-      File.open(@temp_file.path, 'r') do |f|
-        File.delete(f)
-      end
-    rescue Errno::ENOENT
+    File.open(@temp_file.path, 'r') do |f|
+      File.delete(f)
     end
+  rescue Errno::ENOENT
```
maikelmclauflin, line 91: why was this moved?

app/services/bitflyer.rb
```diff
@@ -0,0 +1,14 @@
+module Bitflyer
+  @scope = "balances:read,account:read,payments:create"
```
maikelmclauflin, line 2: i don't know that these same scopes will be used for bitflyer

app/javascript/views/walletServices/braveConnection/NotConnected.tsx
dlipeles: thanks @maikelmclauflin you're totally right. Fixed

app/jobs/cache/browser_channels/responses_for_prefix.rb
dlipeles: Looks like linter optimized for this

app/controllers/payment/connection/bitflyer_connections_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+module Payment
+  module Connection
+    class BitflyerConnectionsController < ApplicationController
+      class BitflyerError < StandardError; end
+      before_action :authenticate_publisher!
+      before_action :validate_connection!, only: :new
+
+      def create
+        BitflyerConnection.find_or_create_by(publisher: current_publisher)
+        redirect_to "http://demo22oy5z2d2lu6pyoum26m7k.azurewebsites.net/ex/OAuth/authorize?client_id=6cd6f1a070afcd467e198c8039b2c97b&scope=assets+create_deposit_id+withdraw_to_deposit_id&redirect_uri=https%3A%2F%2Flocalhost%3A3000%2Fpublisher%2Fbitflyer_connection%2Fnew&state=100&response_type=code"
```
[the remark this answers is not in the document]
dlipeles: Definitely @maikelmclauflin was planning on cleaning this one up, these are just temp staging credentials and will be destroyed after testing, now they are replaced with Rails secrets too

zenparsing (approved):
Changes to `channel_responses.proto` look good.

app/jobs/update_bitflyer_default_currency_job.rb
```diff
@@ -0,0 +1,23 @@
+class UpdateBitflyerDefaultCurrencyJob
```
yachtcaptain23, line 1: Unused?

app/models/bitflyer_connection.rb
```diff
@@ -0,0 +1,107 @@
+# frozen_string_literal: true
+
+class BitflyerConnection < ApplicationRecord
+  SUPPORTED_CURRENCIES = ["BAT", "USD", "BTC", "ETH"].freeze
+
+  belongs_to :publisher
+
+  validates :recipient_id, uniqueness: true, allow_blank: true
```
yachtcaptain23, line 8: Would you be able to group these together based on the order of operations? Also, I don't think we need so much whitespace lol

app/models/bitflyer_connection.rb
```diff
@@ -0,0 +1,107 @@
...
+      access_token: authorization.access_token,
+      refresh_token: authorization.refresh_token,
+      expires_in: authorization.expires_in,
+      access_expiration_time: authorization.expires_in.seconds.from_now
+    )
+    # Reload the model so consumers will have the most up to date information.
+    reload
+  end
+
+  def sync_connection!
```
yachtcaptain23, line 64: Where is this called?

Procfile.dev
```diff
@@ -0,0 +1,3 @@
+web: bundle exec puma -C config/puma.rb -e ${RACK_ENV:-development}
```
yachtcaptain23, line 1: Nit: Prefer removing this file

app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -72,13 +80,15 @@ def get_paypal_wallet_state(paypal_connection:)
     end
   end
 
+  def get_bitflyer_wallet_state(bitflyer_connection:)
+    PublishersPb::BitflyerWalletState::BITFLYER_ACCOUNT_KYC
```
yachtcaptain23, line 84: Just to confirm, we're assuming that every BitFlyer who registers with us is KYC'd already, correct?

app/models/bitflyer_connection.rb
```diff
@@ -0,0 +1,107 @@
...
+  def prepare_state_token!
+    update(state_token: SecureRandom.hex(64).to_s)
+  end
+
+  def payable?
+    true
+  end
+
+  def japanese_account?
+    country&.upcase == 'JP'
```
yachtcaptain23, line 27: Nit: Make this a constant at the top of the file, e.g. 

```
  JAPAN = "JP"
```
And do `country&.upcase == JAPAN`

app/models/bitflyer_connection.rb
```diff
@@ -0,0 +1,107 @@
...
+
+  validates :recipient_id, uniqueness: true, allow_blank: true
+
+  attr_encrypted :access_token, :refresh_token, key: :encryption_key
+
+  after_save :update_default_currency, if: -> { saved_change_to_default_currency? }
+
+  validates :default_currency, inclusion: { in: SUPPORTED_CURRENCIES }, allow_nil: true
+
+  after_destroy :selected_wallet_provider
```
yachtcaptain23, line 16: Hmm, not completely certain but I think we want to use a `before_destroy` (I don't know if calling `.destroy` will wipe the `id` field)

db/migrate/20210131233252_add_bitflyer_deposit_id_to_publishers.rb
```diff
@@ -0,0 +1,5 @@
+class AddBitflyerDepositIdToPublishers < ActiveRecord::Migration[6.0]
+  def change
+    add_column :publishers, :bitflyer_deposit_id, :string
```
yachtcaptain23, line 3: Instead of creating a new column, can we use a table like UpholdConnections that stores deposit id and the publisher id? I don't like adding more columns to already bloated tables.

app/jobs/update_bitflyer_default_currency_job.rb
dlipeles: Fixed, turns out bitFlyer will use only "BAT" as default currency here

app/models/bitflyer_connection.rb
dlipeles: For sure, fixed

app/models/bitflyer_connection.rb
dlipeles: I'll leave this in the code for now, we may need to implement a refresh flow at some point, but not needed for a staging test

Procfile.dev
dlipeles: Fixed, added to gitignore

app/jobs/cache/browser_channels/responses_for_prefix.rb
dlipeles: Yes this is correct, confirmed this with Chris today

app/models/bitflyer_connection.rb
dlipeles: Fixed

db/migrate/20210131233252_add_bitflyer_deposit_id_to_publishers.rb
dlipeles: You're totally right @yachtcaptain23 , Fixed, moved this field out of the publishers table

db/migrate/20210112182218_create_bitflyer_connection.rb
```diff
@@ -0,0 +1,27 @@
...
+      t.string :encrypted_refresh_token_iv, index: { unique: true }
+
+      t.string :expires_in
+      t.datetime :access_expiration_time
+      t.string :display_name
+      t.string :state_token
+
+      t.string :scope
+
+      t.string :status, index: true
```
yachtcaptain23, line 19: Nit: Is there a query that's heavily dependent on `status`? If not, I don't think we should index this.

app/models/channel.rb
```diff
@@ -335,6 +337,22 @@ def notify_slack
     ).perform
   end
 
+  # Needed for bitFlyer, but can likely be used for Uphold too.
+  def create_deposit_id
```
yachtcaptain23, line 341: I'd rather we make this a JSONB column that says `deposit_setting` and one of the values can be `deposit_id`. Otherwise, we have an existing pattern mentioned in `app/models/uphold_connection_for_channel.rb` that I prefer we copy.

db/migrate/20210112182218_create_bitflyer_connection.rb
```diff
@@ -0,0 +1,27 @@
+class CreateBitflyerConnection < ActiveRecord::Migration[6.0]
+  def change
+    create_table :bitflyer_connections, id: :uuid, default: -> { "uuid_generate_v4()"}, force: :cascade do |t|
+      t.belongs_to :publisher, type: :uuid, index: true, null: false
```
yachtcaptain23, line 4: Would you add a uniqueness index? It will prevent duplicate saves

app/services/bitflyer/auth.rb
```diff
@@ -0,0 +1,90 @@
...
+        grant_type: REFRESH_TOKEN,
+        refresh_token: token,
+      }
+      response = post(PATH.expand(segments: 'token'), body)
+
+      Auth.new(JSON.parse(response.body))
+    end
+
+    def api_base_uri
+      Bitflyer.oauth_uri
```
yachtcaptain23, line 87: Nit: if there's no difference between `api_base_uri` and `oauth_uri`, I think we should settle on using one of the two.

PR #3033, opened 2021-02-10
Partial fix before full deprecation of the 2018q1 flag

yachtcaptain23:
The `promo_enabled_2018q1` column in `publishers` was being used to turn on/off promo registrations for users. We changed the default behavior so that this flag is false, but we're essentially causing an erroneous double feature-flag for users. 

In a later PR, a migration will be used to remove the column in `publishers`.

### DIFF
--- a/app/views/admin/publishers/_wallet_info.html.slim
+++ b/app/views/admin/publishers/_wallet_info.html.slim
@@ -33,7 +33,7 @@
       .db-field = "Created by:"
       .db-value = link_to @publisher.created_by, admin_publisher_path(@publisher.created_by)
   - ActiveRecord::Base.connected_to(role: :reading)
-    - if @publisher.promo_enabled_2018q1
+    - if @publisher.feature_flags[UserFeatureFlags::REFERRAL_KYC_REQUIRED]
       .db-info-row
         .db-field = "Referral downloads:"
         .db-value = publisher_referral_totals(@publisher)[PromoRegistration::RETRIEVALS]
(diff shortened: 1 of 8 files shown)


app/views/admin/publishers/_wallet_info.html.slim
```diff
@@ -33,7 +33,7 @@
       .db-field = "Created by:"
       .db-value = link_to @publisher.created_by, admin_publisher_path(@publisher.created_by)
   - ActiveRecord::Base.connected_to(role: :reading)
-    - if @publisher.promo_enabled_2018q1
+    - if @publisher.feature_flags[UserFeatureFlags::REFERRAL_KYC_REQUIRED]
```
yachtcaptain23, line 36:
```suggestion
    - if @publisher.may_create_referrals?
```

test/controllers/publishers/promo_registrations_controller_test.rb
```diff
@@ -2,6 +2,10 @@
 require "webmock/minitest"
 require "shared/mailer_test_helper"
 
+# (Albert Wang): Because we disabled referrals, users can't create referrals for themselves
+# The controller's create function only runs in tests and in development.
+# Although it's unlikely, we retain this in case we need to turn back on the referral program.
```
tsmartt, line 7: Just delete it. It will clean up the codebase and we can always restore from git history if we need to. We can delete the associated test as well, and this comment :)

tsmartt (approved):
Looks good, thanks!

PR #3055, opened 2021-02-22
Query from redshift for confirmations

yachtcaptain23:


yachtcaptain23:
Going to handle this in a separate PR

Issue #3082, opened 2021-03-04
webtest-pj.herokuapp.com shows up as unverified on staging pCDN


jumde:
@dlipeles confirmed that: `webtest-pj.herokuapp.com` is correctly verified and Uphold looks good too. This is not picked up correctly by pCDN

![image](https://user-images.githubusercontent.com/1903815/110021084-2a0d5480-7cdf-11eb-8b80-3f8bf63bfa2b.png)

![image](https://user-images.githubusercontent.com/1903815/110021104-2f6a9f00-7cdf-11eb-9e40-53b2054cf702.png)

yachtcaptain23:
The account is marked as a BitFlyer account, chasing down as to whether or not this is supposed to happen.

PR #3086, opened 2021-03-07
Fix scenario in which an exception would prevent the channel response…

yachtcaptain23:
… from being added to the channel_responses array.

Closes https://github.com/brave-intl/publishers/issues/3082


dlipeles (approved):

app/jobs/channels/approve_channel_transfer_job.rb
```diff
@@ -11,7 +11,7 @@ def perform(channel_id:)
         ## There is a helper function in channel.rb that can help minimize this code
         ## Need to generate new bF deposit ID on channel transfer:
         channel.deposit_id = nil
-        if contested_by.publisher.selected_wallet_provider_type == BITFLYER_CONNECTION
+        if contested_by.publisher.selected_wallet_provider_type == Channel::BITFLYER_CONNECTION
```
dlipeles, line 14: quick q, does Channel::BITFLYER_CONNECTION return 'BitflyerConnection'

app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -38,18 +38,17 @@ def generate_brotli_encoded_channel_response(prefix:)
           wallet.paypal_wallet = paypal_wallet
           channel_response.wallets.push(wallet)
         end
-        if site_banner_lookup.publisher.bitflyer_connection.present?
+        if site_banner_lookup.publisher&.bitflyer_connection.present?
```
dlipeles, line 41: I used to have this line here, which might help too 

site_banner_lookup.publisher.selected_wallet_provider_type == BITFLYER_CONNECTION

if we wanted to try that

app/jobs/channels/approve_channel_transfer_job.rb
yachtcaptain23: Yeah it does. 

```
  BITFLYER_CONNECTION = "BitflyerConnection".freeze
```

app/jobs/cache/browser_channels/responses_for_prefix.rb
yachtcaptain23: I'm reverting, this change wasn't necessary

Issue #3114, opened 2021-03-19
When Creator presses "remove channel", actually remove and unverify the channel from the PCDN


Miyayes:
## Description

When Creator presses "remove channel", actually remove and unverify the channel from the PCDN.

PR #3115, opened 2021-03-19
After a channel transfer, a job gets enqueued to update the publisher list

yachtcaptain23:
Closes https://github.com/brave-intl/publishers/issues/3114

### DIFF
--- a/app/jobs/cache/browser_channels/responses_for_prefix.rb
+++ b/app/jobs/cache/browser_channels/responses_for_prefix.rb
@@ -14,6 +14,7 @@ class Cache::BrowserChannels::ResponsesForPrefix
     pad_file!
     save_to_s3!(prefix: prefix) unless Rails.env.test?
     cleanup!
+    refresh_cdn!
   end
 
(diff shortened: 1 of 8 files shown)


app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -95,6 +96,10 @@ def cleanup!
   rescue Errno::ENOENT
   end
 
+  def refresh_cdn!
+    Faraday.get(ENV["BRAVE_PCDN_URL"], {}, { "Cache-Control" => "no-cache"})
```
bridiver, line 100: This will only clear the cache for whatever edge node it happens to hit. Presumably the CDN has a way to actually invalidate the entire cache

yachtcaptain23: Going to revert this and take it up with devops to see what we can do for edge node invalidation.

dlipeles (approved):

PR #3122, opened 2021-03-23
Don't change the global value of the default_locale. This affects other user's sessions

yachtcaptain23:


dlipeles:
I think the locale issues have already been resolved according to bitFlyer, I'll leave this in for time being as it's required for the bF launch , if we see more reports of locale after the launch I will assign myself a P1 to find a better solution

PR #3154, opened 2021-04-01
Perf improvements for reading the database

yachtcaptain23:
Details of how the separate write and read databases should be properly handled is described in https://github.com/rails/rails/issues/38548

Given that the jobs are consuming what seem to be the primary database's resources, this PR properly moves the work to the replica.


tsmartt:
I'll test this out locally, what method did you use?

tsmartt (approved):

PR #3157, opened 2021-04-01
Don't run the prefix refresh constantly. Should only happen every 24 …

yachtcaptain23:
…hours

### DIFF
--- a/app/jobs/cache/browser_channels/main.rb
+++ b/app/jobs/cache/browser_channels/main.rb
@@ -28,7 +28,7 @@ class Cache::BrowserChannels::Main
 
   def full_refresh_not_ran_recently?(all_prefixes_run_time:)
     return true if all_prefixes_run_time.nil?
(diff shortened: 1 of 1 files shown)


app/jobs/cache/browser_channels/main.rb
```diff
@@ -28,7 +28,7 @@ def perform
 
   def full_refresh_not_ran_recently?(all_prefixes_run_time:)
```
dlipeles, line 29: may also consider calling this full_refresh_ran_recently

dlipeles (approved):

app/jobs/cache/browser_channels/main.rb
yachtcaptain23: haha I think ruby should implement `not` similar to ! but it's different from python from what I remember: https://stackoverflow.com/questions/27429711/difference-between-not-and-in-ruby

PR #3159, opened 2021-04-01
Address a scenario where system time might not have been initialized.

yachtcaptain23:
There was an incident in which
(1) some users who aren't supposed to see the ref panel were able to see the ref panel
(2) some users who are supposed to see the ref panel weren't able to see
the ref panel

Unless the code started bitflipping, the other possibility is that
during startup, the system time never got initialized for the pod.

### DIFF
--- a/app/controllers/health_checks_controller.rb
+++ b/app/controllers/health_checks_controller.rb
@@ -16,7 +16,12 @@ class HealthChecksController < ActionController::Base
       { name: "cache", healthy: cache_connected? },
       { name: "database", healthy: database_connected? },
       { name: "mailer_queue", healthy: mailer_queue_low? },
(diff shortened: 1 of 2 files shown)


tsmartt:
Thanks @yachtcaptain23! Can we separate this PR into two, one with the health checks and another with the model change? I'd like to not mix these. I think the model change is not the issue here, though I agree the implicit conversion is a code smell. When we do fix it, it what do you think of using [Date.parse](https://ruby-doc.org/stdlib-2.6.1/libdoc/date/rdoc/Date.html#method-c-parse) with an explicit format specified? I think that would aid in code readability and explicitly fail if the format changes.

yachtcaptain23:
Root cause found. This is not the fix. Closing.

PR #3185, opened 2021-04-20
Implement Bitflyer Refresh Token Support

tsmartt:
We can now refresh Bitflyer Oauth tokens. This has been taken out of the model and a new service class has been created using the builder pattern. A few tests have been added. Note that for the Auth module, I've also used a DI approach to make testing easier. This allows us to keep using the base_api_client while we migrate the clients to a DI approach.


app/models/bitflyer_connection.rb
```diff
@@ -38,24 +38,6 @@ def access_token_expired?
     access_expiration_time.present? && Time.now > access_expiration_time
   end
 
-  # Makes a request to the Bitflyer API to refresh the current access_token
-  def refresh_authorization!
```
yachtcaptain23, line 42: You'll need to update where this is called too

tsmartt: Yes, so this currently isn't called -- except for in `sync_connection!`, which isn't called at all. It is commented out in the Payout task and will need to be enabled there. The goal is to do that work later this week, but that will require pulling that function into another service class and out of the model -- something not in scope for this PR.

dlipeles (approved):

dlipeles:
This is awesome @tsmartt

PR #3186, opened 2021-04-21
Release 04-21-2021

tsmartt:
This release adds the ability to refresh Bitflyer Oauth tokens. I've refreshed the tokens on staging, now I need to do the same in prod.


tsmartt:
Note that failing test on staging is an intermittent test that I need to fix, so I made a ticket for it https://github.com/brave-intl/publishers/issues/3187

yachtcaptain23 (changes requested):
Comments in line

config/secrets.yml
```diff
@@ -209,6 +209,11 @@ test:
   sendgrid_marketing_list_id: 10300648
   sendgrid_publishers_list_id: 3986776
   slack_webhook_url: false
+  # Bitflyer env variables
+  bitflyer_host: "https:://localhost/bitflyertest"
```
yachtcaptain23, line 213: Duplicate secrets and keys, it should read from the environment

tsmartt: Not in the test environment, I don't to take anything from the environment here. In fact, now, that you mention it, we should remove all ENV settings from this test section so I'll make a ticket.

tsmartt: https://github.com/brave-intl/publishers/issues/3189

yachtcaptain23 (approved):

Issue #3193, opened 2021-04-23
Login link says expired if ?locale=ja, but works if ?locale=en


Miyayes:
## Description

Creator receives login link, but the login link always says it's expired if the `?locale=ja`. However, when the locale is manually changed to `en` in the URL, it works.

PR #3200, opened 2021-04-27
Fix a double authentication issue due to a redirect on the before_action authentication_via_token

yachtcaptain23:
Fix a double authentication issue due to a redirect on the before_action authentication_via_token.

The solution in place is to use a separate before_action to set the
params in the locale.

Refactored the locale logic in the ApplicationController a bit to make
it cleaner.

Closes https://github.com/brave-intl/publishers/issues/3193

#### How To Test

1. Sign-in with an existing user, make sure page loads
2. Sign-in with an existing Japanese user, make sure page loads in Japanese
3. Sign-in with a new Japanese user, make sure Japanese locale params are apparent and the page loads in Japanese

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -24,29 +24,26 @@ class ApplicationController < ActionController::Base
   around_action :switch_locale
 
   def switch_locale(&action)
-    locale = nil
     return I18n.with_locale(I18n.default_locale, &action) if controller_path&.split("/")&.first == 'admin'
 
-    locale = params[:locale] if params[:locale].present?
-    if (japanese_header?(locale) || has_paypal_account?(locale) || controller_path.include?("bitflyer")) && request.get?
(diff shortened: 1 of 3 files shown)


tsmartt:
With this change, I am getting a BAT user showing BAP. Also it seems to default the homepage to jabap, which it should not.

Screen Shot 2021-04-28 at 11 42 02 AM
Screen Shot 2021-04-28 at 11 44 19 AM

tsmartt:
I also noticed 2 emails being sent, 1 for en, 1 for ja.

tsmartt:
OK perhaps it was just a fluke, I only got 1 email this time. 

I noticed that when bitflyer is not enabled, the popup is in English, but the background is in Japanese.

Screen Shot 2021-04-28 at 2 50 32 PM

But this happens on staging too, so it appears to be a bug I need to fix.

tsmartt (approved):

PR #3205, opened 2021-04-29
Enable Bitflyer potential payments

tsmartt:
We can now generate potential payments for Bitflyer creators. Note that I tried to clean up a little of the calling code,
as it appears we no longer need should_send_notifications. Also the class is now easily testable without using web mocks
since we can pass in the refresher.

### DIFF
--- a/app/jobs/include_publisher_in_payout_report_job.rb
+++ b/app/jobs/include_publisher_in_payout_report_job.rb
@@ -26,11 +26,6 @@ class IncludePublisherInPayoutReportJob
     potential_payment_job = nil
     publisher = Publisher.find(publisher_id)
 
-    attrs_payment_job = {
-      publisher: publisher,
-      payout_report: payout_report,
-    }
-
     case arguments[:kind].to_sym
     when GEMINI
       potential_payment_job = Payout::GeminiService
@@ -39,11 +34,13 @@ class IncludePublisherInPayoutReportJob
     when UPHOLD
       potential_payment_job = Payout::UpholdService
     when BITFLYER
-      return Payout::BitflyerService.build.perform(attrs_payment_job)
+      return Payout::BitflyerService.build.perform(publisher: publisher,
+                                                   payout_report: payout_report)
     when MANUAL
       potential_payment_job = Payout::ManualPayoutReportPublisherIncluder
     end
 
-    potential_payment_job.new(attrs_payment_job).perform
+    potential_payment_job.new(publisher: publisher,
+                              payout_report: payout_report).perform
   end
 end
--- a/app/jobs/include_publisher_in_payout_report_job.rb
+++ b/app/jobs/include_publisher_in_payout_report_job.rb
@@ -16,7 +16,6 @@ class IncludePublisherInPayoutReportJob
     # not create payments
     payout_report_id = arguments[:payout_report_id]
     publisher_id = arguments[:publisher_id]
-    should_send_notifications = arguments[:should_send_notifications]
 
     if payout_report_id.present?
       payout_report = PayoutReport.find(payout_report_id)
@@ -27,6 +26,11 @@ class IncludePublisherInPayoutReportJob
     potential_payment_job = nil
     publisher = Publisher.find(publisher_id)
 
+    attrs_payment_job = {
+      publisher: publisher,
+      payout_report: payout_report,
+    }
+
     case arguments[:kind].to_sym
     when GEMINI
       potential_payment_job = Payout::GeminiService
@@ -35,15 +39,11 @@ class IncludePublisherInPayoutReportJob
     when UPHOLD
       potential_payment_job = Payout::UpholdService
     when BITFLYER
-      potential_payment_job = Payout::BitflyerService
+      return Payout::BitflyerService.build.perform(attrs_payment_job)
     when MANUAL
       potential_payment_job = Payout::ManualPayoutReportPublisherIncluder
     end
 
-    potential_payment_job.new(
-      publisher: publisher,
-      payout_report: payout_report,
-      should_send_notifications: should_send_notifications,
-    ).perform
+    potential_payment_job.new(attrs_payment_job).perform
   end
 end
--- a/app/jobs/include_publisher_in_payout_report_job.rb
+++ b/app/jobs/include_publisher_in_payout_report_job.rb
@@ -16,6 +16,7 @@ class IncludePublisherInPayoutReportJob
     # not create payments
     payout_report_id = arguments[:payout_report_id]
     publisher_id = arguments[:publisher_id]
+    should_send_notifications = arguments[:should_send_notifications]
 
     if payout_report_id.present?
       payout_report = PayoutReport.find(payout_report_id)
@@ -41,6 +42,7 @@ class IncludePublisherInPayoutReportJob
     end
 
     potential_payment_job.new(publisher: publisher,
-                              payout_report: payout_report).perform
+                              payout_report: payout_report,
+                              should_send_notifications: should_send_notifications).perform
   end
 end
--- a/app/jobs/payout/paypal_job.rb
+++ b/app/jobs/payout/paypal_job.rb
@@ -17,6 +17,7 @@ class Payout::PaypalJob < ApplicationJob
       IncludePublisherInPayoutReportJob.perform_async(
         payout_report_id: payout_report_id,
         publisher_id: publisher.id,
+        should_send_notifications: should_send_notifications,
         kind: IncludePublisherInPayoutReportJob::PAYPAL
       )
     end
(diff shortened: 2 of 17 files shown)


app/services/payout/bitflyer_service.rb
```diff
@@ -1,58 +1,69 @@
...
       potential_payments = []
 
-      connection = @publisher.bitflyer_connection
+      connection = publisher.bitflyer_connection
       # Sync the connection
-      # connection.sync_connection!
+      @refresher.call(bitflyer_connection: connection)
 
       potential_payments << PotentialPayment.new(
-        payout_report_id: @payout_report&.id,
```
tsmartt, line 15: @nvonpentz This is the section you are most likely interested in as it details what you'll end up seeing.

app/jobs/include_publisher_in_payout_report_job.rb
```diff
@@ -35,15 +39,11 @@ def perform(arguments = {})
...
     when MANUAL
       potential_payment_job = Payout::ManualPayoutReportPublisherIncluder
     end
 
-    potential_payment_job.new(
-      publisher: publisher,
-      payout_report: payout_report,
-      should_send_notifications: should_send_notifications,
-    ).perform
+    potential_payment_job.new(attrs_payment_job).perform
```
yachtcaptain23, line 47: Would you restore using keyword arguments? That way the caller knows what's going on

tsmartt: Sure, I'll put it back. I used a var since I need to use it twice, but if you prefer the readability of the other, I'm not married to it.

tsmartt:
@yachtcaptain23 we should merge https://github.com/brave-intl/publishers/pull/3206 first and then I'll rebase this one.

test/services/payout/bitflyer_service_test.rb
```diff
@@ -0,0 +1,59 @@
...
+      subject
+    end
+
+    it 'creates potential payments for the right creator' do
+      refute_equal PotentialPayment.count, 0
+      PotentialPayment.all.each { |pp| assert_equal publisher.id, pp.publisher_id }
+    end
+
+    it 'the address is not empty' do
+      assert PotentialPayment.all.all? { |pp| pp.address.present? }
```
yachtcaptain23, line 56:
```suggestion
      PotentialPayment.all.each { |potential_payment| assert_equal publisher.id, potential_payment.publisher_id }
    end

    it 'the address is not empty' do
      assert PotentialPayment.all.all? { |potential_payment| potential_payment.address.present? }
```

yachtcaptain23: Would you be able to check to make sure the address matches rather than just if it's present?

app/jobs/payout/paypal_job.rb
```diff
@@ -17,7 +17,6 @@ def perform(should_send_notifications: false, payout_report_id: nil, publisher_i
       IncludePublisherInPayoutReportJob.perform_async(
         payout_report_id: payout_report_id,
         publisher_id: publisher.id,
-        should_send_notifications: should_send_notifications,
```
yachtcaptain23, line 20: If you're going to remove, would you remove it everywhere in the codebase? The particular call-chain can be found with `should_only_notify`.

app/services/payout/bitflyer_service.rb
```diff
@@ -1,58 +1,69 @@
...
+      @refresher = refresher
+      @payout_utils_class = payout_utils_class
+    end
+
+    # Change to call as soon as we refactor the other services
+    def perform(publisher:, payout_report:)
+      payout_utils = @payout_utils_class.new(payout_report: payout_report,
+                                             publisher: publisher,)
+
+      return if payout_utils.skip_publisher?
```
yachtcaptain23, line 17: I believe you need to update `Payout::Service`

app/jobs/include_publisher_in_payout_report_job.rb
```diff
@@ -35,15 +35,14 @@ def perform(arguments = {})
     when UPHOLD
       potential_payment_job = Payout::UpholdService
     when BITFLYER
-      potential_payment_job = Payout::BitflyerService
+      return Payout::BitflyerService.build.perform(publisher: publisher,
```
yachtcaptain23, line 38: IMO I'd rather have all code follow one convention rather than going half-way and implementing a new pattern and any maintainer has to maintain more mental models than necessary. Furthermore, I don't see the benefit of the builder pattern when we want all code to use an inheritance pattern.

yachtcaptain23: Discussed offline, we'll refactor this later.

test/services/payout/bitflyer_service_test.rb
tsmartt: Sure.

app/jobs/payout/paypal_job.rb
tsmartt: I put it back, we should handle all this in a separate PR.

test/services/payout/bitflyer_service_test.rb
tsmartt: Done.

app/services/payout/bitflyer_service.rb
tsmartt: Done. A bunch of code was removed due to being stale.

app/services/payout/bitflyer_service.rb
```diff
@@ -1,58 +1,69 @@
...
+
+    def initialize(refresher:, payout_utils_class:)
+      @refresher = refresher
+      @payout_utils_class = payout_utils_class
+    end
+
+    # Change to call as soon as we refactor the other services
+    def perform(publisher:, payout_report:)
+      payout_utils = @payout_utils_class.new(payout_report: payout_report,
+                                             publisher: publisher,)
```
yachtcaptain23, line 15:
```suggestion
      @payout_utils_class = payout_utils_class.new(payout_report: payout_report,
                                             publisher: publisher,)
    end

    # Change to call as soon as we refactor the other services
    def perform(publisher:, payout_report:)
```

app/services/payout/bitflyer_service.rb
```diff
@@ -1,58 +1,69 @@
...
-          status: @publisher.last_status_update&.status,
+          suspended: publisher.suspended?,
+          status: publisher.last_status_update&.status,
           channel_stats: channel.details.stats,
           channel_type: channel.details_type
         )
       end
 
-      unless should_only_notify?
+      unless payout_utils.should_only_notify?
```
yachtcaptain23, line 59: If you take the change above, then this part will need to be refactored too.

app/services/payout/service.rb
```diff
@@ -51,31 +51,6 @@ def skip_publisher?
...
-          create_message("Publisher is located in Japan and is not paid out through this job")
-          return true
-        end
-
-        if @publisher.paypal_connection.present?
-          create_message("Publisher has a Paypal Connection  and is not paid out through this job")
-          return true
-        end
-      end
-
```
yachtcaptain23, line 78: Why were these deleted? I believe we need to retain this code but also include BitFlyer to include Japanese payments.

test/services/payout/uphold_service_test.rb
```diff
@@ -247,31 +247,13 @@ class UpholdServiceTest < ActiveJob::TestCase
     end
   end
 
-  describe "for a paypal user" do
```
yachtcaptain23, line 250: Why is this test being deleted?

app/services/payout/bitflyer_service.rb
tsmartt: This would require passing in the `publisher` and `payout_report` to `build`, which should only be used for non-dynamic utility/third party modules. As the surrounding code is modularized, the utils class will no longer need to be instantiated and those methods can become `self`, effectively turning it into a module of functions.

app/services/payout/service.rb
tsmartt: I believe these are irrelevant for 2 reasons: 
1) Paypal is going away, so we can clean up this code but more importantly 
2) As the comment suggested, this can be removed once that issue is resolved, which it was. We pay out based on the `selected_wallet_provider`, so these checks are unnecessary.

tsmartt: I'm going to separate out the data cleanup into a separate PR, so I'll restore these for now.

yachtcaptain23 (approved):
LG2M. Please remove running the paypal payouts in either this PR or in a separate PR

tsmartt:
>  Please remove running the paypal payouts in either this PR or in a separate PR

Will do. Was planning a separate PR.

PR #3207, opened 2021-04-30
Add dotenv gem to ease development secret management

tsmartt:
Add dotenv gem to ease development secret management.

This simplifies development `.env` file management. We no longer have to deal with bash scripts as the variables will be loaded automatically and don't pollute the shell environment.


tsmartt:
@evq added you since @yachtcaptain23 mentioned you'd be interested in new gem additions.

Gemfile
```diff
@@ -199,6 +199,8 @@ group :development, :test do
   gem "chromedriver-helper"
   gem 'simplecov', require: false, group: :test
   gem 'selenium-webdriver', "~> 3.142.0"
+
+  gem 'dotenv-rails'
```
yachtcaptain23, line 203: Would you add a version to the Gemfile? If i recall correctly, the security team wants hardcoded versions to prevent an accidental `bundle update` from introducing a security flaw.

tsmartt:
@yachtcaptain23 should be good to go.

yachtcaptain23 (approved):

PR #3209, opened 2021-04-30
Add CONTRIBUTING.md

tsmartt:
This is meant to be a set of guidelines for anyone contributing to follow, and this PR is meant to spark constructive criticism.
Please feel free to propose additions where any might be lacking, and modifications with good justification.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,34 @@
+# Contributing
+
+Pull requests are welcome from anyone, please feel free to contribute by following the below guidelines.
+
+Fork, then clone the repo:
+
+    git clone git@github.com:your-username/publishers.git
+
+Write your code and tests, then make sure the tests pass:
+
+    bin/rake test
+
+Push to your fork and [submit a pull request][pr].
+
+[pr]: https://github.com/brave-intl/publishers/compare/
+
+We like to make sure that all contributions follow a few guidelines, which will increase the likelihood of a merge:
+
+- Follow the [code review guildelines][codereview]
+- Write tests to cover critical functionality
+- Write a [good commit message][commit]
+- Leave code better than you found it, whenever possible:
+  - eliminate [callbacks][callbacks], use a service object instead
+  - use [dependency injection][di] to encourage modularity, [an example][di-ex]
+  - follow Sandi Metz' [rules][sm]
+  - prefer [composition over inheritance][composition]
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,34 @@
...
+- Follow the [code review guildelines][codereview]
+- Write tests to cover critical functionality
+- Write a [good commit message][commit]
+- Leave code better than you found it, whenever possible:
+  - eliminate [callbacks][callbacks], use a service object instead
+  - use [dependency injection][di] to encourage modularity, [an example][di-ex]
+  - follow Sandi Metz' [rules][sm]
+  - prefer [composition over inheritance][composition]
+
+[composition]: https://betterprogramming.pub/prefer-composition-over-inheritance-1602d5149ea1
```
yachtcaptain23, line 28: IMO the article is dogmatic, other's like such as https://thoughtbot.com/blog/reusable-oo-composition-vs-inheritance suggest that the developer should understand the trade-off between composition vs inheritance.

tsmartt:
Another thought I had was mimicking the behavior of [mutations](https://github.com/cypriss/mutations) and [interactor](https://github.com/collectiveidea/interactor#an-example-interactor), but which a much simpler setup, something like:

```ruby
# frozen_string_literal: true
module My
    class Service < BaseService
        class COMMAND
            include ActiveModel::Validations
            validates_presence_of :param1
            validates_presence_of :param2
        end
      def self.build
        new(tp_need: TPNeed)
      end
      def initialize(tp_need:)
        @tp_need = tp_need
      end
      def execute(param1:, param2)
        result = @tp_need.do_something(param1, param2)
        if not result
            return fail("Some error message")
        end
        success("Did thing")
      end
    end
  end
end
```

The base class would be something like 40 lines of ruby and would give us what we need with no extra frameworks. It instantiates the COMMAND class with the parameters and validates, if validation passes, it calls execute.
By passing the params both to self.build and execute, we are free to use ruby's type system features to help stabilize the code. To invoke it you just `My::Service.build.call(param1: '1', param2: '2')`

For most things you could probably even leave out the COMMAND -- since the type system could take care of basic things like non-nil, etc.. could add it in for more complex scenarios like email validation.

CONTRIBUTING.md
tsmartt: I agree, that's why I put prefer.

tsmartt:
@yachtcaptain23 going to merge this tomorrow unless any further thoughts.

yachtcaptain23 (approved):

PR #3210, opened 2021-05-04
Address missing deposit_id's in the channel model

yachtcaptain23:
Closes https://github.com/brave-intl/publishers/issues/3177


app/controllers/payment/connection/bitflyer_connections_controller.rb
```diff
@@ -60,16 +60,8 @@ def edit
...
-          request = Net::HTTP::Get.new(url.to_s)
-          request['Authorization'] = "Bearer " + access_token
-          response = Net::HTTP.start(url.host, url.port, :use_ssl => url.scheme == 'https') do |http|
-            http.request(request)
-          end
-
-          deposit_id = JSON.parse(response.body)["deposit_id"]
-          channel.update_column(:deposit_id, deposit_id)
+          # Intentional blocking call
+          Sync::Bitflyer::UpdateMissingDepositJob.new(channel.id).perform
```
tsmartt, line 64: Nice.

app/jobs/sync/bitflyer/update_missing_deposit_job.rb
```diff
@@ -0,0 +1,22 @@
+class Sync::Bitflyer::UpdateMissingDepositJob
+  include Sidekiq::Worker
+  sidekiq_options queue: :default, retry: true
+
+  def initialize(channel_id)
+    @channel = Channel.find(channel_id)
+  end
+
+  def perform
+    return if @channel.deposit_id.present?
```
tsmartt, line 10: Can we put the core of this code in a service class? It would be nice to separate the _what_ gets run from the _where_ and _when_.

app/jobs/sync/bitflyer/update_missing_deposits_job.rb
```diff
@@ -0,0 +1,9 @@
+class Sync::Bitflyer::UpdateMissingDepositsJob < ApplicationJob
+  queue_as :low
+
+  def perform
+    Publisher.joins(:bitflyer_connection).joins(:channels).where(selected_wallet_provider: BitflyerConnection.class.name.to_s).where( channels: { deposit_id: nil}).select("channels.id").each do |channel_id|
```
tsmartt, line 5: Can we add this as a `self` method on the `Publisher` model? That way we can say something like `Publisher.blank_biflyer_deposit_ids`and keep DB access details in that file.

config/sidekiq.yml
```diff
@@ -61,6 +61,10 @@
     cron: "0 4 * * *"
     description: "For Publishers who have stale uphold codes or access parameters, clean out uphold codes or access parameter."
     queue: scheduler
+  Sync::Bitflyer::UpdateMissingDepositsJob:
```
tsmartt, line 64: Is this a 1 time fix or do we need to run it continuously?

app/jobs/channels/approve_channel_transfer_job.rb
```diff
@@ -12,15 +12,7 @@ def perform(channel_id:)
...
         if contested_by.publisher.selected_wallet_provider_type == Channel::BITFLYER_CONNECTION
-          # Request a deposit id from bitFlyer.
-          url = URI.parse(Rails.application.secrets[:bitflyer_host] + '/api/link/v1/account/create-deposit-id?request_id=' + SecureRandom.uuid)
-          request = Net::HTTP::Get.new(url.to_s)
-          request['Authorization'] = "Bearer " + contested_by.publisher.bitflyer_connection.access_token
-          response = Net::HTTP.start(url.host, url.port, :use_ssl => url.scheme == 'https') do |http|
-            http.request(request)
-          end
-          deposit_id = JSON.parse(response.body)["deposit_id"]
-          contested_by.update(deposit_id: deposit_id)
```
tsmartt, line 23: Nice.

tsmartt (commented):
Thanks for this, left a few comments.

config/sidekiq.yml
yachtcaptain23: Might be a long time run continuously as we don't know the up time of the provider.

app/jobs/sync/bitflyer/update_missing_deposit_job.rb
yachtcaptain23: Strongly prefer we don't move this to a service as we're just creating a lot of boilerplate when a job itself is a service.

config/sidekiq.yml
tsmartt: I should clarify: what's the source of the missing data? If we plug that hold then we can just run this once.

app/jobs/sync/bitflyer/update_missing_deposit_job.rb
```diff
@@ -0,0 +1,22 @@
+class Sync::Bitflyer::UpdateMissingDepositJob
+  include Sidekiq::Worker
+  sidekiq_options queue: :default, retry: true
+
+  def initialize(channel_id)
```
tsmartt, line 5: You need to pass the channel_id to the perform method for this to work with sidekiq.

app/jobs/sync/bitflyer/update_missing_deposits_job.rb
yachtcaptain23: I don't expect the scope to be re-used, but I think the lack of indentation makes it hard to read, which I will fix.

config/sidekiq.yml
yachtcaptain23: The issue is that even if we fix the data now, it's possible that the deposit_id will be missing in the future, i.e. if Bitflyer's API goes down, then it's best to automatically re-fetch the deposit ID eventually.

app/jobs/sync/bitflyer/update_missing_deposit_job.rb
yachtcaptain23: The `@channel` object is shared between the initializer and the perform

tsmartt:
@yachtcaptain23 I'll approve after the test passes.

tsmartt (approved):

PR #3215, opened 2021-05-07
Read from primary DB for admin creator search

tsmartt:
While generating the payout report, this query becomes slow
reading from the follower DB. This may be due to the high
number of connections that happen at that time, jumping from
50 to 250 as all the sidekiq jobs process, or just from general
DB slowness as the query goes from 8 seconds (already slow)
to 330. We need to look into the root cause of this, but for now
we can have this query run on the primary DB.

### DIFF
--- a/app/controllers/admin/publishers_controller.rb
+++ b/app/controllers/admin/publishers_controller.rb
@@ -4,7 +4,7 @@ class Admin::PublishersController < AdminController
   include ActiveRecord::Sanitization::ClassMethods
 
   def index
-    ActiveRecord::Base.connected_to(role: :reading) do
+    ActiveRecord::Base.connected_to(database: :primary) do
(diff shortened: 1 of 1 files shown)


app/controllers/admin/publishers_controller.rb
```diff
@@ -4,7 +4,7 @@ class Admin::PublishersController < AdminController
   include ActiveRecord::Sanitization::ClassMethods
 
   def index
-    ActiveRecord::Base.connected_to(role: :reading) do
+    ActiveRecord::Base.connected_to(database: :primary) do
```
yachtcaptain23, line 7: Since it's by default connected to primary, just remove `ActiveRecord::Base.connected_to(role: :reading) do` completely

yachtcaptain23 (approved):

app/controllers/admin/publishers_controller.rb
tsmartt: Good call, I assumed we did auto-switching, but we don't.

PR #3222, opened 2021-05-12
Fix for the statements view

yachtcaptain23:
I believe the views are fixed in this PR, but this just hides the views from the user until I can test it out in production.

### DIFF
--- a/test/services/publisher_transactions_getter_test.rb
+++ b/test/services/publisher_transactions_getter_test.rb
@@ -1,34 +1,34 @@
 require "test_helper"
 
 class PublisherBalanceGetterTest < ActiveJob::TestCase
+  describe "mocked transactions" do
+    before(:example) do
+      @prev_offline = Rails.application.secrets[:api_eyeshade_offline]
+      Rails.application.secrets[:api_eyeshade_offline] = false
 
-  before(:example) do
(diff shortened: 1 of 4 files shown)


test/services/publisher_transactions_getter_test.rb
```diff
@@ -1,34 +1,34 @@
 require "test_helper"
 
 class PublisherBalanceGetterTest < ActiveJob::TestCase
+  describe "mocked transactions" do
+    before(:example) do
+      @prev_offline = Rails.application.secrets[:api_eyeshade_offline]
+      Rails.application.secrets[:api_eyeshade_offline] = false
```
tsmartt, line 7: Setting this here will affect other tests and can lead to random failures, can we find a way to limit this change to this test only? I'd prefer something injectable, but in a pinch[ a stub](https://www.rubydoc.info/gems/minitest/4.2.0/Object:stub) should work.

tsmartt (commented):
Looks good overall, 1 small comment.

test/services/publisher_transactions_getter_test.rb
yachtcaptain23: The lines below

```   
after(:example) do
      Rails.application.secrets[:api_eyeshade_offline] = @prev_offline
    end
```

will reset the state

tsmartt (approved):

PR #3224, opened 2021-05-12
Fix setting the JA locale in the URL

tsmartt:
Previously we could set locale=JA in the URL and see pages in
Japanese, but broke that feature recently. This restores it and adds
appropriate testing. Note that you can use a browser sending an
English accept header and set this locale and have it work. If you
are using a browser where Japanese is the preferred language,
you will automatically see Japanese pages. You can not yet manually
set the locale to English if we detect you are using an JA browser,
but we could add it later if pressed.

### DIFF
--- a/test/features/locale_test.rb
+++ b/test/features/locale_test.rb
@@ -1,0 +1,36 @@
+require "test_helper"
+
+class LocaleTest < Capybara::Rails::TestCase
+  include ActionMailer::TestHelper
+  include Devise::Test::IntegrationHelpers
+  include Rails.application.routes.url_helpers
+
+  before do
+    publisher = publishers(:small_media_group)
+    sign_in publisher
+  end
+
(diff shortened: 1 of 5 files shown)


test/features/locale_test.rb
```diff
@@ -0,0 +1,36 @@
...
+  test "login with JA accept language and no locale shows Japanese" do
+    Capybara.using_driver("chromeja") do
+        visit home_publishers_path
+        assert_content page, "の広告を配信"
+    end
+  end
+
+  test "login with JA accept language and EN locale shows Japanese" do
+    Capybara.using_driver("chromeja") do
+        visit home_publishers_path(locale: 'EN')
```
yachtcaptain23, line 32: Would you add a comment saying this might not be the best solution?

yachtcaptain23 (approved):

PR #3229, opened 2021-05-18
Add task for backfilling selected wallet

tsmartt:
We have a lot of Uphold wallets that are not set as the selected
provider, so we chunk the work on that one. The duplicates in the
system have already been taken care of, and we don't have to do
anything for paypal.

### DIFF
--- a/lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
+++ b/lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
@@ -4,7 +4,6 @@ namespace :database_updates do
     def handle_gemini
       now = Time.zone.now
       gemini_to_set = Publisher.
-        includes([:gemini_connection]).
         joins(:gemini_connection).
         where('gemini_connections.is_verified = TRUE').
         where('publishers.email IS NOT NULL').
@@ -14,7 +13,6 @@ namespace :database_updates do
       # For testing in dev
       # publisher_ids = GeminiConnection.first(3).map { |p| p.publisher.id }
       # gemini_to_set = Publisher.where(id: publisher_ids).
-      #   includes([:gemini_connection]).
       #   joins(:gemini_connection).
       #   pluck(:id, 'gemini_connections.id')
 
@@ -35,7 +33,6 @@ namespace :database_updates do
     def handle_bitflyer
       now = Time.zone.now
       bitflyer_to_set = Publisher.
-        includes([:bitflyer_connection]).
         joins(:bitflyer_connection).
         where('publishers.email IS NOT NULL').
         where('publishers.selected_wallet_provider_id IS NULL').
@@ -44,7 +41,6 @@ namespace :database_updates do
       # For testing in dev
       # publisher_ids = BitflyerConnection.first(2).map { |p| p.publisher.id }
       # bitflyer_to_set = Publisher.where(id: publisher_ids).
-      #   includes([:bitflyer_connection]).
       #   joins(:bitflyer_connection).
       #   pluck(:id, 'bitflyer_connections.id')
 
@@ -67,7 +63,6 @@ namespace :database_updates do
       now = Time.zone.now
 
       query_base = Publisher.
-        includes([:uphold_connection]).
         joins(:uphold_connection).
         where('uphold_connections.uphold_verified = TRUE').
         where('publishers.email IS NOT NULL').
@@ -76,7 +71,6 @@ namespace :database_updates do
       # For testing in dev
       # publisher_ids = UpholdConnection.first(3).map { |p| p.publisher.id }
       # query_base = Publisher.where(id: publisher_ids).
-      #   includes([:uphold_connection]).
       #   joins(:uphold_connection)
 
       query_base.in_batches(of: limit) do |uphold_batch|
--- a/lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
+++ b/lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
@@ -2,11 +2,9 @@ namespace :database_updates do
   desc 'Backfill Publisher Selected Wallets'
   task :backfill_publisher_selected_wallets => :environment do
     def handle_gemini
-      now = Time.zone.now
       gemini_to_set = Publisher.
         joins(:gemini_connection).
         where('gemini_connections.is_verified = TRUE').
-        where('publishers.email IS NOT NULL').
         where('publishers.selected_wallet_provider_id IS NULL').
         pluck(:id, 'gemini_connections.id') # use pluck to avoid loading user_authentication_token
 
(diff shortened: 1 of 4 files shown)


lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
```diff
@@ -0,0 +1,105 @@
...
+  task :backfill_publisher_selected_wallets => :environment do
+    def handle_gemini
+      now = Time.zone.now
+      gemini_to_set = Publisher.
+        includes([:gemini_connection]).
+        joins(:gemini_connection).
+        where('gemini_connections.is_verified = TRUE').
+        where('publishers.email IS NOT NULL').
+        where('publishers.selected_wallet_provider_id IS NULL').
+        pluck(:id, 'gemini_connections.id') # use pluck to avoid loading user_authentication_token
```
tsmartt, line 12: @yachtcaptain23 Note this is optimized for speed. Pluck avoids loading the associated token, and below with Uphold we chunk the work.

tsmartt: Note we also use upsert_all to avoid updating one at a time. One downside of that is that the timestamps are reset, but the speed is worth it.

lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
```diff
@@ -0,0 +1,105 @@
+namespace :database_updates do
+  desc 'Backfill Publisher Selected Wallets'
+  task :backfill_publisher_selected_wallets => :environment do
+    def handle_gemini
+      now = Time.zone.now
+      gemini_to_set = Publisher.
+        includes([:gemini_connection]).
+        joins(:gemini_connection).
```
yachtcaptain23, line 8: Would you use `joins` in this case? I don't think there's any performance gain by using `includes` in our case

lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
```diff
@@ -0,0 +1,105 @@
+namespace :database_updates do
+  desc 'Backfill Publisher Selected Wallets'
+  task :backfill_publisher_selected_wallets => :environment do
+    def handle_gemini
+      now = Time.zone.now
```
yachtcaptain23, line 5:
```suggestion
      now = Time.now
```

lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
```diff
@@ -0,0 +1,105 @@
...
+      gemini_to_set = Publisher.
+        includes([:gemini_connection]).
+        joins(:gemini_connection).
+        where('gemini_connections.is_verified = TRUE').
+        where('publishers.email IS NOT NULL').
```
yachtcaptain23, line 10:
```suggestion
```

lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
```diff
@@ -0,0 +1,105 @@
...
+      #   includes([:gemini_connection]).
+      #   joins(:gemini_connection).
+      #   pluck(:id, 'gemini_connections.id')
+
+      gemini_new_records = gemini_to_set.map do |publisher_gemini|
+        {
+          id: publisher_gemini[0], # publisher id
+          selected_wallet_provider_id: publisher_gemini[1], # gemini connections id
+          selected_wallet_provider_type: 'GeminiConnection',
+          created_at: now,
```
yachtcaptain23, line 26: We should not be updating `created_at`

tsmartt: One of the current [downsides of upsert_all](https://medium.com/@retrorubies/upcoming-rails-6-bulk-insert-upsert-feature-2d642419557d). We could change the defaults [in a migration](https://stackoverflow.com/a/65546384), I'll work on a migration tomorrow.

lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
tsmartt: Any reason to go against the recommendation [here](https://rails-bestpractices.com/posts/2014/10/22/use-time-zone-now-instead-of-time-now/)?

lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
tsmartt: If you don't have an email, does it matter? I don't mind changing it, but I remember it made a difference. Will have to look it up. Perhaps it's due to bad data in the DB.

lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
tsmartt: Joins doesn't automatically eager load the data, which I need when getting the connection id. 

Well, this used to be relevant until I switched to pluck -- since it's array based now, then includes is unnecessary. I can remove it tomorrow.

tsmartt: Done.

lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
yachtcaptain23: I really am against updating the `created_at` field because we use it for metrics like when user's joined. Alternatively you could just refactor this to use https://stackoverflow.com/questions/6629088/bulk-record-update-with-sql

Also totally fine with an N+1 query given that there's 448 uphold accounts that are missing a type.

tsmartt: This has been fixed with the migration.

lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
tsmartt: This is now gone.

lib/tasks/database_updates/backfill_publisher_selected_wallet.rake
```diff
@@ -0,0 +1,67 @@
...
+    def handle_gemini
+      # Can't use update_all https://github.com/rails/rails/issues/522
+      # https://stackoverflow.com/questions/1293330/how-can-i-do-an-update-statement-with-join-in-sql-server
+      sql = """
+      UPDATE publishers
+        SET selected_wallet_provider_id = gc.id,
+            selected_wallet_provider_type = 'GeminiConnection',
+            updated_at = '#{Time.zone.now}'
+      FROM gemini_connections AS gc
+      WHERE gc.publisher_id = publishers.id
```
yachtcaptain23, line 13:
```suggestion
      FROM publishers
      INNER JOIN gemini_connections as gc
      WHERE gc.publisher_id = publishers.id
```

yachtcaptain23 (approved):

yachtcaptain23:
LG2M! Thanks for the refactor

PR #3238, opened 2021-05-21
Rely on selected_wallet_provider for payouts

tsmartt:
Now that the duplicates have been removed and the wallets
backfilled, we can rely on this for pretty much anything in
the system. Let's start with the payouts.


app/services/payout/service.rb
```diff
@@ -25,7 +25,7 @@ def create_message(message)
     # Internal: Creates entries for any reason for not paying out a publisher.
     #
     # Returns a boolean
-    def skip_publisher?(is_bitflyer: false)
+    def skip_publisher?()
```
yachtcaptain23, line 28:
```suggestion
    def skip_publisher?
```

app/models/publisher.rb
```diff
@@ -103,6 +103,32 @@ class Publisher < ApplicationRecord
...
+      or(
+        joins(:gemini_connection).
+        where.not("gemini_connections.country ILIKE '#{GeminiConnection::JAPAN}'").
+        where(selected_wallet_provider_type: 'GeminiConnection')
+      )
+  end
+
+  def self.bitflyer_creators
+    where(selected_wallet_provider_type: 'BitflyerConnection')
+  end
```
yachtcaptain23, line 130: Nit: I think we should use constants. This particular column has given me issues.

tsmartt: Good idea, I'll update!

app/models/publisher.rb
```diff
@@ -103,6 +103,32 @@ class Publisher < ApplicationRecord
...
 
+  def self.uphold_creators
+    joins(:uphold_connection).
+      where('uphold_connections.country IS NULL').
+      where(selected_wallet_provider_type: 'UpholdConnection').
+      or(
+        joins(:uphold_connection).
+        where.not("uphold_connections.country ILIKE '#{UpholdConnection::JAPAN}'").
+        where(selected_wallet_provider_type: 'UpholdConnection')
+      )
```
yachtcaptain23, line 114:
```suggestion
    joins(:uphold_connection).
      where(selected_wallet_provider_type: 'UpholdConnection').
      where.not("uphold_connections.country ILIKE '#{UpholdConnection::JAPAN}'").
      where(uphold_connections: { is_member: true})
```

I think this simplifies things a bit. Also we should be paying `is_member` this whole time

tsmartt: You have to include the `or` or you won't pick up items where the country `IS NULL`.

tsmartt: Made a ticket to track this https://github.com/brave-intl/publishers/issues/3243

app/models/publisher.rb
```diff
@@ -103,6 +110,31 @@ class Publisher < ApplicationRecord
...
   store_accessor :feature_flags, VALID_FEATURE_FLAGS
 
+  def self.valid_payable_uphold_creators
+    where(selected_wallet_provider_type: UPHOLD_CONNECTION)
+    where('uphold_connections.country IS NULL').
+      or(
+        where(selected_wallet_provider_type: UPHOLD_CONNECTION).
+        where.not("uphold_connections.country ILIKE '#{UpholdConnection::JAPAN}'")
+      ).
+      joins(:uphold_for_join)
```
yachtcaptain23, line 120:
```suggestion
      joins(:uphold_for_join).
      where.not("uphold_connections.country ILIKE '#{UpholdConnection::JAPAN}'")
```

I don't think you'd need to do `(country == null || country != "JP")`

tsmartt: If you just say country != JP then it won't pick up NULL 
https://stackoverflow.com/questions/8036691/sql-not-displaying-null-values-on-a-not-equals-query

yachtcaptain23: Oh good point. I don't want to hold up the PR but the joins part at the bottom and the naming of the `belongs_to` kind of violates principle of least surprise. I think a scope that does a join with `ON publishers.selected_wallet_provider_id = uphold_connections.id AND publishers.selected_provider_type_provider = 'UpholdConnection'` would save on readability.

tsmartt: Ah yes, I should clarify that AR complains if I put it at the top, but b/c of the way AREL works wants it at the bottom. I agree it looks better at the top.

app/services/payout/uphold_job_implementation.rb
```diff
@@ -0,0 +1,49 @@
...
+      kind = IncludePublisherInPayoutReportJob::UPHOLD
+      if manual && payout_report.present?
+        kind = IncludePublisherInPayoutReportJob::MANUAL
+      end
+
+      publishers.find_each do |publisher|
+        @payout_report_job.perform_async(
+          payout_report_id: payout_report_id,
+          publisher_id: publisher.id,
+          kind: kind,
```
yachtcaptain23, line 41: Nit: Out-of-scope on this PR, but we should just `select` or `pluck` the publisher id rather than possibly fetching extra data.

app/services/payout/bitflyer_job_implementation.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Payout
+  class BitflyerJobImplementation
```
yachtcaptain23, line 4: Nit picking, but this is a service rather than a job.

The underlying issue I see that existed before this code is that the caller could be killed for whatever reason before it completes without ever retrying. As a a result I'm punting it to https://github.com/brave-intl/publishers/issues/3245

app/services/payout/uphold_job_implementation.rb
yachtcaptain23: Punting to https://github.com/brave-intl/publishers/issues/3245

yachtcaptain23 (approved):
LG2M. There were a few issues unrelated to this PR that I've noted and punted to https://github.com/brave-intl/publishers/issues/3245

PR #3256, opened 2021-05-28
Pass the display_name to antifraud

tsmartt:
We also don't need to generate potential payments
for referrals for Bitflyer for the time being, until
we dedide we want to.

Note that display_name is only set on successful authentication, so will be blank if oauth didn't complete, but then so will deposit_id. I made a task to clean this filtering up later https://github.com/brave-intl/publishers/issues/3257

### DIFF
--- a/app/services/payout/bitflyer_service.rb
+++ b/app/services/payout/bitflyer_service.rb
@@ -36,7 +36,7 @@ module Payout
           kind: ::PotentialPayment::CONTRIBUTION,
           url: "#{channel.details.url}",
           address: channel.deposit_id || '',
-          wallet_provider_id: connection.display_name || '',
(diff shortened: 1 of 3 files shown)


app/services/payout/bitflyer_service.rb
```diff
@@ -47,7 +36,7 @@ def perform(publisher:, payout_report:)
           kind: ::PotentialPayment::CONTRIBUTION,
           url: "#{channel.details.url}",
           address: channel.deposit_id || '',
-          wallet_provider_id: channel.deposit_id || '',
+          wallet_provider_id: connection.display_name || '',
```
yachtcaptain23, line 39:
```suggestion
          wallet_provider_id: connection.display_name || '', # this is a hash of the account_id
```

yachtcaptain23 (approved):

PR #3259, opened 2021-05-28
Upgrade to Rails 6.1

tsmartt:
I'll go through and leave comments on the sections I thought were interesting.


test/controllers/site_channels_controller_test.rb
```diff
@@ -91,123 +91,117 @@ class SiteChannelsControllerTest < ActionDispatch::IntegrationTest
     assert_redirected_to controller: "/site_channels", action: "verification_wordpress", id: channel.id
   end
 
-  test "can't create verified Site Channel with an existing verified Site Channel with the same brave_publisher_id" do
```
tsmartt, line 94: To Fix.

test/controllers/active_storage/blobs_controller_test.rb
```diff
@@ -40,9 +40,7 @@ class ActiveStorage::BlobsControllerTest < ActionDispatch::IntegrationTest
         subject
       end
 
-      it 'flashes an unauthorized message' do
```
tsmartt, line 43: To fix, though I don't think this actually worked before, but was ignored due to missing translations.

test/controllers/active_storage/blobs_controller_test.rb
```diff
@@ -65,8 +63,6 @@ class ActiveStorage::BlobsControllerTest < ActionDispatch::IntegrationTest
   describe 'when the user is not authenticated' do
     before { subject }
 
-    it 'redirects and flashes a message' do
```
tsmartt, line 68: To fix, though I don't think this actually worked before, but was ignored due to missing translations.

test/controllers/site_channels_controller_test.rb
tsmartt: Fixed.

test/controllers/active_storage/blobs_controller_test.rb
tsmartt: Removed due to being not needed.

app/models/base_channel_details.rb
```diff
@@ -1,4 +1,4 @@
-class BaseChannelDetails < ActiveRecord::Base
+class BaseChannelDetails < ApplicationRecord
```
tsmartt, line 1: This one caused a headache as Rails changed the way connects_to works. In the tests, it was using a separate connection which couldn't see the changes in the transaction for the site_channels_controller test. Since it shares the same DB preferences as the application_controller, I just subclassed it.

app/models/concerns/public_s3.rb
```diff
@@ -60,16 +60,16 @@ def #{name}=(attachable)
         # When building a blob we must ensure the correct S3 service is being specified so that the backing code
         # uploads to the correct bucket. Otherwise it will use the default configured in the config/storage.yml
         def upload_public_#{name}(file)
-          blob = ActiveStorage::Blob.new.tap do |blob|
```
tsmartt, line 63: Quite a few Active Storage changes in 6.1, so I had to redo how we set up blobs. Namely, the service cannot be set directly and is immutable, so you construct it at build time.

db/migrate/20210527190724_add_service_name_to_active_storage_blobs.active_storage.rb
```diff
@@ -0,0 +1,18 @@
+# This migration comes from active_storage (originally 20190112182829)
+class AddServiceNameToActiveStorageBlobs < ActiveRecord::Migration[6.0]
```
tsmartt, line 2: FYI that we will need a migration in prod.

test/jobs/channels/transfer_channels_job_test.rb
```diff
@@ -38,7 +38,9 @@ class Channels::TransferChannelsJobTest < ActiveJob::TestCase
       assert_enqueued_jobs 1
       assert_performed_jobs 0
 
-      perform_enqueued_jobs
+      perform_enqueued_jobs do
```
tsmartt, line 41: This needs a block now to also process jobs added during the run.

yachtcaptain23 (approved):
LG2M. We should test in staging to make sure images for publisher banner gets loaded and saved correctly.

PR #3261, opened 2021-06-01
Closes #3035

### LINKED ISSUE #3035: Payout statements broken for Gemini creators
## Description

Currently, the payout statements section in the creators.brave.com dashboard only expects Uphold. If the Creator is connected to Gemini, the payout page displays a notice/error saying that there is an issue reading the transactions from Uphold, even though the Creator does not use Uphold. 

**Screenshot:**

image

tsmartt:
Closes #3035
For now we will not show this message if the user's selected wallet
is not an uphold wallet. If it is and we don't have the proper scope,
then show the message for reauthorization.


app/models/publisher.rb
```diff
@@ -210,6 +210,10 @@ def sync_wallet_connections
     end
   end
 
+  def uphold_publisher?
```
yachtcaptain23, line 213:
```suggestion
  def is_selected_wallet_provider_uphold?
```

Prefer renaming to something more understandable at-a-glance.

app/views/publishers/statements/index.html.slim
```diff
@@ -9,31 +9,8 @@ noscript
   )
   = render partial: "choose_channel_button"
 
-/ .single-panel--wrapper.single-panel--wrapper--xlarge#statements_panel
```
yachtcaptain23, line 12: Would you see if `app/assets/stylesheets/pages/statements.scss` is removable too?

app/models/publisher.rb
tsmartt: Fixed.

app/views/publishers/statements/index.html.slim
tsmartt: It was, fixed.

yachtcaptain23 (approved):

Issue #3124, opened 2021-03-23
spectre mitigation


diracdeltas:
ensure publishers is following all the guidelines in https://w3c.github.io/webappsec-post-spectre-webdev/#tldr

tsmartt:
Most likely falls onto AWS, look into it.

tsmartt:
Indeed https://aws.amazon.com/speculative-execution-os-updates/

PR #3270, opened 2021-06-05
Enable CSP policy in reporting mode

tsmartt:
We want to ensure a good CSP policy. This was initially driven by
the need to look at spectre mitigations (#3124), but as part of that
involves a good CSP policy, i.e.
https://security.googleblog.com/2020/07/towards-native-security-defenses-for.html

we need to get a good CSP policy in place to begin with.

For now, we'll just log the violations. We can quickly clean them up
and keep an eye on any new violations that come in over time, eventually
phasing out this code.

### DIFF
--- a/app/controllers/csp_violations_report_controller.rb
+++ b/app/controllers/csp_violations_report_controller.rb
@@ -1,0 +1,22 @@
+class CspViolationsReportController < ApplicationController
+  skip_before_action :verify_authenticity_token
+
+  def create
+    # Reports look something like:
+    # {"document-uri"=>"https://localhost:3000/",
+    #   "referrer"=>"",
+    #   "violated-directive"=>"style-src-elem",
+    #   "effective-directive"=>"style-src-elem",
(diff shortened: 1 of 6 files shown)


app/controllers/csp_violations_report_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    #   "line-number"=>4,
+    #   "source-file"=>"https://localhost:3000/",
+    #   "status-code"=>0,
+    #   "script-sample"=>""}
+    # We only care about the document-uri, blocked-uri, and directives for now
+    report = JSON.parse(request.body.read)['csp-report']
+    sliced_report = report.slice('document-uri', 'violated-directive', 'effective-directive', 'blocked-uri')
+    begin
+      CspViolationReport.create(report: sliced_report)
+    rescue
```
yachtcaptain23, line 22: Why do we drop the error? Would be good to know if an error happens when it gets reported to Sentry

tsmartt: Ah, in this case it was reporting the validation error during testing, i.e. someone else already hit this error. Since this is not critical functionality and since it will be hit often, I thought it best to just drop it.

yachtcaptain23 (approved):

PR #3272, opened 2021-06-08
Destroy Uphold Wallets Instead of Clearing

### LINKED ISSUE #3271: TwoFactorAuthenticationRemoval failing due to selected_wallet_provider being nil
There's a bug where if no selected wallet exists, we hit a nil exception here https://github.com/brave-intl/publishers/blob/staging/app/models/uphold_connection.rb#L98

We should just destroy the connections like we do for gemini and bitflyer.

tsmartt:
We need to keep track of deletions of wallet connections, so
we add papertrail on gemini and bitflyer.

We also now just destroy the uphold connection when disconnecting,
this will fix the issue with TwoFactorAuth removal as well as
bring it into line with the other connection types.

### DIFF
--- a/test/controllers/publishers_controller_test.rb
+++ b/test/controllers/publishers_controller_test.rb
@@ -21,14 +21,14 @@ class PublishersControllerTest < ActionDispatch::IntegrationTest
 
   SIGNUP_PARAMS = {
     email: "alice@example.com",
-    terms_of_service: true
-  }
+    terms_of_service: true,
+  }.freeze
 
   COMPLETE_SIGNUP_PARAMS = {
     publisher: {
       name: "Alice the Pyramid",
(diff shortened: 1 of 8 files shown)


test/controllers/publishers_controller_test.rb
```diff
@@ -519,7 +515,7 @@ def request_login_email_uppercase_email(publisher:)
     assert_response 200
 
     publisher.reload
-    refute publisher.uphold_connection.uphold_verified?
+    refute publisher.uphold_connection
```
tsmartt, line 518: This is the real test change. The rest was Rubocop formatting the file, which was apparently not in good shape.

yachtcaptain23 (approved):
LG2M. Please change PR header

PR #3273, opened 2021-06-08
Remove remaining 5 BAT messaging and code

### LINKED ISSUE #2499: Do not show pending payout progress for payouts below 5.00 BAT, or show appropriate notice
### Issue

Currently, the interface for payout progress leads creators to believe that they will be receiving payouts of any amount, including payouts below 5.00 BAT. However, there is currently a ~5.00 BAT minimum for payouts.

### Solution

Implement appropriate conditional logic in the interface for the less than 5.00 BAT case. If `current payout < minimum payout threshold`, display appropriate notice telling the creator that they will not be receiving their payout because it falls below the minimum payout threshold. 

Example of issue:

image

### LINKED ISSUE #2706: Fix "5 BAT minimum" messaging (e.g., next to each individual channel instead of only next to the total sum / pending amount)
Currently, the 5 BAT minimum notice is written below the total pending sum in the top left corner. However, the 5 BAT minimum applies to _each channel individually_, meaning that a user can have (example case):

- Channel A: 3.25 BAT
- Channel B: 4.00 BAT
- Channel C: 4.75 BAT

The sum of channels A + B + C = 12 BAT, which is what is displayed in the top left corner. 12 BAT is > 5 BAT, so creators are led to believe they have passed the threshold for a payout.

However, the creator will **not** receive their payout for that month, given no individual channel is above 5 BAT. 

Perhaps we can introduce some more precise messaging to reflect the above payout logic.

tsmartt:
Closes #2499 
Closes #2706

### DIFF
--- a/app/helpers/publishers_helper.rb
+++ b/app/helpers/publishers_helper.rb
@@ -56,12 +56,14 @@ module PublishersHelper
 
   def has_balance?(publisher)
     penny = 0.01
-    publisher.wallet&.contribution_balance&.channel_amounts_usd&.any? do |channel_amount_usd|
-      channel_amount_usd >= penny
+    publisher.wallet&.contribution_balance&.channel_amounts_usd&.each do |channel_amount_usd|
+      if channel_amount_usd >= penny
+        return true
+      end
     end
 
     referral_balance = publisher.wallet&.referral_balance&.amount_usd
-    return referral_balance && referral_balance >= penny
+    referral_balance && referral_balance >= penny
   end
 
   def publisher_overall_bat_balance_amount(publisher)
@@ -108,14 +110,6 @@ module PublishersHelper
       end
     end
     result
-  end
-
-  def has_minimum_usd_for_payout?(publisher)
-    amount = publisher_overall_bat_balance_amount(publisher)
-    if amount.present? && amount >= PayoutReport::MINIMUM_BALANCE_AMOUNT
-      return true
-    end
-    false
   end
 
   def publisher_referral_bat_balance(publisher)
--- a/app/helpers/publishers_helper.rb
+++ b/app/helpers/publishers_helper.rb
@@ -96,6 +96,29 @@ module PublishersHelper
     result
   end
 
+  def qualifies_for_payout?(publisher)
+    publisher.selected_wallet_provider_type == Publisher::UPHOLD_CONNECTION && has_minimum_usd_for_payout?(publisher)
+  end
+
+  def has_minimum_usd_for_payout?(publisher)
+    amount = nil
+
+    sentry_catcher do
+      if publisher.only_user_funds?
+        amount = publisher.wallet&.contribution_balance&.amount_usd
+      elsif publisher.no_grants?
+        amount = publisher.wallet&.overall_balance&.amount_usd - publisher.wallet&.contribution_balance&.amount_usd
+      else
+        amount = publisher.wallet&.overall_balance&.amount_usd
+      end
+
+      if amount.present? && amount >= PayoutReport::MINIMUM_BALANCE_AMOUNT
+        return true
+      end
+    end
+    false
+  end
+
   def publisher_referral_bat_balance(publisher)
     balance = I18n.t("helpers.publisher.balance_unavailable")
     sentry_catcher do
--- a/app/helpers/publishers_helper.rb
+++ b/app/helpers/publishers_helper.rb
@@ -54,8 +54,13 @@ module PublishersHelper
     today.strftime("%B 8th")
   end
 
-  def publisher_overall_bat_balance(publisher)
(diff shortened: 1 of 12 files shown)


app/services/payout/uphold_service.rb
```diff
@@ -111,19 +101,13 @@ def send_emails(uphold_connection, total_amount)
       end
     end
 
-
     def should_send_emails?(total_probi:, uphold_connection:)
```
yachtcaptain23, line 104: Would you remove this block too?

tsmartt: Merged your other PR which did it.

app/models/payout_report.rb
```diff
@@ -4,7 +4,6 @@ class PayoutReport < ApplicationRecord
 
   LEGACY_PAYOUT_REPORT_TRANSITION_DATE = "2018-12-01 09:14:58 -0800".freeze
 
-  MINIMUM_BALANCE_AMOUNT = 5
```
yachtcaptain23, line 7: We should keep this and move this to about 0.01 USD, as users with 0 balance shouldn't see the message.

tsmartt: JA no longer needed since we should not have any minimums for BF.

app/views/publishers/_home_balances.html.slim
```diff
@@ -15,10 +15,13 @@
...
-              - if converted_balance.present? && converted_balance.to_i>= PayoutReport::MINIMUM_BALANCE_AMOUNT
-                i = t(".payout_in_progress")
+              - if @publisher.is_selected_wallet_provider_uphold?
+                - if qualifies_for_payout?(@publisher)
+                  = t(".payout_in_progress_uphold", amount: PayoutReport::MINIMUM_BALANCE_AMOUNT)
+                - else
+                  = t(".payout_minimum_balance", amount: PayoutReport::MINIMUM_BALANCE_AMOUNT)
               - else
-                i = t(".payout_minimum_balance", amount: PayoutReport::MINIMUM_BALANCE_AMOUNT, currency: t("bat_in_locale"))
+                = t(".payout_in_progress")
```
yachtcaptain23, line 24: You should only show if they have a balance

app/helpers/publishers_helper.rb
```diff
@@ -96,6 +96,29 @@ def publisher_converted_overall_balance(publisher)
     result
   end
 
+  def qualifies_for_payout?(publisher)
+    publisher.selected_wallet_provider_type == Publisher::UPHOLD_CONNECTION && has_minimum_usd_for_payout?(publisher)
```
yachtcaptain23, line 100: The caller already has `publisher.selected_wallet_provider_type == Publisher::UPHOLD_CONNECTION`. I think you should completely remove this block and move `has_minimum_usd_for_payout?(publisher)` to the caller's block

tsmartt: Done.

app/views/publishers/_home_balances.html.slim
tsmartt: Done.

lib/eyeshade/contribution_balance.rb
```diff
@@ -1,20 +1,27 @@
...
     def initialize(rates, default_currency, accounts)
       super(rates, default_currency)
 
       channel_accounts = accounts.select { |account| account["account_type"] == CHANNEL}
 
+      @channel_amounts_usd = []
+
       channel_accounts.each do |account|
         channel_balance = Eyeshade::ChannelBalance.new(@rates, @default_currency, account)
+        @channel_amounts_usd << convert(probi_to_bat(channel_balance.amount_probi), "USD")
```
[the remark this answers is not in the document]
tsmartt: We are checking if any of the channels have at least something to pay out.

app/helpers/publishers_helper.rb
```diff
@@ -54,8 +54,20 @@ def next_deposit_date(today: DateTime.now)
     today.strftime("%B 8th")
   end
 
-  def publisher_overall_bat_balance(publisher)
-    balance = I18n.t("helpers.publisher.balance_unavailable")
+  def has_balance?(publisher)
+    penny = 0.01
```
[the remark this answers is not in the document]
tsmartt: Done.

app/helpers/publishers_helper.rb
```diff
@@ -54,8 +54,20 @@ def next_deposit_date(today: DateTime.now)
...
+    publisher.wallet&.contribution_balance&.channel_amounts_usd&.each do |channel_amount_usd|
+      if channel_amount_usd >= penny
+        return true
```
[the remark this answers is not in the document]
tsmartt: Done.

yachtcaptain23 (approved):

Issue #3286, opened 2021-06-21
Add retry mechanism for payout reports


yachtcaptain23:
Given that the UpholdJob doesn't have a retry mechanism, this has given payouts issues with paying out for referrals and contributions. 

To make retrying possible, one needs to redesign the affected payout jobs to use the Sidekiq retry mechanism and restarting from the previous position.

PR #3287, opened 2021-06-21
Re-architecture of UpholdJob, BitflyerJob, and GeminiJob to use shared

yachtcaptain23:
Re-architecture of UpholdJob, BitflyerJob, and GeminiJob to use shared logic and to handle restarts in case if enqueueing payouts fails.

This implementation is currently meant to be run single threaded, and might need refactoring to use less memory in the :pluck field. However, given that uuid has 32 characters, we likely won't hit on-device memory constraints anytime soon.

Closes https://github.com/brave-intl/publishers/issues/3286


app/jobs/enqueue_publishers_for_payout_job.rb
```diff
@@ -38,29 +38,27 @@ def perform(should_send_notifications: false, final: true, manual: false, payout
...
-    )
-    Payout::GeminiJob.perform_later(
-      payout_report_id: payout_report.id,
-      publisher_ids: publisher_ids
-    )
-    Payout::BitflyerJob.perform_later(
-      payout_report_id: payout_report.id,
-      publisher_ids: publisher_ids
-    )
+    }.to_json
```
tsmartt, line 58: Nice.

app/jobs/payout/concerns/retries_from_last_enqueued_publisher.rb
```diff
@@ -0,0 +1,42 @@
+module RetriesFromLastEnqueuedPublisher
+  extend ActiveSupport::Concern
+
+  def enqueue_retryable_potential_payout!
```
tsmartt, line 4: 👏  for making the 3 classes 1. Nice reduction in code. Going to a concern here is a problem as it spreads variables across files and makes it harder to reason about. Please extract this to a service class to handle this logic.

test/jobs/payout/uphold_job_test.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  test "for order and key presence" do
+    Sidekiq::Worker.clear_all
+    uphold_job = Payout::UpholdJob.new
+    uphold_job.perform(json_args: { payout_report_id: payout_reports(:one).id }.to_json)
+    visited_publisher_id = nil
+    assert IncludePublisherInPayoutReportJob.jobs.count > 1
+    IncludePublisherInPayoutReportJob.jobs.each do |job|
+      job_publisher_id = job["args"].first["publisher_id"]
+      assert visited_publisher_id <= job_publisher_id if visited_publisher_id.present?
```
tsmartt, line 29: What if the `visited_publisher_id` is never present?

app/jobs/payout/concerns/retries_from_last_enqueued_publisher.rb
```diff
@@ -0,0 +1,42 @@
...
+    # self.class.name is the name of the caller function
+    last_enqueued_publisher_key = "#{self.class.name}-#{@payout_report_id || ""}"
```
tsmartt, line 6: Can you write up a section about what the technical underlying issue is and why this method fixes it?

tsmartt (changes requested):

test/jobs/payout/uphold_job_test.rb
yachtcaptain23: Test checks it at the end

app/jobs/payout/concerns/retries_from_last_enqueued_publisher.rb
yachtcaptain23: Sure, I'll move it to a Service.

test/jobs/payout/uphold_job_test.rb
tsmartt: It is compared to something which might be nil.

app/jobs/payout/service/retries_from_last_enqueued_publisher.rb
```diff
@@ -0,0 +1,47 @@
...
+        number_of_payments = PayoutReport.expected_num_payments(publishers)
+        payout_report.with_lock do
+          payout_report.reload
+          payout_report.expected_num_payments = number_of_payments + payout_report.expected_num_payments
+          payout_report.save!
+        end
+      end
+    end
+
+    publishers.pluck(:id).each do |publisher_id|
```
tsmartt, line 37: maybe use [in_batches ](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-in_batches)here?

test/jobs/payout/uphold_job_test.rb
yachtcaptain23: It's checked at the end of the line with `if visited_publisher_id.present?`

app/jobs/payout/service/retries_from_last_enqueued_publisher.rb
yachtcaptain23: `in_batches` returns a ActiveRecord::Relation, whereas `find_each` handles batching

app/jobs/payout/concerns/retries_from_last_enqueued_publisher.rb
yachtcaptain23: Discussed offline. Resolving

tsmartt (approved):

PR #3308, opened 2021-06-30
Feature/separate referral queue

yachtcaptain23:


test/jobs/payout/enqueue_top_referrer_payout_job_test.rb
```diff
@@ -0,0 +1,20 @@
+require 'test_helper'
+
+class EnqueueTopReferrerPayoutJobTest < ActiveJob::TestCase
+  test 'it enqueues only top referrers to to call EnqueuePublishersForPayoutJob' do
+    Sidekiq::Worker.clear_all
```
tsmartt, line 5: Since we've done this twice recently, maybe we can move this to a global before call for all ActiveJob related tests?

test/services/payout/uphold_service_test.rb
```diff
@@ -412,7 +412,7 @@ class UpholdServiceTest < ActiveJob::TestCase
                 kind: 'referral'
               )
 
-              assert referral_payments.size > 0
+              assert_empty referral_payments
```
tsmartt, line 415: So where's the test that potential payments for referrals are created? This is what this test was going, but now it asserts that they aren't created.

tsmartt (changes requested):
Looks good with a few comments.

test/services/payout/uphold_service_test.rb
yachtcaptain23: Good catch! I'll fix this

tsmartt (approved):
I requested an additional test, but won't hold up the PR on it. LG2m.

yachtcaptain23:
> I requested an additional test, but won't hold up the PR on it. LG2m.

I added one :)

PR #3311, opened 2021-07-01
Update payout date wording

tsmartt:
Screen Shot 2021-06-30 at 6 24 04 PM

Screen Shot 2021-07-01 at 4 34 06 PM

### DIFF
--- a/app/helpers/publishers_helper.rb
+++ b/app/helpers/publishers_helper.rb
@@ -51,8 +51,8 @@ module PublishersHelper
   end
 
   def next_deposit_date(today: DateTime.now)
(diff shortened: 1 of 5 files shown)


yachtcaptain23 (approved):
I would prefer putting `t(".payout_moved_to_thirteenth")` as an alert rather than in-line. But this is fine for now.

tsmartt:
> I would prefer putting `t(".payout_moved_to_thirteenth")` as an alert rather than in-line. But this is fine for now.

We need to add support for storing dismissals. Or we can use one of the many gems available if we have to add a lot of them later.

PR #3320, opened 2021-07-06
Rotate attr_encrypted

tsmartt:
In order to do so, in the secrets.yml file, set the current
attr_encrypted_key old field to attr_encrypted_key_old and then
put the new key in attr_encrypted_key. Then run `rake rekey`.

Note that we bulk update and we have support for reading the value
with the old key while rekeying. This requires a monkey-patch
which I think we can remove from the code once rekeying is done.

Note that now we pass in the key: to encryption_key for all the
models since some of them use the old slice method on the keys
and some use the new pack. Getting them all on the same method
is out of the scope of this PR. Note also that I've made all
of the encryption_key methods class methods for consistency
and so that we don't have to repeat code unnecessarily,
espcially now that we pass the key in.

If we ever want to rekey again, we have to reset all the columns to be false,
re-enable the monkey-patch to read values using the old key as a fallback,
and then rerun the rake task.


app/models/daily_metric.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
```
yachtcaptain23, line 1: We actually can drop this table, but let's do it in another PR.

db/migrate/20210706181815_add_rekeyed_field_to_encrypted_fields.rb
```diff
@@ -0,0 +1,16 @@
+class AddRekeyedFieldToEncryptedFields < ActiveRecord::Migration[6.1]
```
yachtcaptain23, line 1: My concern is that this migration will probably freeze our database and it's adding another unnecessary column to each model, and we could refactor this to not need to add any columns at all. My suggestion is that you could check against the `updated_at` timestamp

tsmartt: I didn't think that would provide enough resolution, i.e. what if something else updated the record recently? Then we wouldn't rekey it which would lead to unreadable data. 

The performance issue is gone in PG 11 https://dba.stackexchange.com/a/211222

tsmartt: This is removed in favor of retrying from the beginning and skipping if already rekeyed.

lib/tasks/rekey_encrypted_fields.rake
```diff
@@ -0,0 +1,79 @@
...
+        end
+      end
+    end
+  end
+end
+
+def perform_sql_update(records:, field:)
+  # Using postgres UPDATE FROM syntax
+  # https://www.postgresql.org/docs/current/sql-update.html
+  rekeyed_field_name = "#{field}_rekeyed"
```
yachtcaptain23, line 63: Would you update this part since we no longer use this column?

tsmartt: Done.

lib/tasks/rekey_encrypted_fields.rake
```diff
@@ -0,0 +1,77 @@
...
+  new_key = Rails.application.secrets[:attr_encrypted_key]
+
+  hash_model_to_columns = Util::AttrEncrypted.get_all_encrypted_fields
+
+  ActiveRecord::Base.connection.truncate(UserAuthenticationToken.table_name)
+
+  hash_model_to_columns.except(UserAuthenticationToken).each do |model, encrypted_fields|
+    encrypted_fields.each do |field|
+      puts "Rekeying #{model} #{field}"
+      model.find_in_batches(batch_size: 10000) do |batch_of_records|
```
yachtcaptain23, line 13: Would you be able to sort this so we'll know where to restart from if this function fails mid-way rather than starting from a seemingly random point?

app/services/util/attr_encrypted.rb
```diff
@@ -0,0 +1,72 @@
+class Util::AttrEncrypted
+  # taken from https://github.com/attr-encrypted/attr_encrypted/issues/314
+  # Rekey (change the key of) a given field for an object, given the old key.
+  # This does not SAVE the user data - do a save afterwards if you want that!
+  # This will raise an exception if the old key doesn't work.
+  def self.rekey(object:, field:, old_key:, new_key:, field_value:)
+    field_name_encrypted = object.send("encrypted_#{field}")
+    field_name_iv = object.send("encrypted_#{field}_iv")
```
yachtcaptain23, line 8: Do we need to call `Base64.decode64(` on this field name?

yachtcaptain23: Nevermind, this logic seems fine

lib/tasks/rekey_encrypted_fields.rake
yachtcaptain23: To add on, it's preferable that we store the value into Redis so we know the last executed object and ID

tsmartt: I can add an ordering, sure, but the intention is just to rerun it as it should be quick enough.

tsmartt: Done.

yachtcaptain23 (approved):

PR #3334, opened 2021-07-21
Add a working Uphold Oauth Refresh Service

tsmartt:
This enables support for refreshing auth and refresh tokens for Uphold Connections. 
This still needs to be enabled for production, and we are waiting on them for the batch of first refresh tokens.

Todo: 

- [x] Add support for importing the initial refresh tokens
- [x] Add task for refreshing tokens monthly, so it can be run outside of the potential payments generation.


Gemfile
```diff
@@ -9,6 +9,9 @@ gem "rack", "~> 2.1.0"
 # Serialize models for JSON APIs
 gem "active_model_serializers", "~> 0.10.0"
 
+# For bulk updates/imports
+gem "activerecord-import", "1.2.0"
```
yachtcaptain23, line 13: Are we able to avoid this gem? https://blog.saeloun.com/2019/11/26/rails-6-insert-all.html indicates that `activerecord-import` is no longer necessary since we run Rails 6.

tsmartt: I thought about that too, but this gem is quite nice and adds a lot of functionality that Rails doesn't support. Will be useful in the future, too.

tsmartt: For ex: mentioned, [here](https://github.com/zdennis/activerecord-import/issues/604), people switched to the built-in functionality and then switched back since the gem has better functionality.

app/jobs/sync/launch_token_refresh_jobs_job.rb
```diff
@@ -0,0 +1,13 @@
+class Sync::LaunchTokenRefreshJobsJob
+  include Sidekiq::Worker
+
+  def perform
+    UpholdConnection.with_details.all.each do |uphold_connection|
```
yachtcaptain23, line 5:
```suggestion
    UpholdConnection.with_details.find_each do |uphold_connection|
```

The `all` forces a fetch from the database whereas `find_each` properly batches for `each`.

config/sidekiq.yml
```diff
@@ -70,6 +70,10 @@
     args: ['send_notifications']
     description: "Notifies publishers their wallet is disconnected on the 23rd of every month."
     queue: scheduler
+  Sync::LaunchTokenRefreshJobsJob:
```
yachtcaptain23, line 73: I'm not a fan of this since this seems to generate a lot of unnecessary requests. We had something similar to this in the past and I believe we got rid of it.

tsmartt: Removed, we'll just refresh on login and update the UI

app/services/uphold/models/refreshment.rb
```diff
@@ -0,0 +1,49 @@
...
+        request = Net::HTTP::Post.new(uri)
+        request.basic_auth(Rails.application.secrets[:uphold_client_id], Rails.application.secrets[:uphold_client_secret])
+        request.content_type = "application/x-www-form-urlencoded"
+        request.set_form_data(
+          "grant_type" => "refresh_token",
+          "refresh_token" => refresh_token,
+        )
+
+        req_options = {
+          use_ssl: uri.scheme == "https",
```
diracdeltas, line 40: would this ever get posted over plaintext http?

tsmartt: No, all interactions with this endpoint should come through here which specified https.

diracdeltas: in that case can this just be true?

tsmartt: Done.

app/models/uphold_connection.rb
```diff
@@ -50,6 +50,10 @@ class UpholdAccountState
       where("updated_at < ?", UPHOLD_CODE_TIMEOUT.ago)
   }
 
+  scope :with_details, -> {
+    where.not(encrypted_uphold_access_parameters: nil)
+  }
+
```
yachtcaptain23, line 56:
```suggestion
```

app/services/uphold/models/refreshment.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'net/http'
+require 'uri'
+
+module Uphold
+  module Models
+    class Refreshment < Client
```
yachtcaptain23, line 8: Would you clarify naming, e.g. rename `def refresh` to `def refresh_authorization` and rename `Refreshment` to `Authorization`?

app/services/uphold/refresher.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Uphold
+  class Refresher
+    REFRESH_TOKEN = "refresh_token"
```
yachtcaptain23, line 5:
```suggestion
``` I don't think this is being used.

File: app/models/uphold_connection.rb [hunk 8]
```diff
@@ -207,6 +215,7 @@ def sync_connection!
     # Set uphold_details to a variable, if uphold_access_parameters is nil
     # we will end up makes N service calls everytime we call uphold_details
     # this is a side effect of the memoization
+    Uphold::Refresher.build.call(uphold_connection: self)
```
yachtcaptain23, line 218: Please only call this if the access token is expired.

tsmartt: Telling if it is expired is not straightforward as we only get back the number of seconds the access token is good for (3600). But we don't store the last time that we updated this field, so the comparison would involve adding a new column. Since this is not expected to be called very often, I think it's fine to leave it as is.

app/services/uphold/refresher.rb
tsmartt: Done.

app/services/uphold/models/refreshment.rb
tsmartt: Done.

app/models/uphold_connection.rb
tsmartt: Done.

app/models/uphold_connection.rb
tsmartt: I'll have to add a new key to the JSON which we generate ourselves, that is, the expiration time.

app/models/uphold_connection.rb
```diff
@@ -202,11 +202,21 @@ def valid_card?
     false
   end
 
+  def refresh_token
+    JSON.parse(uphold_access_parameters || '{}').try(:[], 'refresh_token')
```
yachtcaptain23, line 206:
```suggestion
    JSON.parse(uphold_access_parameters || '{}').fetch('refresh_token', nil)
```

Nit: I'm not opposed to using what you're suggested, but I think other parts of the code uses `.fetch` over `.try`.

app/models/uphold_connection.rb
```diff
@@ -202,11 +202,21 @@ def valid_card?
...
+  end
+
+  def authorization_expires_at
+    exp_date = JSON.parse(uphold_access_parameters || '{}').try(:[], 'expiration_date')
+    exp_date.to_datetime if exp_date
```
yachtcaptain23, line 211:
```suggestion
    JSON.parse(uphold_access_parameters || '{}').fetch('expiration_date', nil)&.to_datetime
```

app/services/uphold/refresher.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      return if refresh_token.blank?
+      return if uphold_connection.authorization_expires_at&. > Time.zone.now
+
+      authorization = @impl_refresher.refresh_authorization(uphold_connection)
+
+      # add expiration time
+      authorization_hash = JSON.parse(authorization)
+
+      authorization_hash["expiration_date"] = authorization_hash["expires_in"].to_i.seconds.from_now.to_s
```
yachtcaptain23, line 26: Would you make `expiration_date` a constant? Also, would you call it `expiration_time`?

app/services/uphold/refresher.rb
```diff
@@ -0,0 +1,35 @@
...
+      @impl_refresher = impl_refresher
+    end
+
+    # Makes a request to the Uphold API to refresh the current access_token
+    def call(uphold_connection:)
+      # Ensure we have an refresh_token.
+      refresh_token = uphold_connection.refresh_token
+
+      return if refresh_token.blank?
+      return if uphold_connection.authorization_expires_at&. > Time.zone.now
```
yachtcaptain23, line 19: Would you create a method called `uphold_connection.authorization_expired?`? In case we need to re-use the same method.

app/services/uphold/refresher.rb
tsmartt: I can rename it, but it's a value we receive from uphold, so it should stay a var.

app/models/uphold_connection.rb
tsmartt: Sure

app/models/uphold_connection.rb
tsmartt: Sure, no problem. Best to be consistent.

app/services/uphold/refresher.rb
tsmartt: Sure.

tsmartt: Done.

app/models/uphold_connection.rb
tsmartt: Done.

app/models/uphold_connection.rb
tsmartt: Done.

app/services/uphold/refresher.rb
tsmartt: Renamed.

app/models/uphold_connection.rb
```diff
@@ -202,11 +202,24 @@ def valid_card?
...
+  def refresh_token
+    JSON.parse(uphold_access_parameters || '{}').fetch('refresh_token', nil)
+  end
+
+  def authorization_expires_at
+    JSON.parse(uphold_access_parameters || '{}').fetch('expiration_time', nil)&.to_datetime
+  end
+
+  def authorization_expired?
+    authorization_expires_at&. > Time.zone.now
```
yachtcaptain23, line 214:
```suggestion
    authorization_expires_at.present? && authorization_expires_at > Time.zone.now
```

This should always guarantee a boolean return type

app/services/uphold/refresher.rb
```diff
@@ -0,0 +1,35 @@
...
+      authorization = @impl_refresher.refresh_authorization(uphold_connection)
+
+      # add expiration time
+      authorization_hash = JSON.parse(authorization)
+
+      authorization_hash["expiration_time"] = authorization_hash["expires_in"].to_i.seconds.from_now.to_s
+
+      # Update with the latest Authorization
+      uphold_connection.uphold_access_parameters = JSON.dump(authorization_hash)
+      uphold_connection.save!
```
yachtcaptain23, line 30:
```suggestion
      uphold_connection.update(uphold_access_parameters: authorization_hash.to_json)
```

lib/tasks/database_updates/add_initial_uphold_refresh_tokens.rake
```diff
@@ -0,0 +1,44 @@
...
+    def handle_uphold
+      access_token_to_refresh_token = {}
+      CSV.foreach("initial_uphold_tokens.csv", headers: true) do |row|
+        access_token_to_refresh_token[row["accessToken"]] = row["refreshToken"]
+      end
+
+      expiration_time = 1.hour.seconds.from_now.to_s
+
+      records_to_update = []
+      UpholdConnection.find_each do |uphold_connection|
```
yachtcaptain23, line 17:
```suggestion
      UpholdConnection.where.not(encrypted_uphold_access_parameters: nil).order(id: :asc).find_each do |uphold_connection|
```

lib/tasks/database_updates/add_initial_uphold_refresh_tokens.rake
```diff
@@ -0,0 +1,44 @@
...
+        access_params = uphold_connection.uphold_access_parameters
+        if access_params
+          parsed_access_params = JSON.parse(access_params)
+          refresh_token = access_token_to_refresh_token[parsed_access_params["access_token"]]
+          if refresh_token
+            parsed_access_params["refresh_token"] = refresh_token
+            parsed_access_params["expiration_time"] = expiration_time
```
yachtcaptain23, line 24:
```suggestion
          if access_token_to_refresh_token.include?(parsed_access_params["access_token"])
            parsed_access_params["refresh_token"] = access_token_to_refresh_token[parsed_access_params["access_token"]]
            parsed_access_params["expiration_time"] = expiration_time
```

lib/tasks/database_updates/add_initial_uphold_refresh_tokens.rake
tsmartt: find_each ignores ordering

tsmartt:
@yachtcaptain23 should be ready.

yachtcaptain23 (approved):

PR #3335, opened 2021-07-21
Wasmer image

tsmartt:
Let's us resize images using wasmer-thumbnail.

Also eliminates our monkey-patching of Active Storage since 6.1 supports different services per attachment.


app/helpers/image_conversion_helper.rb
```diff
@@ -2,26 +2,80 @@ module ImageConversionHelper
...
+    temp_file.write(actual_jpeg_hex_to_save)
+    temp_file.path
+  end
 
-    mini_magick_image.write(temp_file)
+  def resize_image_with_wasm(file_bytes:, dimensions:, size: 250000)
+    require "wasmer"
+    store = Wasmer::Store.new
+    # Let's compile the module to be able to execute it!
+    module_ = Wasmer::Module.new store, IO.read("#{Rails.root}/wasm_thumbnail.wasm", mode: "rb")
```
yachtcaptain23, line 32: I think we should make this a global object that's initialized in the initializers folder rather than creating a new object every time.

evq: better yet - let's create a wrapper gem that packages the wasm file and creates the global

evq (changes requested):
we can't merge this as is because it removes the image padding which IIRC is required per the privacy model for site banner image - will follow up out of band on best path forward

app/helpers/image_conversion_helper.rb
tsmartt: I've at least moved the compilation out to the module level. We can leave the gem for a later exercise.

tsmartt:
The padding issue has been fixed.

app/helpers/image_conversion_helper.rb
tsmartt: I ran into issues moving the instance itself into a global, weird cryptic ruby errors, still tracking down. For now, I've moved the compilation to the module level and keep the instance in the method.

tsmartt: I've got a local rubygem, will push it up later.

tsmartt:
The last step is to add the rubygem, which I have locally. Will push Monday morning.

app/helpers/image_conversion_helper.rb
tsmartt: Gem here https://github.com/brave-intl/wasm-thumbnail/pull/2

tsmartt:
Gem here https://github.com/brave-intl/wasm-thumbnail/pull/2

app/helpers/image_conversion_helper.rb
```diff
@@ -1,85 +1,37 @@
...
       LogException.perform(StandardError.new("Unknown attachment type:" + attachment_type), params: {})
       return nil
     end
 
-    # Use JPG as its the most efficient standard
-    mini_magick_image.format "jpg"
+    actual_jpeg_to_save = Wasm::Thumbnail::Rb.resize_and_pad(file_bytes: file_bytes,
+                                                             width: dimensions[0],
+                                                             height: dimensions[1],
+                                                             size: 250000)
```
evq, line 24: do we want this to be `250000 + 4`? since the output buffer would be `250000` bytes before we chop the first 4 off containing the length

config/storage/production.yml
```diff
@@ -0,0 +1,16 @@
+# For case notes so far, the default bucket option
+amazon_case_notes:
```
yachtcaptain23, line 2: I think we should use a better name, like `amazon_internal_bucket` for this one and `amazon_public_bucket` for the other.

app/helpers/image_conversion_helper.rb
tsmartt: The gem takes care of this now.

config/storage/production.yml
tsmartt: Done.

tsmartt:
OK thanks for the rereviews, should be ready to merge now.

evq (approved):

yachtcaptain23 (changes requested):
Messaging privately

app/helpers/image_conversion_helper.rb
```diff
@@ -21,20 +21,35 @@ def resize_to_dimensions_and_convert_to_jpg(source_image_path:, attachment_type:
...
+                                       quality: quality)
+  rescue RuntimeError
+    raise t("banner.upload_too_big") if quality <= 15
+
+    # Try with reduced quality, start with 50, go down by 5 each time
+    retry_quality(file_bytes: file_bytes,
+                  width: width,
+                  height: height,
+                  size: size,
+                  quality: quality - 5)
```
yachtcaptain23, line 50: quality is staying at 50, I think this will endlessly loop

app/helpers/image_conversion_helper.rb
```diff
@@ -21,20 +21,35 @@ def resize_to_dimensions_and_convert_to_jpg(source_image_path:, attachment_type:
...
-                                                               height: dimensions[1],
-                                                               size: target_file_size)
-    rescue RuntimeError
-      raise t("banner.upload_too_big")
-    end
+    actual_jpeg_to_save = retry_quality(file_bytes: file_bytes,
+                                        width: dimensions[0],
+                                        height: dimensions[1],
+                                        size: target_file_size,
+                                        quality: 50)
```
yachtcaptain23, line 28: you could start at 80 or even 100

app/helpers/image_conversion_helper.rb
```diff
@@ -21,20 +21,35 @@ def resize_to_dimensions_and_convert_to_jpg(source_image_path:, attachment_type:
...
+
     new_filename = filename + "_resized"
     temp_file = Tempfile.new([new_filename, ".jpg"], binmode: true)
     temp_file.write(actual_jpeg_to_save)
     temp_file.path
   end
 
+  def retry_quality(file_bytes:, width:, height:, size:, quality:)
```
yachtcaptain23, line 36: Would you move this away from recursive? This could use quite a bit of space because of recursion, as worst case it's 10*file_bytes

config/environments/staging.rb
```diff
@@ -62,18 +62,19 @@
...
+      url: Rails.application.secrets[:redis_url],
+      connect_timeout: 30, # Defaults to 20 seconds
+      read_timeout: 5, # Defaults to 1 second
+      write_timeout: 10, # Defaults to 1 second
+
+      error_handler: -> (method:, returning:, exception:) {
+                       # Report errors to Sentry as warnings
+                       Raven.capture_exception exception, level: 'warning',
+                                                          tags: { method: method, returning: returning }
+                     },
```
yachtcaptain23, line 76:
```suggestion
                     }
```

test/helpers/image_conversion_helper_test.rb
```diff
@@ -27,69 +28,6 @@ class ImageConversionHelperTest < ActionView::TestCase
...
-      source_image_path: temp_file.path,
-      attachment_type: SiteBanner::LOGO,
-      filename: "brave_lion_logo"
-    )
-
-    add_padding_to_image(
-      source_image_path: temp_file_path,
-      attachment_type: SiteBanner::LOGO,
-      quality: ImageConversionHelper::IMAGE_QUALITY
-    )
```
yachtcaptain23, line 48: Would you re-add a test to make sure file size output matches as expected?

config/environments/staging.rb
tsmartt: This was rubocop autocorrect.

app/helpers/image_conversion_helper.rb
tsmartt: Done

test/helpers/image_conversion_helper_test.rb
tsmartt: Done

app/helpers/image_conversion_helper.rb
```diff
@@ -1,85 +1,62 @@
...
+    quality_to_try = quality
+
+    while quality_to_try > 15
+      begin
+        return Wasm::Thumbnail::Rb.resize_and_pad(
+          file_bytes: file_bytes,
+          width: width,
+          height: height,
+          size: size,
+          quality: quality
```
yachtcaptain23, line 46:
```suggestion
          quality: quality_to_try
```

tsmartt: I pushed the change.

test/controllers/publishers/site_banners_controller_test.rb
```diff
@@ -38,27 +38,31 @@ class PublishersControllerTest < ActionDispatch::IntegrationTest
     site_banner = site_banners(:default)
     sign_in publisher
 
-    fake_data = "A" * Publishers::SiteBannersController::MAX_IMAGE_SIZE
+    fake_data = "A" * SiteBanner::LOGO_UNIVERSAL_FILE_SIZE
```
yachtcaptain23, line 41:
```suggestion
    fake_data = "A" * Publishers::SiteBannersController::MAX_IMAGE_SIZE
```

yachtcaptain23: This shouldn't the test's result, but this matches what the controller is checking

tsmartt: Fixed.

yachtcaptain23 (approved):

PR #3338, opened 2021-07-22
Limit attempts against a single user when trying to sign-in via totp

yachtcaptain23:
Closes https://github.com/brave-intl/creators-private-issues/issues/17


diracdeltas:
Any update here?

tsmartt (approved):

PR #3339, opened 2021-07-23
Add solargraph for IDEs which use LSP

tsmartt:


yachtcaptain23:
Would you instead add solargraph to your local Gemfile? https://phili.pe/posts/loading-additional-ruby-gems-in-development/ seems to have pretty good instructions

PR #3346, opened 2021-07-28
Use NULLDB to compile assets when no DB available

tsmartt:
The monkey patch we introduced for rotating attr-encrypted fields
causes the deploy to fail since when we go to precompile assets,
we need to connect to the DB. There is no DB at that point, so use
a fake one.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -201,6 +201,8 @@ group :development, :test do
   gem 'selenium-webdriver', "~> 3.142.0"
   gem 'solargraph'
   gem 'dotenv-rails', "2.7.6"
+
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -201,6 +201,8 @@ group :development, :test do
   gem 'selenium-webdriver', "~> 3.142.0"
   gem 'solargraph'
   gem 'dotenv-rails', "2.7.6"
+
+  gem 'activerecord-nulldb-adapter'
```
yachtcaptain23, line 205: Would you lock to a version?

tsmartt: Done.

yachtcaptain23 (approved):

PR #3347, opened 2021-07-28
DO NOT MERGE: Run as limited user

yachtcaptain23:


db/schema.rb
```diff
@@ -14,7 +14,6 @@
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gin"
-  enable_extension "pg_stat_statements"
```
tsmartt, line 17: Why this removal?

Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM ruby:2.7-slim
+FROM ruby:2.7.3
```
tsmartt, line 1: We need to keep the non-specific version of the container, as recommended by security.

docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
   web: &web_base
     build:
       context: .
-      dockerfile: Dockerfile.dev
```
tsmartt, line 7: The main dockerfile runs in the production environment, compiles assets, etc. We should keep the dev file.

tsmartt (changes requested):

docker-compose.yml
yachtcaptain23: Do you make use of the Dockerfile.dev? If not, my impression try to approximate our Dockerfile builds close to staging/production

tsmartt: Hah, only recently when having to change it. If you're doing development in Docker, I think it makes sense to. That's what compose is used for, so I think we should keep it. Also, more gems are installed so more dependencies are needed. AFAIK no one develops using it, but they should be kept together. I have some local changes/fixes to push up for it in any case.

Dockerfile
yachtcaptain23:
```suggestion
FROM ruby:2.7-slim
```

Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM ruby:2.7-slim
```
hspencer77, line 2: Is there a particular reason we don't use say `2.7-alpine` instead?  Based upon image scans, `2.7-slim` contains more vulnerabilities than `2.7-alpine`:

#### 2.7-slim scan

Screen Shot 2021-07-29 at 2 00 07 PM

#### 2.7-alpine

Screen Shot 2021-07-29 at 1 53 21 PM

hspencer77 (changes requested):
Some minor changes to help with making the image smaller.  If we could leverage `ruby:2.7-alpine` here we could cut down image size (`2.7-slim` is `146.11 MB` versus `2.7-alpine` is `48.95 MB`), and minimizes vulnerability footprint.  In addition, adding a multi-stage build would help out with image size as well.

docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
...
+      dockerfile: Dockerfile
```
yachtcaptain23, line 7:
```suggestion
      dockerfile: Dockerfile.dev
```

Dockerfile
yachtcaptain23: Using alpine. Thanks!

Dockerfile
```diff
@@ -20,7 +22,6 @@ RUN gem install bundler
 RUN NODE_ENV=production
```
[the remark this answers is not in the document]
yachtcaptain23: Setting it in the docker-compose file: https://github.com/brave-intl/creators-ops/pull/185

RAILS_ENV is already set

PR #3350, opened 2021-08-02
Don't decrypt if field is null

tsmartt:


app/services/util/attr_encrypted.rb
```diff
@@ -41,11 +41,13 @@ def self.get_all_encrypted_fields
   def self.get_value_using_key(record:, field:, key:)
     field_name_iv = record.send("encrypted_#{field}_iv")
     field_name_encrypted = record.send("encrypted_#{field}")
-    iv = Base64.decode64(field_name_iv)
-    record.class.send("decrypt_#{field}",
-                      field_name_encrypted,
-                      iv: iv,
-                      key: record.class.encryption_key(key: key))
+    if field_name_iv
```
yachtcaptain23, line 44:
```suggestion
    return nil if field_name_iv
```

There's an edge case where `field_name_encrypted` might not be nil, in which it'll returned the encrypted field.

tsmartt: Good catch. I fixed it another way and added a test.

yachtcaptain23 (approved):

PR #3358, opened 2021-08-07
Have docker cache wasmer before installing

tsmartt:
Wasmer has a lot of trouble properly installing
in our Docker environment when we are using deployment
mode. If not deploying, the asset compile can't find
it. So we use deployment mode, but have to cache the wasmer
gem before we do that or we get a git error due, I suspect
due to the gem not being formed properly, but wasn't able to
ascertain the exact reason.


Dockerfile
```diff
@@ -35,6 +37,9 @@ COPY package.json yarn.lock .nvmrc ./
 
 # Install the dependencies.
 RUN nvm install && nvm use
+RUN gem install wasmer
+RUN bundle package --all
+RUN bundle config set --local deployment 'true'
```
yachtcaptain23, line 42: This dockerfile is run in prod. Why do we need this line?

Dockerfile
```diff
@@ -35,6 +37,9 @@ COPY package.json yarn.lock .nvmrc ./
 
 # Install the dependencies.
 RUN nvm install && nvm use
+RUN gem install wasmer
```
yachtcaptain23, line 40:
```suggestion
RUN gem install wasmer -v 1.0.0
```

yachtcaptain23 (changes requested):

Dockerfile
tsmartt: This is needed to set deployment mode which will vendor the gems. The local flag is to tell bundler to check the local cache (where we installed wasmer) and use that. This gets around the bundler error we are currently seeing.

PR #3360, opened 2021-08-11
Completely remove imagemagick

tsmartt:
Don't need it anymore, use `fastimage` in the tests to get the dimensions.


tsmartt:
Ah, good point! Will update both.

tsmartt:
@evq done.

evq (approved):

PR #3375, opened 2021-08-30
Bulk insert potential payments

tsmartt:
Initial stab at simplifying potential payments
We want to bulk insert these for performance reasons. We no longer send
out emails and with this, we will no longer sync connection information
for each of the wallets. That is done when a user logs in.

This lets us delete and simplify the code. I'm sure there are things
that can still be done, and might be -- but this is a first stab.

Things to be done for this PR before merge:

- [x] tests pass
- [x] Make sure all records loaded up front, so no n+1

Things that still might be done, but maybe for v2:
- [ ] Eliminate the create_message calls and just return the objects
   for bulk insertion
- [x] Make uphold_service and gemini_service style like bitflyer_service


app/models/publisher.rb
```diff
@@ -112,8 +112,7 @@ class Publisher < ApplicationRecord
   ###############################
 
   scope :uphold_selected_provider, -> {
-    joins("INNER JOIN uphold_connections
-           ON uphold_connections.id = publishers.selected_wallet_provider_id
+    where("uphold_connections.id = publishers.selected_wallet_provider_id
```
yachtcaptain23, line 115: Unless I'm mistaken, scopes like these should still work in isolation.

yachtcaptain23:
Would you make it so that the code is idempotent/retryable? If the primary jobs fails part way, it doesn't seem like there's a way to restart from the middle

app/models/publisher.rb
tsmartt: Yeah these are only ever called from our job, and with eager loading it throws a double reference error to have the join here. I'll work on a workaround.

tsmartt: Fixed.

yachtcaptain23: So the confusing part is that below, you're using both `includes` and `joins`. Is it possible to revert back to the original definition of `uphold_selected_provider`, and if not, why?

app/services/payout/service.rb
```diff
@@ -2,23 +2,17 @@ module Payout
...
     end
 
     def create_message(message)
-      return if should_only_notify?
-
       PayoutMessage.create(
         payout_report: @payout_report,
         publisher: @publisher,
-        message: "#{self.class.name} #{message}"
+        message: "#{@classs_name} #{message}"
```
yachtcaptain23, line 15: Misspelling
```suggestion
        message: "#{@class_name} #{message}"
```

app/services/payout/gemini_service.rb
```diff
@@ -1,62 +1,52 @@
 module Payout
-  class GeminiService < Service
-    def perform
-      # Writes a message if we should skip a publisher or not.
-      # Implemented in Payout::Service
-      return if skip_publisher?
+  class GeminiService
+    def perform(payout_report:, publisher:)
+      return [] if Payout::Service.new(class_name: self,
```
yachtcaptain23, line 4:
```suggestion
      return [] if Payout::Service.new(class_name: self.name,
```

app/services/payout/gemini_service.rb
```diff
@@ -1,62 +1,52 @@
 module Payout
-  class GeminiService < Service
-    def perform
-      # Writes a message if we should skip a publisher or not.
-      # Implemented in Payout::Service
-      return if skip_publisher?
+  class GeminiService
+    def perform(payout_report:, publisher:)
```
yachtcaptain23, line 3: These were part of the same `Payout::Service` class to use the same initializer so we didn't have to repeat ourselves, may we revert back?

app/models/publisher.rb
tsmartt: You need to use `includes` for eager loading since we enforce strict loading, but we need to `joins` for filtering.

app/services/payout/gemini_service.rb
tsmartt: This is to bring these in line with the `Bitflyer Service` class for consistency. The methods in that class are utility classes and can be called as such; there's no need to bind these classes together in a hierarchy -- they can each just call the helper. Having it use inheritance causes a lot of pain when trying to add the `Bitflyer Service` class due to the complexity in that class, so we began refactoring it then. This is the completion of that work.

app/jobs/enqueue_publishers_for_payout_job.rb
```diff
@@ -2,63 +2,101 @@
...
-      manual: manual,
-      publisher_ids: publisher_ids
-    }.to_json
-    Payout::UpholdJob.perform_async(json_args)
+  def enqueue_payout(payout_report:, manual:, publisher_ids:)
+    base_publishers = Publisher.strict_loading.includes(
+      :channels,
+      :status_updates,
+      :user_authentication_token
+    )
```
yachtcaptain23, line 33:
```suggestion
    base_publishers = Publisher.strict_loading.includes(
      :channels
    )
```

tsmartt: Done.

app/services/payout/service.rb
```diff
@@ -2,23 +2,17 @@ module Payout
...
     end
 
     def create_message(message)
-      return if should_only_notify?
-
       PayoutMessage.create(
         payout_report: @payout_report,
         publisher: @publisher,
-        message: "#{self.class.name} #{message}"
+        message: "#{@class_name} #{message}"
```
yachtcaptain23, line 15: Rather than passing by argument, the child class inherits this function so refactoring was unnecessary.

tsmartt: I prefer composition, you inheritance here. I'll make the change.

yachtcaptain23 (approved):

Issue #3387, opened 2021-09-20
TwoFactorAuthenticationRemovalJob could put user in a double-locked status


yachtcaptain23:
The offending block is found in 
```
ActiveRecord::Base.transaction do
  if publisher.locked?
    # Status_updates is sorted by entry, first is "locked", second is the status the publisher was in before
    previous_status = publisher.status_updates&.second&.status || PublisherStatusUpdate::ACTIVE
    publisher.status_updates.create(status: previous_status)
  end

  two_factor_authentication_removal.destroy
end
``` 
where `previous_status` could be `locked` if there is a double entry. We should put a fix here and also make sure `locked` doesn't get inserted twice when we put the user in a locked state.

PR #3390, opened 2021-09-20
Fix double locked publisher status

yachtcaptain23:
Fix an issue where a user could be put in a permanent "locked" state after resetting their 2FA twice.

Closes https://github.com/brave-intl/publishers/issues/3387

### DIFF
--- a/app/jobs/two_factor_authentication_removal_job.rb
+++ b/app/jobs/two_factor_authentication_removal_job.rb
@@ -32,8 +32,7 @@ class TwoFactorAuthenticationRemovalJob < ApplicationJob
         # Return publisher to previous state.
         ActiveRecord::Base.transaction do
           if publisher.locked?
-            # Status_updates is sorted by entry, first is "locked", second is the status the publisher was in before
-            previous_status = publisher.status_updates&.second&.status || PublisherStatusUpdate::ACTIVE
+            previous_status = publisher.status_updates.select { |status_update| status_update.status != PublisherStatusUpdate::LOCKED }.first.status
             publisher.status_updates.create(status: previous_status)
           end
 
--- a/lib/tasks/database_updates/reset_permalocked_users.rake
+++ b/lib/tasks/database_updates/reset_permalocked_users.rake
@@ -1,0 +1,18 @@
+namespace :database_updates do
+  desc 'Reset Permalocked users'
+  task :reset_permalocked_users => :environment do
+    publisher_ids = []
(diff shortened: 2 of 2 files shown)


app/jobs/two_factor_authentication_removal_job.rb
```diff
@@ -32,8 +32,7 @@ def perform
         # Return publisher to previous state.
         ActiveRecord::Base.transaction do
           if publisher.locked?
-            # Status_updates is sorted by entry, first is "locked", second is the status the publisher was in before
-            previous_status = publisher.status_updates&.second&.status || PublisherStatusUpdate::ACTIVE
+            previous_status = publisher.status_updates.select { |status_update| status_update.status != PublisherStatusUpdate::LOCKED }.first.status
```
tsmartt, line 35: Was going to comment that we should keep the default in case there is a missing previous status, but if you're locked you must have had a previous one. If you don't, that's a bug and the data and code should be fixed.

lib/tasks/database_updates/reset_permalocked_users.rake
```diff
@@ -0,0 +1,18 @@
...
+    publisher_ids = []
+    PublisherStatusUpdate.select(:status, :publisher_id).where(status: "locked").group(:publisher_id, :status).having("count(*) > 1").each do |psu|
+      publisher_ids.append(psu.publisher_id)
+    end
+
+    publisher_ids -= TwoFactorAuthenticationRemoval.pluck(:publisher_id)
+
+    Publisher.includes(:status_updates).where(id: publisher_ids).find_each do |publisher|
+      if publisher.locked?
+        previous_status = publisher.status_updates.select { |status_update| status_update.status != PublisherStatusUpdate::LOCKED }.first.status
```
tsmartt, line 13: Are these ordered properly so that we want the first one? If we want the latest non-locked status, would we reverse the order and then choose first?

tsmartt (commented):
There was a question but overall looks good.

app/jobs/two_factor_authentication_removal_job.rb
yachtcaptain23: There's always a previous status.

lib/tasks/database_updates/reset_permalocked_users.rake
yachtcaptain23: Yes, it's sorted by default: https://github.com/brave-intl/publishers/blob/d5b83e4e7c015854ff2568ab4401c9b6f2c458cc/app/models/publisher.rb#L42

tsmartt (approved):

PR #3395, opened 2021-09-23
Eager load the polymorphic channel details

tsmartt:
This lets us eager load polymorphic associations for when we call `publication_title` in `gemini_service`, `uphold_service`, `bitflyer_service`.

Rails doesn't support polymorphic eager loading through `includes`, only though `preload`.

### DIFF
--- a/app/jobs/enqueue_publishers_for_payout_job.rb
+++ b/app/jobs/enqueue_publishers_for_payout_job.rb
@@ -27,8 +27,9 @@ class EnqueuePublishersForPayoutJob < ApplicationJob
 
   def enqueue_payout(payout_report:, manual:, publisher_ids:)
     base_publishers = Publisher.strict_loading.includes(
-      :channels,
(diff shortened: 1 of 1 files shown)


app/jobs/enqueue_publishers_for_payout_job.rb
```diff
@@ -27,8 +27,9 @@ def perform(final: true, manual: false, payout_report_id: "", publisher_ids: [],
 
   def enqueue_payout(payout_report:, manual:, publisher_ids:)
     base_publishers = Publisher.strict_loading.includes(
-      :channels,
       :status_updates,
+    ).preload(
+      channels: :details
```
yachtcaptain23, line 32:
```suggestion
      :status_updates,
      :user_authentication_token
    ).preload(
      channels: :details
```

yachtcaptain23 (approved):

PR #3396, opened 2021-09-24
Add gemini recipient IDs to channels

tsmartt:
Closes https://github.com/brave-intl/creators-private-issues/issues/20
Like the other wallet providers, we should send a deposit ID per channel for gemini. 

For this to work, we need the `create addresses` permission in Gemini.


app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -46,7 +46,7 @@ def generate_brotli_encoded_channel_response(prefix:)
           wallet = PublishersPb::Wallet.new
           gemini_wallet = PublishersPb::GeminiWallet.new
           gemini_wallet.wallet_state = get_gemini_wallet_state(gemini_connection: site_banner_lookup.publisher.gemini_connection)
-          gemini_wallet.address = site_banner_lookup.publisher.gemini_connection.recipient_id
+          gemini_wallet.address = site_banner_lookup.channel.gemini_recipient_id
```
yachtcaptain23, line 49: Can we revert this for now and add it in later? That way there's no issues for users during tipping (I assume it'll be nil when users fetch this publisher's list)

app/models/gemini_connection.rb
```diff
@@ -90,6 +95,12 @@ def sync_connection!
       recipient = Gemini::RecipientId.find_or_create(token: access_token)
       update(recipient_id: recipient.recipient_id)
       update_default_currency
+
+      # Add bitFlyer deposit id to each of the publisher's channels
+      publisher.channels.each do |channel|
+        channel_recipient = Gemini::RecipientId.find_or_create(token: access_token, label: channel.id)
+        channel.update_column(:gemini_recipient_id, channel_recipient.recipient_id)
+      end
```
yachtcaptain23, line 103: Some creators have hundreds of channels. May we think of a better way to do this?

tsmartt: I can make it a job, sure.

app/jobs/cache/browser_channels/responses_for_prefix.rb
tsmartt: Per private PM, going to default this to the deposit ID on the connection since channels might not yet have them.

app/jobs/create_gemini_recipient_ids_job.rb
```diff
@@ -0,0 +1,31 @@
+# Creates the Uphold Cards for a publisher
+class CreateGeminiRecipientIdsJob < ApplicationJob
```
yachtcaptain23, line 2: Please use Sidekiq. `ApplicationJob` doesn't retry

app/jobs/create_gemini_recipient_ids_job.rb
```diff
@@ -0,0 +1,31 @@
...
+  queue_as :default
+
+  def perform(gemini_connection_id:)
+    gemini_connection = GeminiConnection.find(gemini_connection_id)
+    # Users aren't able to create a recipient id if they are not fully verified
+    if gemini_connection.payable?
+      recipient = Gemini::RecipientId.find_or_create(token: gemini_connection.access_token)
+      gemini_connection.update(recipient_id: recipient.recipient_id)
+      gemini_connection.update_default_currency
```
yachtcaptain23, line 11:
```suggestion
```

If i recall correctly, there is no default currency for Gemini. @Miyayes would you confirm?

app/jobs/create_gemini_recipient_ids_job.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      gemini_connection.publisher.channels.each do |channel|
+        gemini_connection_for_channel = GeminiConnectionForChannel.where(
+          gemini_connection: gemini_connection,
+          currency: gemini_connection.default_currency,
+          channel_identifier: channel.details.channel_identifier
+        ).first_or_create(channel_id: channel.id)
+
+        channel_recipient = Gemini::RecipientId.find_or_create(token: gemini_connection.access_token, label: channel.id)
```
yachtcaptain23, line 20:
```suggestion
        channel_recipient = Gemini::RecipientId.find_or_create(token: gemini_connection.access_token, label: channel.details.brave_publisher_id)
```
My impression is that the label could be created on Gemini's server, similar to Uphold card's name.

test/models/gemini_connection_test.rb
```diff
@@ -47,43 +46,4 @@ class GeminiConnectionTest < ActiveSupport::TestCase
       assert_equal gemini_connection.access_token, 'km2bylijaDkceTOi2LiranELqdQqvsjFuHcSuQ5aU9jm'
     end
   end
-
-  describe '#sync_connection' do
```
yachtcaptain23, line 51: Out of curiosity, why are we removing this test?

app/jobs/create_gemini_recipient_ids_job.rb
tsmartt: That's fine as an identifier.

app/jobs/create_gemini_recipient_ids_job.rb
tsmartt: Confirmed we can remove it.

app/models/channel.rb
```diff
@@ -14,6 +14,7 @@ class Channel < ApplicationRecord
   SUBSCRIBER_COUNT = :subscriber_count
   ADVANCED_SORTABLE_COLUMNS = [YOUTUBE_VIEW_COUNT, TWITCH_VIEW_COUNT, VIDEO_COUNT, SUBSCRIBER_COUNT, FOLLOWER_COUNT].freeze
   BITFLYER_CONNECTION = "BitflyerConnection".freeze
+  GEMINI_CONNECTION = "GeminiConnection".freeze
```
yachtcaptain23, line 17: Nit: So I don't think we should store the polymorphic types in the models. I'd rather see the following to keep down code density.

```
if publisher.selected_wallet_provider_type == GeminiConnection.to_s
```

Thoughts? (This is a non-blocker)

app/jobs/create_gemini_recipient_ids_job.rb
```diff
@@ -0,0 +1,30 @@
...
+    # Users aren't able to create a recipient id if they are not fully verified
+    if gemini_connection.payable?
+      recipient = Gemini::RecipientId.find_or_create(token: gemini_connection.access_token)
+      gemini_connection.update(recipient_id: recipient.recipient_id)
+
+      gemini_connection.publisher.channels.each do |channel|
+        gemini_connection_for_channel = GeminiConnectionForChannel.where(
+          gemini_connection: gemini_connection,
+          channel_identifier: channel.details.channel_identifier
+        ).first_or_create(channel_id: channel.id)
```
yachtcaptain23, line 17:
```suggestion
        ).find_or_initialize_by(channel_id: channel.id)
```

yachtcaptain23: This saves on an insert because `update` a few lines down will insert if needed

db/migrate/20210929150807_remove_currency_from_gemini_connections.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveCurrencyFromGeminiConnections < ActiveRecord::Migration[6.1]
+  def change
+    remove_column :gemini_connections, :default_currency
+    remove_column :gemini_connection_for_channels, :currency
```
yachtcaptain23, line 4: Would you `currency` remove from the previous migration? That way we don't create a field and then immediately remove it.

db/migrate/20210929150807_remove_currency_from_gemini_connections.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveCurrencyFromGeminiConnections < ActiveRecord::Migration[6.1]
+  def change
+    remove_column :gemini_connections, :default_currency
```
yachtcaptain23, line 3: Can you make sure this won't break anything? If it's safer to leave, then we can leave it.

db/migrate/20210928154029_create_gemini_connection_for_channel.rb
```diff
@@ -0,0 +1,17 @@
...
+      t.belongs_to :channel, type: :uuid, index: true, null: false
+
+      t.string :currency, index: true
+      t.string :channel_identifier, index: true
+      t.string :recipient_id
+
+      t.timestamps
+    end
+
+    add_index :gemini_connection_for_channels, [:channel_identifier, :currency, :gemini_connection_id], unique: true, name: 'unique_gemini_connection_for_channels'
```
yachtcaptain23, line 14: This index needs to be updated to not include currency

db/migrate/20210929150807_remove_currency_from_gemini_connections.rb
yachtcaptain23: Actually I saw the method definition. LG2M

app/models/gemini_connection.rb
```diff
@@ -3,20 +3,21 @@
...
   validates :recipient_id, uniqueness: true, allow_blank: true
 
   attr_encrypted :access_token, :refresh_token, key: proc { |record| record.class.encryption_key }
 
-  after_save :update_default_currency, if: -> { saved_change_to_default_currency? }
-
-  validates :default_currency, inclusion: { in: SUPPORTED_CURRENCIES }, allow_nil: true
+  scope :payable, -> {
+    where(is_verified: true).
+      where(status: 'Active')
```
yachtcaptain23, line 19: Is there a case in which this goes from an Active state to non-Active?

app/jobs/create_gemini_recipient_ids_job.rb
```diff
@@ -0,0 +1,30 @@
+# Creates the Uphold Cards for a publisher
+class CreateGeminiRecipientIdsJob
+  include Sidekiq::Worker
+  sidekiq_options queue: :scheduler
+
+  def perform(gemini_connection_id:)
```
yachtcaptain23, line 6:
```suggestion
  def perform(gemini_connection_id)
```
I don't think keyword arguments are supported by Sidekiq (I could be wrong)

app/models/gemini_connection.rb
tsmartt: Yes, in the `sync_connection` code, Gemini can return to us a different status.

db/migrate/20210928154029_create_gemini_connection_for_channel.rb
tsmartt: Done.

db/migrate/20210929150807_remove_currency_from_gemini_connections.rb
tsmartt: Done.

app/models/channel.rb
tsmartt: Done.

app/jobs/create_gemini_recipient_ids_job.rb
tsmartt: Done.

yachtcaptain23 (approved):

db/migrate/20210928154029_create_gemini_connection_for_channel.rb
```diff
@@ -0,0 +1,16 @@
...
+  def change
+    create_table :gemini_connection_for_channels, id: :uuid, default: -> { "uuid_generate_v4()"} do |t|
+      t.belongs_to :gemini_connection, type: :uuid, index: true, null: false
+      t.belongs_to :channel, type: :uuid, index: true, null: false
+
+      t.string :channel_identifier, index: true
+      t.string :recipient_id
+
+      t.timestamps
+    end
```
yachtcaptain23, line 11:
```suggestion
    create_table :gemini_connection_for_channels, id: :uuid, default: -> { "uuid_generate_v4()"} do |t|
      t.belongs_to :gemini_connection, type: :uuid, null: false
      t.belongs_to :channel, type: :uuid, index: true, null: false

      t.string :channel_identifier
      t.string :recipient_id

      t.timestamps
    end
```

You'll have to regenerate the `db/schema.rb` file, but basically we have duplicate indices which adds space/performance overhead. Postgres has partial indices unlike MySQL

yachtcaptain23 (changes requested):
Comments in-line, need to regenerate `db/schema.rb`

yachtcaptain23 (approved):

PR #3402, opened 2021-09-27
Refresh the Uphold token if authorization expired

tsmartt:
We shouldn't skip refresh if we have an expired auth. That's the whole point of refresh.


tsmartt:
Merging since this is needed immediately.

PR #3425, opened 2021-10-04
Uphold don't erase 401

tsmartt:
When we check the connection details, we don't sync first, and then we wipe out the details that currently exist. This PR removed the wiping out of details and forces a sync before returning the details.


tsmartt:
Merging due to immediate need.

PR #3476, opened 2021-11-09
Add Wallet Syncer which Updates on Login

tsmartt:
We don't want to update on home page view, and we want to cover
all wallet providers. Currently we won't sync Bitflyer since I need to 
verify refreshing works properly and this is
mainly addressing Uphold, for the moment. Will need to come back and
turn that on for Bitflyer, though it's not as big of a priority since
at connection, the user is already KYCd.

### DIFF
--- a/app/services/util/wallet/connection_syncer.rb
+++ b/app/services/util/wallet/connection_syncer.rb
@@ -1,0 +1,30 @@
+# The hope is to make this the start of ripping out the
+# oauth connection functions out of the model and into
+# service classes
+class Util::Wallet::ConnectionSyncer
+
+  def self.build
+    new
+  end
+
+  def call(publisher:)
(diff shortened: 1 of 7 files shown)


porteron (approved):

app/services/util/wallet/connection_syncer.rb
```diff
@@ -0,0 +1,29 @@
...
+    new
+  end
+
+  def call(publisher:)
+    selected_wallet = publisher.selected_wallet_provider
+
+    # Could make this dynamic, but the intention is to add typing and make this a union type
+    # for exhaustiveness checking. Spelling it out to reduce magic in the codebase.
+    syncer = case selected_wallet
+    in UpholdConnection
```
porteron, line 15: I like the implementation of the connection syncer.

PR #3495, opened 2021-12-07
Serve the public cloudfront URL instead of the private

tsmartt:
When we migrated to ruby-wasmer, we removed an activestorage plugin we had written which essentially just sets the URL to a cloudfront URL. Restoring that functionality.


porteron (approved):
👍

PR #3505, opened 2021-12-17
Nightly report updates

tsmartt:
This is the data as the dashboard displays it, i.e., we mark a download fiinalized today, but the download_ts
is what we go off of when generating this report. This means the confirmation is backported 30+ days.
The next section will include the report where we group by the finalized_ts, but that's only included
_additionally_ in the nightly CSV since the dashboard still groups data by the download_ts. That is, we still need
a section in the nightly CSV that matches the data you see on the dashboard.


porteron (commented):
👍

porteron (approved):

PR #3515, opened 2022-01-03
Only run dependabot on code that runs in production

tsmartt:


diracdeltas (changes requested):
I don't think we want to suppress dev dependency warnings, since compromise of a dev/build machine can lead to compromise of production code. however i am ok with running the dev dependency audits less often if they're too noisy.

tsmartt:
> I don't think we want to suppress dev dependency warnings, since compromise of a dev/build machine can lead to compromise of production code. however i am ok with running the dev dependency audits less often if they're too noisy.

Yes, most come from react-scripts which [recommend only reporting prod](https://github.com/facebook/create-react-app/issues/11174) dependencies.

diracdeltas:
i'd rather just exclude individual audit errors that we know are low-risk (like regex DoS's) or force resolutions to fix the audit errors

diracdeltas:
https://github.com/brave-intl/publishers/pull/3516

tsmartt:
That works.

PR #3527, opened 2022-01-20
Fix Totp tests

tsmartt:
Fix the failing TOTP Tests.


Gemfile
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-def x86_64?
-  `uname -p`.strip == "x86_64"
+def not_m1?
+  `uname -p`.strip != "arm64"
```
thypon, line 4: On M1 `uname -p` is returning "arm" and not "arm64". I was using x86_64 since even Linux arm is broken.

tsmartt: On an x86 mac, `uname -p` returns `i386`. I'll just update it to check for that.

tsmartt: I got `arm64` on my `m1`, weird. I can update to include both `arm` and `arm64`

thypon (approved):

PR #3533, opened 2022-01-31
This PR adds webauthn support

tsmartt:
Chromium is phasing out U2F support in Feb 2022
https://developer.chrome.com/blog/deps-rems-95/

We need to move to the newer, better Webauthn standard.

Since the majority of our 2factor auths are stored in U2F format, we
will use a compatibility layer to verify. See:

https://www.stackallocated.com/blog/2019/u2f-to-webauthn/
https://github.com/cedarcode/webauthn-ruby/blob/master/docs/u2f_migration.md
https://github.com/hansnielsen/webauthn-rails-demo/blob/v1-sign-with-webauthn/app/controllers/two_factor_controller.rb

Note that this PR covers just signing in using the Webauthn to U2F
compatibility layer (by passing app_id in extensions). We still need to
register with Webauthn. That will require DB migrations and is the subject
of a second PR.

To test: 
Register a U2F key locally and then attempt to sign in. You should see the browser-based webauthn popup to verify your key. Successful authentication should redirect you to the homepage.


thypon:
@tsmartt are tests required to pass for this release?

thypon (approved):
Seems fine

app/services/two_factor_auth/webauthn_verify_service.rb
```diff
@@ -0,0 +1,53 @@
...
+        Rails.logger.debug("WebAuthn::Error! #{e}")
+        return OpenStruct.new(success?: false, result: nil, errors: [e])
+      ensure
+        session.delete(:current_authentication)
+      end
+
+      counter = assertion_response.authenticator_data.sign_count
+      if registration.counter >= counter
+        Rails.logger.debug("Credential replay detected!")
+        return OpenStruct.new(success?: false, result: nil, errors: [StandardError.new("Cannot authenticate.")])
```
OAuthBringer, line 45: @tsmartt Are the errors being stored in service classes being saved and reused for flash messages?

PR #3536, opened 2022-02-03
poc: Single call w/ interface, error handling, and openstruct return

OAuthBringer:
## Pull Request Name

POC of a single call method for service classes

#### Features

This is a POC of a builder implementation that uses a single class method constructor.  

#### How To Test

NA. This is just a

#### Pull Request Checklist


OAuthBringer:
@tsmartt After our convo yesterday I wanted to just throw up a quick implemention of the builder pattern syntax I've been using for the passed few years.  It is compatible with the existing `build.call` implementation but it is just a shortcut really.

OAuthBringer:
Hurray! I've finally gotten a build to pass.

test/services/builder_base_service_test.rb
```diff
@@ -0,0 +1,78 @@
...
+
+class BuilderBaseServiceTest < ActiveSupport::TestCase
+  test "BuilderBaseService.call" do
+    assert_raises NotImplementedError do
+      BuilderBaseService.call
+    end
+  end
+
+  test "UseBuildBaseService" do
+    class UseBuildBaseService < BuilderBaseService
```
OAuthBringer, line 14: @tsmartt I added a basic example of using `build` in the single method context as we discussed.

app/services/builder_base_service.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    # Error handling by default
+    begin
+      result = inst.call(*args, **kwargs)
+    rescue => e
+      errors.append(e)
+    end
+
+    OpenStruct.new(success?: errors.empty?, errors: errors, result: result)
+  end
```
OAuthBringer, line 24: @tsmartt My core problem right now with the the instance based `.call` method (i.e. `Service.build.call` ) is that we lose some of the coherence in the `Class.call` api.

When using the `call` method alone we gaurantee that all children must implement an instance call, can optionally override the class `build`, and all calls to `Class.call` will never raise an exception and will always return the same response.  All I have to do as a developer is deal with my childclass and return a result of any type and it will be reported. 

If we use `Service.build.call` we loose this because we are now faced with a few possibilities:
1. We override `Service.new.call` which means we loose the potency of the interface and the best we can do is abstract out the error handling and OpenStruct methods into some helper methods, but it then is left to the developer to know about/use those methods.
2. We define a BaseClass `BaseService.new.call` method that implements error handling and return values and then we use some new method in the child class to implement the primary API. `run` or some such.

I truly don't care about whether or not we call `build` in production, but I do care that the primary `call` method defines a coherent interface that has to be explicitly overridden in any given subclass rather than explicitly defined in all subclasses.

app/services/builder_base_service.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def call(*args, **kwargs)
```
OAuthBringer, line 26: @tsmartt This is an example of `#2` on my comment above.  Primary child class api is `run` method (or whatever, just for example).

Primary usage is `Service.build.call(*args, **kwargs)` and developer implements required `run` interface and optionally overrides `Class.build`

app/services/builder_base_service.rb
```diff
@@ -0,0 +1,35 @@
+# typed: true
+
+class BuilderBaseService
+  extend T::Helpers
+  extend T::Sig
+
+  class ::BSuccess < T::Struct
+    prop :result, T.untyped
```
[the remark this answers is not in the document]
OAuthBringer: i.e. define it in the global namespace as a nested class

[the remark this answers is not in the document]
tsmartt: pretty neat, I did it just so we wouldn't have to reference the whole thing every time

tsmartt (approved):

PR #3541, opened 2022-02-09
This adds the ability to add a security key using webauthn

tsmartt:
The old U2F protocol is being phased out. The new Webauthn protocol is
more flexible and nicer to work with. Since most of the registrations are
in U2F, we just keep them there but mark the format. All new registrations
will be done using the Webauthn format. Note we must run the u2f rake job
in prod after deploy to mark them and standardize the base64 encoding of
the already-stored keys.

Tested locally with both a Yubikey that was registered using U2F and a
separate one under Webauthn. With both registered I could log in using
both.

Currently in draft as service refactoring and tests are needed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,7 +48,7 @@ RUN yarn install --frozen-lockfile
 # The second dot will copy it to the WORKDIR!
 COPY . .
 
-RUN RAILS_ENV=production SECRET_KEY_BASE="1" bundle exec rails assets:precompile DB_ADAPTER=nulldb DATABASE_URL='nulldb://nohost'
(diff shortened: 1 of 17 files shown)


OAuthBringer (approved):
LGTM

OAuthBringer:
Implmentation looks good but builds is failing

thypon:
Even after fixing the missing `CREATORS_HOST` variable in WebAuthN initializer many tests are still failing :S

thypon:
```
Fabulous run in 608.908418s, 1.2038 runs/s, 3.7198 assertions/s.
733 runs, 2265 assertions, 7 failures, 4 errors, 0 skips```

PR #3542, opened 2022-02-09
Builder Base Class changes

tsmartt:
We introduce a union type which can be distinguished by pattern matching

The reason for this is we want Sorbet to help us as much as possible. If
we have an error, we don't want to accidentally put those in a success
callback. With the fields of the success/error structs being different,
sorbet can now help prevent sensitive information escaping.

An example of usage:

```ruby
  case res = ChildService.build.call
  when BFailure
    puts "I can access errors! #{res.errors}"
    problem("1")
  when BSuccess
    puts "I can access result!"
    pass(res.result.add(2))
  else
    T.absurd(res)
  end
```

### DIFF
--- a/app/services/builder_base_service.rb
+++ b/app/services/builder_base_service.rb
@@ -4,28 +4,32 @@ class BuilderBaseService
   extend T::Helpers
   extend T::Sig
 
-  class SuccessStruct < T::Struct
-    prop :success, TrueClass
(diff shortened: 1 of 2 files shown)


OAuthBringer (approved):
I also toyed around with T.must, I was hoping it could be used in the struct defintions themselves but no luck.

PR #3574, opened 2022-03-09
Improvement/eyeshade balances tasks

OAuthBringer:
1. Adds direct docker network access to eyeshade so access to postgres db is available
2. Adds new tasks to populate balances for all SiteChannels, all other balance types can be added as necessary.
3. Various Readme updates
4. Add python to docker dev installation, this is required to compile the javascript elements on the M1 (for some reason)


OAuthBringer:
Investigating the BA values issue

tsmartt (approved):

OAuthBringer:
I ran into a caveat of this approach, the eyeshade network is required for docker-compose to initialize. I.e., you have to have installed eyeshade in order for publishers to run.

In some ways this is not terrible behavior, but I'm seeing if I can bypass this.

PR #3576, opened 2022-03-14
Groundwork fields for better oauth error messaging

tsmartt:
We want to unify the way we deal with wallet connections and associated
failures to refresh tokens.

We want to run monthly jobs to refresh all the wallets that were paid out
last month. New wallets since then we'll consider OK since they were
created within 30 days.

We're going to need 2 jobs:
1) Monthly job that runs and refreshes all wallets that were paid out
last month. If a failure occurs, notes this in the DB.

2) Monthly email job. Finds all noted failures and emails the creator.
Notes in the DB that the creator was emailed. Only one email will be sent.
The PotentialPayout table will also take emailed failures into account.


tsmartt:
Opening another PR that is not across forks

PR #3579, opened 2022-03-22
task: Do not fail builds for dev dependencies

OAuthBringer:
We are getting smacked in the face WRT to momentum because we are blocking builds on audits of dev-dependencies that never leave the local stack.  I've updated the lint tasks to run against the actual production dependencies and removed the allowed CVE list.

We keep getting hit with vulnerabilities in `webpack-dev-server` which never leaves the local context.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -70,8 +70,10 @@ jobs:
         run: bundle exec standardrb
       - name: Brakeman
         run: bundle exec brakeman
-      - name: Yarn audit
-        run: npx audit-ci --config .audit-ci.json && (cd public/creators-landing && npx audit-ci --config ../../.audit-ci.json)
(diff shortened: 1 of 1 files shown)


tsmartt:
@diracdeltas or @thypon -- thoughts?

thypon:
I'm not 100% in if we want to disable the check altogether.
However, we can push the dev-dependencies check to the end so that we can see that everything else is fine.

In the past, these dependencies were taken over to perform malicious activities on developers' workstations, and these checks should also inform us if something like that is going to happen.

OAuthBringer:
@thypon What about reducing the severity to a warning?  The concern of developer work stations is valid, but blocking all work because of it seems overly aggressive.  I'm getting blocked and forced to context switch to fix a vuln on a dev dependency at least once a week and it's drastically hurting momentum.

PR #3581, opened 2022-03-22
Run security audits on production/master PR. Run tests/lint on push

OAuthBringer:
I closed the previous PR as I was testing triggers, but I think the changes here should satisfy the concerns.  We can merge/deploy into staging so we can iterate quickly, but to deploy to prod we need to have passing builds on PR for security audits.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -70,8 +70,10 @@ jobs:
         run: bundle exec standardrb
       - name: Brakeman
         run: bundle exec brakeman
-      - name: Yarn audit
-        run: npx audit-ci --config .audit-ci.json && (cd public/creators-landing && npx audit-ci --config ../../.audit-ci.json)
+      - name: Yarn audit js assets
+        run: yarn audit --groups dependencies
+      - name: Yarn audit landing-page
+        run: cd public/creators-landing && yarn audit --groups dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -7,12 +7,7 @@
 
 name: Lint
 
-on:
-  push:
-    branches: [ master, staging, dev ]
-  pull_request:
```
thypon, line 13: Is there a reason for removing the `lint` pass from PRs?

OAuthBringer: @thypon I've made it so that it runs on every push to the remote. I had another PR open that was making these changes anyway so I just merged them together.  From a workflow perspective it is actually much more useful to have the basic linting tasks run a lot.

thypon: It seems the checks are somehow shown in this PR as well, so it does not seem to affect the output.

thypon (approved):

.github/workflows/lint.yml
OAuthBringer: @thypon The linting checks were run when I pushed the branch to the remote repository, the other checks were triggerred by me opening the PR.

thypon (approved):

tsmartt (approved):

PR #3596, opened 2022-03-25
task: Audit runtime dependencies only

OAuthBringer:
Audit on staging push, check only runtime dependencies.

I updated the trigger to be any push/merge of code to staging to ensure it was always run.  I oddly enough I noticed that the audit task was run only on the initial PR to master, and not afterwards.  Github actions are sometimes difficult to reason about, but this definitely runs anytime code is integrated into stagnig.


OAuthBringer:
Closing this for the time being, may not be necessary.

PR #3598, opened 2022-03-28
Release 3/28/22: Addition of country dropdown data to Settings [ Feature Flagged ]

OAuthBringer:
* task: Update dev Dockerfile

The previous command seemed to be overly complicated and I suspect that
it was masking failures.  This seems to work just fine

* feat: Countries Dropdown and minimal location form

Testing: Should be fully backwards compatible/feature flagged.  Settings section should not be visible for staging/production users unless feature is enabled via rake task or console.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -29,7 +29,7 @@ RUN RAILS_ENV=development
 
 # Install the gems.
 RUN gem install nokogiri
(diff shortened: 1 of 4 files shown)


OAuthBringer:
Hmm, the test action is being run twice, once with pull and once with pull request.

OAuthBringer:
Beautiful

tsmartt (approved):

PR #3601, opened 2022-03-30
Task/unpin chrome version

OAuthBringer:
This is the progress I've made on updating to the latest chromedriver version.

The core issue is that we need to initialize `chromedriver` as `chromedriver --allowed-ips=127.0.0.1`

Chrome driver helper is legacy/deprecated and appears to only be there in order to pin the version of chrome.


tsmartt:
Closing until such a time when we have time to fix this.

PR #3605, opened 2022-04-03
chore: update Webauthn config to align supported algorithms with policy

kdenhartog:
#### Features

This is meant to be updating the supported algorithms policy to check that the new algorithms can be used for verification. This closes https://github.com/brave/security/issues/802

#### How To Test

1. Use webauthn with an authenticator that uses one of the new algorithms

#### Pull Request Checklist

- [X] Adequate test coverage exists to prevent regressions -- [Guide to Testing](https://guides.rubyonrails.org/testing.html)
- [X] XSS is mitigated -- [Guide to XSS Prevention](https://guides.rubyonrails.org/security.html#cross-site-scripting-xss)
- [X] No raw SQL -- [Guide to SQL Injection Prevention](https://guides.rubyonrails.org/security.html#sql-injection)
- [X] UI/UX is responsive -- [Guide to Responsive UI/UX](https://developers.google.com/web/fundamentals/design-and-ux/responsive/)
- [X] Passes checklist for Progressive Web App -- [Guide to PWAs](https://developers.google.com/web/progressive-web-apps/checklist)

### DIFF
--- a/config/initializers/webauthn.rb
+++ b/config/initializers/webauthn.rb
@@ -31,5 +31,6 @@ WebAuthn.configure do |config|
   # Possible values: "ES256", "ES384", "ES512", "PS256", "PS384", "PS512", "RS256", "RS384", "RS512", "RS1"
   # Default: ["ES256", "PS256", "RS256"]
   #
-  # config.algorithms << "ES384"
+  # sets the algorithms we support for verification with webauthn
+  config.algorithms = ["ES256", "ES384", "ES512", "PS256","PS384", "PS512"]
 end
--- a/test/webauthn_config_test.rb
+++ b/test/webauthn_config_test.rb
@@ -1,0 +1,15 @@
+# typed: false
+require "test_helper"
+
+class WebauthnConfigurationTest < ActiveSupport::TestCase
+
+  test "Webauthn supported verification algorithms are correctly configured" do
+    assert WebAuthn.configuration.algorithms.include?("ES256")
+    assert WebAuthn.configuration.algorithms.include?("ES384")
+    assert WebAuthn.configuration.algorithms.include?("ES512")
+    assert WebAuthn.configuration.algorithms.include?("PS256")
+    assert WebAuthn.configuration.algorithms.include?("PS384")
+    assert WebAuthn.configuration.algorithms.include?("PS512")
+    assert !WebAuthn.configuration.algorithms.include?("RS256")
+  end
+end


kdenhartog:
@tsmartt or @thypon do either of you know a good way to generate an equivalent unit test for one of the new algorithms so that we can properly regression test this?

tsmartt (approved):

kdenhartog:
Was thinking about another one of these [test response](https://github.com/brave-intl/publishers/blob/2b1da8a703bab908ab4f307c31d33b272d04597b/test/controllers/u2f_authentications_controller_test.rb#L7) @thypon since it looks like we've been able to generate them before, but I couldn't figure out how.

thypon:
I think this is real data coming from a real token, saved there; the check is then stubbed to always return `true` on verify, so it's just a placeholder.

Not sure if we can easily test them or if we want to test them programmatically though. Last word on @tsmartt , since I feel this is an improvement over the status quo either way, with the test or without them.

tsmartt:
We can add a test to verify the config is correct, that should be easier. And we can trust the library's implementation.

tsmartt:
I can do early next week if you cannot do it sooner.

OAuthBringer:
@kdenhartog You should be able to do a trivial assertion against the Webauthn configuration in a spec.

In irb you can access the allowed algorithms via `WebAuthn.configuration.algorithms`

So you can explicitly check to assert that the config contains/does not contain whatever you are looking to test

So something like `test/webauthn_config_test.rb`

`assert WebAuthn.configuration.algorithms.include?("PS512")`
`assert !WebAuthn.configuration.algorithms.include?("RS256")

FWIW you can also complete override the allowed algorithms with

`config.algorithms = [only, the, ones, I, want]`

kdenhartog:
Sounds like an easy way to handle this to me! I can add that today

OAuthBringer:
@thypon this appears to be failing due to a login failure for semgrep?

OAuthBringer:
@kdenhartog I'm not sure why the semgrep job keeps failing.  I'm wondering if it's because the branch is forked?

tsmartt:
Yes, if you reopen this as a branch on the main repo, it has access to secrets and should pass.

tsmartt (approved):

kdenhartog:
Ahh ok. I think at the time when I originally opened I didn't have the proper access for that and so just forked to handle this. Looks like we managed to get it through now and going forward I'll push branches directly to origin remote rather than a forked remote.

PR #3609, opened 2022-04-06
Feat/refresher service

OAuthBringer:
1. Oauth2::ClientCredentials -Generic client for OAuth2 spec client credentials requests and responses
2. Oauth2::Structs - To spec data containers for Oauth2 requests/errors
3. Oauth2ProviderProperties - Sorbet abstract module to create interface for any Oauth2 client credentials provider (i.e. wallets)
4. Oauth2RefresherService - Implementation of the other utilities.

This is built to be generic and to be able to handle refresh token flows for any to spec OAuth2 client credentials grant refresh token flow.

The OAuth2::ClientCredentials client can be extended to support all client credentials grant spec associated flows including the generation of an access token.


lib/oauth2/client_credentials.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  sig { params(client_id: String, client_secret: String, token_url: String).void }
+  def initialize(client_id:, client_secret:, token_url:)
+    @client_id = client_id
+    @client_secret = client_secret
+    @token_url = token_url
+    @options = {use_ssl: true}
+  end
+
+  #      def access_token(scope = nil)
```
tsmartt, line 21: Please remove comments, but otherwise LG2M

PR #3614, opened 2022-04-08
Feat/refresh job

OAuthBringer:
Oauth2BatchRefreshJob - Fills queue of RefreshJobs on schedule
Oauth2RefreshJob - Handles individual connections, handles retries/failures/parallelism

### DIFF
--- a/app/jobs/oauth2_batch_refresh_job.rb
+++ b/app/jobs/oauth2_batch_refresh_job.rb
@@ -1,0 +1,24 @@
+class Oauth2BatchRefreshJob < ApplicationJob
+  queue_as :scheduler
+
+  def perform(wait: 0, limit: nil)
+    count = 0
+    [UpholdConnection].each do |klass|
+      base_query = limit.present? ? klass.limit(limit) : klass
+
+      # The idea here is I'm filling a queue to refresh tokens from each connection type
+      # while adding a delay so I do not overwhelm the API rate limit of a given provider
+      base_query.select(:id).find_in_batches do |batch|
+        batch.each do |connection|
+          count += 1
+          # I'm using this job to call individual jobs so I can take advantage of retries/failover due to thinks like rate limits
+          # I've used exactly this pattern before to handle mass email jobs, though I don't always wait between enqueing tasks.
+          Oauth2RefreshJob.perform_later(connection.id, klass.name)
+          sleep(wait)
+        end
+      end
+    end
+
+    count
+  end
+end
--- a/app/jobs/oauth2_refresh_job.rb
+++ b/app/jobs/oauth2_refresh_job.rb
@@ -1,0 +1,11 @@
+class Oauth2RefreshJob < ApplicationJob
+  extend T::Sig
+  queue_as :default
+
+  # Serialization of the model is awkward here. Will need to look into it
+  sig { params(connection_id: String, model: Class).returns(Oauth2RefresherService::TYPES) }
+  def perform(connection_id, model)
+    connection = model.find_by_id!(connection_id)
+    Oauth2RefresherService.build.call(connection)
+  end
+end
--- a/app/jobs/oauth2_refresh_job.rb
+++ b/app/jobs/oauth2_refresh_job.rb
@@ -1,11 +1,20 @@
+# typed: true
+
 class Oauth2RefreshJob < ApplicationJob
-  extend T::Sig
   queue_as :default
 
-  # Serialization of the model is awkward here. Will need to look into it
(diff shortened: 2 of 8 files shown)


app/jobs/oauth2_refresh_job.rb
```diff
@@ -0,0 +1,20 @@
+# typed: true
+
+class Oauth2RefreshJob < ApplicationJob
+  queue_as :default
+
+  def perform(connection_id, klass_name)
+    case klass_name
+    when "UpholdConnection"
```
tsmartt, line 8: This is a thought experiment and won't hold up the PR b/c it looks pretty good as is...

A thought experiment for (potential) simplification. 

If we raise this `case` up one level and call connection-specific refreshers, for ex `UpholdWalletRefreshJob`, `GeminiWalletRefreshJob`, etc.

Those classes could handle all of the details internally for updating a connection's refresh information. That would be the only contract. I call you, you refresh the token for your particular provider. You're free to do that however you want.

Those jobs could make the HTTP call in one line(`ClientCredentials`, then do the equivalent of what 
`Success.new(connection: connection.update_access_tokens!(result))` is doing now -- in one class. I _think_ it would be easier to read and see "oh, this is doing the 2-step process in here` -- just a thought.

But all in one class, there to see. Just an idea. We wouldn't need `oath2_refresh_job`, or `client_credentials` becomes that...

This gives us a little flexibility for the provider-specific wallet providers to customize as needed, but they are still able to compose the lower level pieces. But we wouldn't force those lower level pieces to conform to the http specifics of each provider.

app/jobs/oauth2_batch_refresh_job.rb
```diff
@@ -0,0 +1,24 @@
...
+  queue_as :scheduler
+
+  def perform(wait: 0, limit: nil)
+    count = 0
+    [UpholdConnection].each do |klass|
+      base_query = limit.present? ? klass.limit(limit) : klass
+
+      # The idea here is I'm filling a queue to refresh tokens from each connection type
+      # while adding a delay so I do not overwhelm the API rate limit of a given provider
+      base_query.select(:id).find_in_batches do |batch|
```
tsmartt, line 11: Nice optimization.

tsmartt (approved):

app/services/oauth2_refresher_service.rb
```diff
@@ -13,8 +15,13 @@ def self.build
...
   end
 
-  sig { override.params(connection: UpholdConnection).returns(T.any(Oauth2RefresherService::Success, BFailure)) }
+  sig { override.params(connection: UpholdConnection).returns(Oauth2RefresherService::TYPES) }
   def call(connection)
+    if connection.refresh_token.nil?
+      connection.record_refresh_failure!
+      return BFailure.new(errors: ["Cannot refresh without refresh token"])
+    end
+
```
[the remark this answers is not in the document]
OAuthBringer: @tsmartt That's not what it's doing.  It is intended to handle all oauth2 client credentials spec methods eventually, but  right now it is only handling refresh tokens.  The intended direction of `ClientCredentials` is to have an `access_token` and `code` method alongside the `refresh_token` because that is the logical grouping of possible grant types that is associated with the client credentials. See the spec for the client credentials grant: https://datatracker.ietf.org/doc/html/rfc6749#section-4.4

If we encounter a wallet authentication strategy that is not using ClientCredentials I will write another class that handles that specific flow and update the various jobs accordingly and probably subclass the basic HTTP request mechanisms to be shared between the different OAuth2 flows.  Broadly, I am solving for all cases of client credentials grant flows so we don't have to keep reimplementing the same logic in different classes.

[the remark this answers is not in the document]
OAuthBringer: Oauth2::ClientCredentials is an HTTPClient, not a service.  I write them very differently because their concerns are quite different.

app/jobs/oauth2_refresh_job.rb
OAuthBringer: I don't want to have 3 separate classes that implement the same refresh token flow when I can write an HTTP client, a service, and an interface that implements an API that can be used for all three cases/potentially all cases.

Let me show you the power of OOP.  COME TO THE DARK SIDE TYLER.

OAuthBringer: I considered converting OAuth2RefreshJob into <wallet type>RefreshJob, but they each end up being pretty much identical with only the names different.  I might end up doing that in the end because passing the klass_name is a bit awkward, but I didn't want to linger on it too much right now.  I'll likely end up modifying it as I continue with the the refresh token elements.  Everything I do is iterative and so I will refine different pieces as I go.

PR #3627, opened 2022-04-15
Task/debugging

OAuthBringer:
Adds a development only Oauth2 controller that implements each auth code flow for each of the child Oauth2::AuthenticationBase implementations.  This was necessary for debugging locally.

Updates the Oauth2::AuthorizationCodeClient to handle `application/json` based auth flows though they are not to spec.

Relaxes the error handling a bit.  Bitflyer threw a 400 that didn't follow oauth2 error specifications and was masking the input.  I modified some things to make it easier to debug the individual requests.

### DIFF
--- a/test/controllers/oauth2_controller_test.rb
+++ b/test/controllers/oauth2_controller_test.rb
@@ -1,0 +1,7 @@
+require "test_helper"
+
+describe Oauth2Controller do
+  # it "does a thing" do
+  #   value(1+1).must_equal 2
+  # end
+end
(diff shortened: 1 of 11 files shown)


test/controllers/oauth2_controller_test.rb
```diff
@@ -0,0 +1,7 @@
+require "test_helper"
+
+describe Oauth2Controller do
+  # it "does a thing" do
```
tsmartt, line 4: Empty test.

tsmartt (approved):

test/controllers/oauth2_controller_test.rb
OAuthBringer: generator artifact.  I can update it in a follow up.

PR #3637, opened 2022-04-22
Feat/wallet refresh failure

OAuthBringer:
1. Updates RBI files for new methods
2. Adds minimal mailer for wallet refresh notification
3. Updates batch refresh task for all connections
4. Adds conditional notification flag for connection refresh job

### DIFF
--- a/app/mailers/publisher_mailer.rb
+++ b/app/mailers/publisher_mailer.rb
@@ -32,15 +32,16 @@ class PublisherMailer < ApplicationMailer
     )
   end
 
-  def wallet_refresh_failure(publisher)
+  def wallet_refresh_failure(publisher, wallet_provider)
     @publisher = publisher
     @publisher_log_in_url = log_in_publishers_url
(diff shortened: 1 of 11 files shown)


app/mailers/publisher_mailer.rb
```diff
@@ -32,6 +32,19 @@ def verification_done(channel)
     )
   end
 
+  def wallet_refresh_failure(publisher)
+    @publisher = publisher
+    @publisher_log_in_url = log_in_publishers_url
+    @wallet_provider = "Uphold"
```
OAuthBringer, line 38: artifact

tsmartt (approved):
LG2M minus lint failure.

OAuthBringer:
Yeah, static RBI stuff.  I'm starting to get a better sense of how those static RBI's are used and it's sort of annoying on an M1.  I have to switch over to my other device to update the rbis.  In this case i'm just going to untype the job as sorbet doesn't seem to play that well with async tasks.

PR #3640, opened 2022-04-26
fix: Raise instance instead of class

OAuthBringer:
Several new relic errors were being raised because UnknownError takes params and rather than raising the result/instance I was raising the class.

### DIFF
--- a/test/models/gemini_connection_test.rb
+++ b/test/models/gemini_connection_test.rb
@@ -27,6 +27,16 @@ class GeminiConnectionTest < SidekiqTestCase
           assert_instance_of(klass, conn.refresh_authorization!)
         end
       end
+
+      describe "when unsuccessful" do
+        before do
+          mock_unknown_failure(klass.oauth2_client.token_url)
+        end
+
+        test "it should raise an error" do
+          assert_raises(Oauth2::Errors::UnknownError) { conn.refresh_authorization! }
+        end
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


test/models/gemini_connection_test.rb
```diff
@@ -27,6 +27,16 @@ class GeminiConnectionTest < SidekiqTestCase
...
         end
       end
+
+      describe "when unsuccessful" do
+        before do
+          mock_unknown_failure(klass.oauth2_client.token_url)
+        end
+
+        test "it should raise an error" do
+          assert_raises(Oauth2::Errors::UnknownError) { conn.refresh_authorization! }
```
tsmartt, line 37: Can we test that the value is raised. It looks like we're testing for the thing we deleted.

OAuthBringer: @tsmartt No, the failing error was "ArugmentError"

OAuthBringer: `Loading development environment (Rails 6.1.4.7)
irb(main):001:0> raise Oauth2::Errors::UnknownError
/var/www/lib/oauth2/errors.rb:6:in `initialize': missing keywords: :response, :request (ArgumentError)
irb(main):002:0>`

OAuthBringer:
Hopefully fixing non-deterministic test

tsmartt (approved):

PR #3644, opened 2022-04-27
task: Delete "Direct User Tips" column in Statements

### LINKED ISSUE #3607: Delete "Direct User Tips" column in Statements
image

Remove this column as we have no vision into direct tips (i.e., verified Uphold user sends a tip to verified Uphold creator).

jlbyrne:
#### Before:
Screen Shot 2022-04-27 at 5 18 15 PM

Screen Shot 2022-04-27 at 5 18 05 PM

#### After:
Screen Shot 2022-04-27 at 5 05 54 PM
Screen Shot 2022-04-27 at 5 05 44 PM

### Features

Closes #3607 .  Because the app doesn't have access to direct tipping information, the values in that column will always be 0.  Removed both the 'Direct User Tips' column on the statements page, and the 'Direct User Tips' row on the statement detail view.

### DIFF
--- a/app/javascript/locale/en.ts
+++ b/app/javascript/locale/en.ts
@@ -226,7 +226,6 @@ export default {
         total: "Total",
         type: "Type"
       },
-      directUserTips: "Direct User Tips",
       download: "Download",
       earningPeriod: "Earning Period",
       fees: "Fees",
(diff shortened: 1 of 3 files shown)


OAuthBringer:
@jlbyrne Rebase off of staging and that test failure will be resolved.

@tsmartt I vote that we remove the bundler audit test as it is redundant with the 3 other scanning utils we have including the audit action which is picking up the same issue.

tsmartt (approved):
LG2M, thanks for the first PR!

jlbyrne:
🥳

PR #3653, opened 2022-04-28
task: Persist node_modules on container

### LINKED ISSUE #3651: Docker-compose overrides `node_modules` directory due to volume mounting.
**Describe the bug**
In development context yarn install is run in the build step where it successfully installs node_modules.  However, once the container is initialized in the docker-compose context, the volume share syncs and removes the node_modules directory.

**To Reproduce**
1. If you have a functional docker context state, recursively remove `node_modules` in your project dir
2. Rebuild docker compose context from scratch `docker-compose down; docker-compose up --build --force-recreate`
3. Attach to the web container `make docker-shell` and observe that `node_modules` is not present in the `/var/www` directory despite the `yarn install` step of Dockerfile.dev completing successfully.
4. Run `yarn install` locally or in the Docker context and this will resolve the issue.

**Expected behavior**
Executing `make` from a clean installation state should handle all application dependencies including node modules.

Just documenting this for posterity.  I'll try to come up with a work around later.

cc @tsmartt

@jlbyrne Reported the underlying behavior and I tracked it down.  This has caused me confusion in the past as well.

OAuthBringer:
Ensures local directory does not override content of node_modules installed by docker.  I rebuilt from scratch several times and confirmed though I did not test what the behavior was when you had a pre-existing node-modules directory.

See: https://stackoverflow.com/questions/29181032/add-a-volume-to-docker-but-exclude-a-sub-folder

Closes #3651

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,6 @@ Development with Docker and `docker-compose` is recommended for anyone just gett
 1. Ensure `openssl` is installed. `brew install openssl`
 1. In your browser, navigate to `brave://flags`.  Make sure `Allow invalid certificates for resources loaded from localhost.
 ` is enabled. 
-1. Ensure `yarn` is installed. `npm install --global yarn`
-1. Run `yarn`
 1. Run `make`
 1. Create an admin user. `make admin EMAIL="email@example.com"`
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,6 +22,7 @@ services:
       - redis
     volumes:
       - .:/var/www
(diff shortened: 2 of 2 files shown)


jlbyrne (approved):
It worked!  Nice fix.

OAuthBringer:
@jlbyrne I was mulling over all sorts of ways to handle creating/copying over the node_modules directory during the docker build and then I was like... wait... certainly someone else has run into this before.  Aaaaannndd they had.

PR #3658, opened 2022-05-10
Update ejs dep on landing page

tsmartt:


thypon:
Do we ever have any non-generated source code under `public/static/js/.*` ?

tsmartt:
> Do we ever have any non-generated source code under `public/static/js/.*` ?

It's all generated.

tsmartt:
@OAuthBringer updated the dep, should be good to go. Also updated Rails since it complained.

OAuthBringer (approved):

PR #3663, opened 2022-05-10
Fix/deposit

OAuthBringer:
1. Abstract notification logic from job to service
2. Re-enable deposit id job, update to user refresher/notification
3. Remove artifact refresher/syncing implementations to cut down on unused code


OAuthBringer:
@tsmartt There are some issues with this PR that I am addressing in a follow up as I don't want to get too bogged down in scope creep/debt paydown.  This should be enough however to at least test/verify the basic fix in staging and if it works I can unblock the user and refine things afterwards.

tsmartt (approved):

PR #3665, opened 2022-05-11
improvement: Add scopes for debugging and sanity

OAuthBringer:
Hard to debug in staging without being able to quickly see things.

### DIFF
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -36,6 +36,7 @@ class Channel < ApplicationRecord
 
   has_one :site_banner, dependent: :destroy
   has_one :site_banner_lookup, dependent: :destroy
+  has_one :bitflyer_connection, through: :publisher
 
   has_many :potential_payments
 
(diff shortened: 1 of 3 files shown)


app/models/channel.rb
```diff
@@ -112,6 +113,17 @@ class Channel < ApplicationRecord
     where.not(contested_by_channel_id: nil).where("contest_timesout_at < ?", Time.now)
   }
 
+  scope :missing_deposit_id, -> { where(deposit_id: nil) }
+  scope :using_active_bitflyer_connection, -> {
+    joins(:publisher)
```
OAuthBringer, line 118: this might actually be redundant but the specs work.  I could remove it later

tsmartt: This would be useful to have for Uphold and Gemini too, eventually.

OAuthBringer: @tsmartt For sure, pattern is the same.

tsmartt: maybe just `with_active_bitflyer_connections`

OAuthBringer: Yeah, some standard would be better.  Not overthinking it atm.  If/when we add the others we can clean it up.

jlbyrne (approved):
LGTM!

PR #3666, opened 2022-05-11
fix: Remove this, redundant, shouldn't be here, causing a fail

OAuthBringer:
Pretty sure this is breaking staging and it's not necessary.  The same thing is done in... a lot of places.


OAuthBringer:
I setup the PR to make sure all jobs run.  I was maybe premature in requesting reviews, sorry about that.

tsmartt (approved):

PR #3669, opened 2022-05-11
task: Don't run jobs if connection is broken

OAuthBringer:


app/models/channel.rb
```diff
@@ -376,15 +376,16 @@ def notify_slack
...
-        publisher.selected_wallet_provider.sync_connection!
-      end
+    # No need to run if it won't work
+    return if !wallet || wallet.oauth_refresh_failed
+
+    case wallet
+    when BitflyerConnection
+      Sync::Bitflyer::UpdateMissingDepositJob.perform_async(id) if deposit_id.nil?
+    when GeminiConnection
+      wallet.sync_connection! if gemini_connection_for_channel.blank?
```
OAuthBringer, line 388: I'm honestly not really sure why we would do this, but I didn't want to get hung up on it.

tsmartt (approved):

PR #3671, opened 2022-05-13
task: Disable Retry

OAuthBringer:
In an abundance of caution I'm disabling retries until I have a better sense of actual production behavior here.


OAuthBringer:
@jlbyrne For context, the underlying reason I'm performing all these updates/fix is because the update deposit job was basically failing all the time and throwing 401s, in response, Bitflyer blacklisted our IP addresses so that we couldn't make any requests at all.  @tsmartt disabled the deposit update job entirely to hotfix the issue and now hopefully I have resolved it.

tsmartt (approved):

PR #3673, opened 2022-05-13
improvement: Add exhaustiveness checks

OAuthBringer:
This adds the sorbet exhaustiveness check.  The core value here is that these are the only possible responses that can be returned from the service.  It is self documenting and statically typed.  We should basically end any sorbet typed function this way (which was something I didn't appreciate before).

T.absurd() implies the end of code execution so this always has to be at the end of the function (which I didn't know).

### DIFF
--- a/app/services/bitflyer/update_deposit_id_service.rb
+++ b/app/services/bitflyer/update_deposit_id_service.rb
@@ -1,43 +1,49 @@
 # typed: true
 
 class Bitflyer::UpdateDepositIdService < BuilderBaseService
+  extend T::Helpers
+  extend T::Sig
   include Oauth2::Responses
 
   def self.build
     new
   end
 
+  sig { override.params(channel: Channel).returns(BServiceResult) }
   def call(channel)
     return shrug("NOOP: Deposit id was present on the channel.") if channel.deposit_id.present?
 
-    conn = channel.publisher.bitflyer_connection
+    conn = channel.bitflyer_connection
 
     return shrug("NOOP: No bitflyer connection detected.") if conn.nil?
 
     result = conn.refresh_authorization!
 
+    # Man exhaustiveness checking is awesome.
     case result
+    when BitflyerConnection
+      # FIXME: Ideally this would be part of it's own client
+      url = URI.parse(Rails.application.secrets[:bitflyer_host] + "/api/link/v1/account/create-deposit-id?request_id=" + SecureRandom.uuid)
+      request = Net::HTTP::Get.new(url.to_s)
+
+      request["Authorization"] = "Bearer " + T.must(conn.access_token)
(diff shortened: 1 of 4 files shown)


app/services/bitflyer/update_deposit_id_service.rb
```diff
@@ -1,43 +1,49 @@
...
+      # Passing the channel along mostly for debugging purposes
+      # If I really care about/want the channel I will create an explicit
+      # result type and pass it with a static type of Channel
+      pass([channel])
     when BFailure
-      return result
+      result
     when ErrorResponse
-      return BFailure.new(errors: [result])
+      BFailure.new(errors: [result])
```
tsmartt, line 44: What's the difference between an ErrorResponse and a BFailure? Ideally we should just have to check `BFailure` as the service should wrap up any errors and return them in a `BFailure`. That will simplify our checking to Success or Failure. Thoughts?

OAuthBringer: ErrorResponse is not returned from a service, it is a a to-spec data container that represents an Oauth2 error response schema object, is returned from Oauth2::AuthenticationCodeClient, and is one of the statically defined response types defined in `AuthenticationCodeBase.refresh_authorization!`.  

As you can see, that's actually exactly what I'm doing, is wrapping the ErrorResponse in BFailure.  But ErrorResponse is highly specific.  It tells me exactly what has happened, which is that an Oauth2 request has failed but has returned a valid oauth2 spec response object which always has an `error`, `error_description` and another optional arg that I forget.

OAuthBringer: Also, I only partially agree with the idea behind BFailure, especially when you are using exhaustiveness checking.  There is no ambiguity about what responses are being returned, and depending on what you plan on doing in a failure state, an explicit failure object is much more informative than a generic one.  I only use BFailure or BSuccess when I'm only concerned with Success or Failure States.  If I need to return data from one service to another, I always use a unique object type because it tells me exactly what has happened and has a rigorously defined schema.

tsmartt: Due to the limits of Sorbet not being able to dynamically wrap whatever return value we want in `BFailure`, this seems a fine way to go.

tsmartt (approved):

app/services/bitflyer/update_deposit_id_service.rb
OAuthBringer: Yeah, if we could easily define subschemas I _might_ feel differently, though the experience of knowing exactly what has happened simply by virtue of a response being an explicit object has been eye opening.  It makes reasoning about the system dramatically easier.

PR #3713, opened 2022-06-02
refactor: Generic Bitflyer Oauth2 Auth Code Flow

OAuthBringer:
See: brave-intl/creators-private-issues#88
Subset of: brave-intl/creators-private-issues#62

Updates:
1. Implements the generic Oauth2Controller for a functional auth code pathway (Bitflyer)
2. Removes the insecure/legacy state verification process for bitflyer and uses an expiring encrypted http only state cookie.  The state token field on the bitflyer model is no longer necessary.
3. Ensures that a bitflyer connection is only created after a successful access token request, and that all the required updates for the publisher, connection, and associated channels are done synchronously wrapped in a single transaction.
4. Updates refresh_authorization! logic to handle artifact missing values
5. Expands Oauth2 auth code client to support code exchange verification: https://datatracker.ietf.org/doc/html/rfc7636#section-4.1


tsmartt (approved):

OAuthBringer:
@tsmartt I'm holding off on this until the other changes go out.

PR #3716, opened 2022-06-03
improvement: Notify commit and environment after initialiation

OAuthBringer:
This actually notifies in the sidekiq container instead of web... but it is more than nothing.  Also having a clear notice of when a build has actually succeeded would be really helpful in general.


jlbyrne (approved):
A slack notification is great though!!

PR #3719, opened 2022-06-06
Task/minor followups

OAuthBringer:
I'm just doing some follow up tasks related to all of the oauth2 connection work I've been doing as well as the bitflyer updates.  For now I'm adding a global exception handler and logging for any oauth2 controller connections that fail (Currently only bitflyer, but will eventually apply to all wallet connections).

For bitflyer specifically this sends up a generic flash message for any failure related to creating a bitflyer connection.  

Screen Shot 2022-06-06 at 5 24 48 PM

Part of brave-intl/creators-private-issues#62


OAuthBringer:
See commit message for commentary on removal of wallet syncing

PR #3734, opened 2022-06-09
updates: More bitflyer tweaks/debugging.

OAuthBringer:
For the sake of reasonability I'm advocating we remove most of these callbacks.  They are prone to failure, create weird side effects and race conditions as well as get called in contexts that are inappropriate.  Let's just run these jobs in tight well controlled high frequency cadences and leave models to manage database records as deterministically as we can.

Callbacks man.  One of rails sharpest knives.  Especially on models.

### DIFF
--- a/app/jobs/sync/bitflyer/update_missing_deposits_job.rb
+++ b/app/jobs/sync/bitflyer/update_missing_deposits_job.rb
@@ -5,10 +5,7 @@ class Sync::Bitflyer::UpdateMissingDepositsJob
 
   # THere is definitely a rate limit here despite being told there wasn't.  I manually
   # executed the service because the wrong id value was being passed to the job and it succeeded and eventually hit a 429.
-  #
-  # Update: There really really really is a rate limit despite being repeatedly told otherwise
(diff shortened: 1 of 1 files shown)


jlbyrne (commented):
We're still leaving the job in `publishers/app/services/bitflyer/auth_completer.rb`?

OAuthBringer:
@jlbyrne Where are you seeing auth_completer?  I deleted that code recently as part of the bitflyer wallet auth rewrite

jlbyrne:
ah, my bad.  I haven't pulled from staging recently.

jlbyrne (approved):

OAuthBringer:
@jlbyrne All good, I'm trying to purge all of this stuff so extra eyes are appreciated

tsmartt (approved):
I approve of removing callbacks and shunting everything into service objects that are easy to read and discover what the steps are. 
Or like in this case, an async job.

OAuthBringer:
@tsmartt Yeah, I mean I honestly can't come up with any other way tons of failed bitflyer requests have occurred other than through weird callback behavior so I'm just ready to knife it all.

PR #3739, opened 2022-06-13
Fix/defensive codes

OAuthBringer:
Newrelic threw an error specifically related to this.  Not actually sure how it is possible given that production creates an uphold connection for every user, but might as well be defensive anyway.


jlbyrne (approved):

OAuthBringer:
Without this and an uphold connection you get a 500 when attempting to visit statements when you do not have any sort of uphold connection.

PR #3743, opened 2022-06-15
feat: Uphold::FindOrCreateCardService

OAuthBringer:
This stand alone replaces the full set of functionality found in Jobs::CreateUpholdCardsJob.  Once implemented it will also patch the bug (with specs) that allows users who have changed the label of their uphold card to set a wallet address that is != to whatever the default currency is in creators.  This bug is probably what caused the payout issue for DDG

### DIFF
--- a/app/models/uphold_connection.rb
+++ b/app/models/uphold_connection.rb
@@ -309,6 +309,10 @@ class UpholdConnection < Oauth2::AuthorizationCodeBase
     self
   end
 
+  def uphold_client
+    @_uphold_client ||= Uphold::ConnectionClient.new(self)
+  end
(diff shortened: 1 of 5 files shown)


OAuthBringer:
For reference, the offending lines of code are: https://github.com/brave-intl/publishers/blob/staging/app/jobs/create_uphold_cards_job.rb#L41-L49

What happens is that the first card that is found gets set regardless of whether or not it matches the default currency of the connection.  Uphold has a USD card by default.  Ergo, first card == USD.  This case is explicitly tested in the service specs.

tsmartt (approved):

PR #3750, opened 2022-06-17
fix: Update to use current API implementation

OAuthBringer:


OAuthBringer:
Wonder how many tests this breaks.

jlbyrne (approved):

PR #3751, opened 2022-06-17
Improvement/authorize channel transfers

OAuthBringer:


app/models/publisher.rb
```diff
@@ -340,6 +341,22 @@ def locked?
     last_status_update&.status == PublisherStatusUpdate::LOCKED
   end
 
+  sig { returns(T::Boolean) }
+  def is_associated_with_suspended_uphold_ids?
+    uphold_id = uphold_connection&.uphold_id
+    return false if !uphold_id
+
+    UpholdConnection.where(uphold_id: uphold_id, publisher_id: Publisher.suspended.select(:id)).count > 0
```
tsmartt, line 349: We should be able to do this with a join and avoid 2 queries.

OAuthBringer: @tsmartt It's not 2 queries, it's a select, it gets executed in 1, but you are correct that we could do it in a join.

PR #3755, opened 2022-06-21
Refactor/uphold connection

OAuthBringer:


OAuthBringer:
@tsmartt @jlbyrne This is a big PR, but this was the smallest relevant chunk I could muster here.  The effort of continuing to maintain backwards compatibility became too great so I just moved it forward.

I've left quite a few legacy methods/jobs in place for the sake of keeping things tight, but several follow ups are imminent including the refresher job for uphold itself which is mostly copy paste as well as a lot of deleted code.

tsmartt (approved):

PR #3758, opened 2022-06-22
poc: Unit test scaffold

OAuthBringer:
1. Moves the core confirmation logic into the publisher model, adds scaffold of specs in the model.
2. Implements the method in the controller.

### DIFF
--- a/test/models/publisher_test.rb
+++ b/test/models/publisher_test.rb
@@ -17,6 +17,42 @@ class PublisherTest < ActiveSupport::TestCase
     Rails.application.secrets[:api_eyeshade_offline] = @prev_offline
   end
 
+  describe "#confirm_pending_email!" do
+    let(:publisher) { publishers(:default) }
+    let(:confirm_email) { nil }
+
+    describe "confirm_email" do
+      describe "pending_email" do
+        describe "!email" do
+          it "should set email to pending_email" do
+          end
+        end
+
+        describe "confirm_email == pending_email" do
+          it "should set email to pending_email" do
+          end
+
+          it "should set pending_email to nil" do
+          end
+        end
+
+        describe "confirm_email != pending_email" do
+          it "should not change email" do
+          end
+
+          it "should set pending_email to nil" do
+          end
+        end
+      end
+
+      describe "!pending_email" do
+      end
+    end
+
+    describe "!confirm_email" do
+    end
+  end
+
   test "verified publishers have both a name and email and have agreed to the TOS" do
     publisher = Publisher.new
     refute publisher.email_verified?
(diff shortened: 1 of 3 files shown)


OAuthBringer:
@thypon This is my suggestion for specs.  In general the email confirmation logic is most appropriate for the publisher model itself as it is a recordkeeping task.

The scaffold spec I filled out demonstrates mine and @tsmartt 's concerns about not having tests for this particular piece.  The number of states to consider and reason about is actually quite high and there are a few cases in there where I'm not sure what the desired outcome is.  If you could flesh out the scaffold with specs I could thumbs up the bugfix.

Our test examples are kind of all over the place, but if you need any assistance with handling the fixture state I can provide it.

test/models/publisher_test.rb
```diff
@@ -17,6 +17,42 @@ class PublisherTest < ActiveSupport::TestCase
...
   end
 
+  describe "#confirm_pending_email!" do
+    let(:publisher) { publishers(:default) }
+    let(:confirm_email) { nil }
+
+    describe "confirm_email" do
+      describe "pending_email" do
+        describe "!email" do
+          it "should set email to pending_email" do
```
thypon, line 27: Set the email to the pending one, cleanup the pending email.
This case is the initial one, when an email was not completely set-up and the user is still "pending"

test/models/publisher_test.rb
```diff
@@ -17,6 +17,42 @@ class PublisherTest < ActiveSupport::TestCase
...
+          end
+        end
+
+        describe "confirm_email == pending_email" do
+          it "should set email to pending_email" do
```
thypon, line 32: Only set the email if pending_email is == to the confirm email HTTP parameter.
This is an extra safeguard against attackers that got hold of the auth token, without knowing the email address.
It's also kinda unneeded since we use UUIDs. It works as an extra security layer.

test/models/publisher_test.rb
```diff
@@ -17,6 +17,42 @@ class PublisherTest < ActiveSupport::TestCase
...
+          end
+
+          it "should set pending_email to nil" do
```
thypon, line 35: Not sure if we want to have that in this way. Maybe we want to set it to nil only when the email is changed,
otherwise, other auth logins will be greedy and delete the pending_email. It's not a super huge issue anyway and I would not change this unless it's needed.

test/models/publisher_test.rb
```diff
@@ -17,6 +17,42 @@ class PublisherTest < ActiveSupport::TestCase
...
+          end
+        end
+
+        describe "confirm_email != pending_email" do
+          it "should not change email" do
+          end
+
+          it "should set pending_email to nil" do
```
thypon, line 43: As above, the opposite case

thypon (commented):
.

PR #3764, opened 2022-06-23
feat: Uphold.users.get_capability

OAuthBringer:
Adds basic client support.  I have no real ability to test this ATM because of some temporary account limitations with uphold sandbox.


OAuthBringer:
@jlbyrne Can I get a peak at this?  It's just an expansion of the existing client.  I want to test this out in staging before rolling out the full implementation

jlbyrne:
Sure!  Looking now.

PR #3765, opened 2022-06-24
Fix/confirm email

OAuthBringer:
Fixes the param issue and adds minimal specs for the method in question.

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -110,14 +110,12 @@ class PublishersController < ApplicationController
 
   def ensure_email
     token = params[:token]
-    confirm_email = params[:confirm_email]
-
     publisher = Publisher.find(params[:id])
 
-    if PublisherTokenAuthenticator.new(publisher: publisher, token: token, confirm_email: confirm_email).perform(consume: false)
+    if token && PublisherTokenAuthenticator.new(publisher: publisher, token: token).perform(consume: false)
       @publisher = publisher
     else
-      flash[:alert] = t(".token_invalid")
+      flash[:alert] = t("shared.error")
       redirect_to expired_authentication_token_publishers_path(id: publisher.id)
     end
   end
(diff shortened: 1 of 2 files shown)


app/controllers/publishers_controller.rb
```diff
@@ -110,14 +110,12 @@ def protect
...
-    confirm_email = params[:confirm_email]
-
     publisher = Publisher.find(params[:id])
 
-    if PublisherTokenAuthenticator.new(publisher: publisher, token: token, confirm_email: confirm_email).perform(consume: false)
+    if token && PublisherTokenAuthenticator.new(publisher: publisher, token: token).perform(consume: false)
       @publisher = publisher
     else
-      flash[:alert] = t(".token_invalid")
+      flash[:alert] = t("shared.error")
```
OAuthBringer, line 118: I had to change this to a shared error because it was missing the JA translation and we don't have time for that.

jlbyrne (approved):

app/controllers/publishers_controller.rb
```diff
@@ -110,14 +110,12 @@ def protect
 
   def ensure_email
     token = params[:token]
-    confirm_email = params[:confirm_email]
-
     publisher = Publisher.find(params[:id])
 
-    if PublisherTokenAuthenticator.new(publisher: publisher, token: token, confirm_email: confirm_email).perform(consume: false)
+    if token && PublisherTokenAuthenticator.new(publisher: publisher, token: token).perform(consume: false)
```
OAuthBringer, line 115: This fails if token isn't present, and if the token isn't present it can't be valid so we throw an error.

test/controllers/publishers_controller_test.rb
```diff
@@ -32,6 +32,37 @@ class PublishersControllerTest < ActionDispatch::IntegrationTest
...
+    end
+
+    describe "when token" do
+      let(:token) { "token" }
+
+      before do
+      end
+
+      it "should redirect" do
+        PublisherTokenAuthenticator.any_instance.stubs(:perform).returns(true)
```
OAuthBringer, line 55: The previous failure was on initialization so even though the perform method is stubbed it catches the previous failure case.

thypon:
Closing this issue since it seems was an artifact introduced by d9921d4bffe730e16ae0ca1fed33b1d65723a529

PR #3766, opened 2022-06-24
Production Release: Oauth2 Refreshers and Email Confirmation Bugfix

OAuthBringer:
1. Adds refresher jobs, though only Gemini is currently scheduled
2. Fixes an arg error causing 500s for some users.

### DIFF
--- a/app/controllers/publishers_controller.rb
+++ b/app/controllers/publishers_controller.rb
@@ -110,14 +110,12 @@ class PublishersController < ApplicationController
 
   def ensure_email
     token = params[:token]
-    confirm_email = params[:confirm_email]
-
     publisher = Publisher.find(params[:id])
 
-    if PublisherTokenAuthenticator.new(publisher: publisher, token: token, confirm_email: confirm_email).perform(consume: false)
+    if token && PublisherTokenAuthenticator.new(publisher: publisher, token: token).perform(consume: false)
       @publisher = publisher
     else
-      flash[:alert] = t(".token_invalid")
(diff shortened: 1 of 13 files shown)


jlbyrne (approved):
Nothing like a Friday afternoon hotfix!

OAuthBringer:
Given that I shipped a rewrite to the core wallet authentication mechanism for our biggest wallet provider today, a few 500s for a security patch bother me very little.

app/controllers/publishers_controller.rb
```diff
@@ -110,14 +110,12 @@ def protect
 
   def ensure_email
     token = params[:token]
-    confirm_email = params[:confirm_email]
-
     publisher = Publisher.find(params[:id])
```
thypon, line 113: I think we have still an issue here: this line should be converted to `publisher = Publisher.find(params[:id]).confirm_pending_email!(confirm_email)` to have `ensure_email` semantic equivalence.

OAuthBringer: @thypon Feel free to go ahead and make whatever changes you need to introduce the functionality you are looking for.  My goal was the fix the 500 impacting users in production.  As I already said, I literally have no idea what the intent of this section of the controller was, I simply knew what was causing the 500 error.  Please enhance the specs I laid out for this method as needed for whatever changes you make.

If you need to confirm an existing email in multiple places it may make the most sense to just create an isolated method that gets called in a callback or within the context of another method.

PR #3780, opened 2022-07-01
enhancement: Block if not capability

OAuthBringer:
Enforce requirement for "deposits" capability in order to create an uphold account.

### DIFF
--- a/app/models/uphold_connection.rb
+++ b/app/models/uphold_connection.rb
@@ -413,7 +413,7 @@ class UpholdConnection < Oauth2::AuthorizationCodeBase
 
     case result
     when UpholdUserCapability
-      true
+      result.requirements.empty? && result.restrictions.empty?
     else
       false
     end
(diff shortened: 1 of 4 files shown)


app/models/uphold_connection.rb
```diff
@@ -406,6 +408,17 @@ def find_or_create_uphold_card!
     end
   end
 
+  def has_deposit_capability?
+    result = connection_client.users.get_capability("deposits")
+
+    case result
+    when UpholdUserCapability
+      true
```
OAuthBringer, line 416: Hmm, I'm not actually sure this is valid. There are additional checks based on requirements/restrictions that inform whether or not they have the capability.

OAuthBringer: Following up with Uphold on this one

OAuthBringer:
Just as a note.  Given that I'm out on PTO tomorrow and there have been lots of support requests related to the last set of changes I made I decided it wise to just back off on this until I get back.  I don't want to deploy an entirely new set of changes and not be around to assist with any issues that arise.

app/models/uphold_connection.rb
```diff
@@ -406,6 +408,17 @@ def find_or_create_uphold_card!
...
   end
 
+  def has_deposit_capability?
+    result = connection_client.users.get_capability("deposits")
+
+    case result
+    when UpholdUserCapability
+      result.requirements.empty? && result.restrictions.empty?
+    else
+      false
```
OAuthBringer, line 418: This actually isn't the best way to do this.  I should really throw another exception here because this would only happen if the request failed or something.

tsmartt (approved):

jlbyrne (approved):

PR #3781, opened 2022-07-01
fix: Assign to correct value

OAuthBringer:
Refresh values weren't actually being assigned.  The various linters + encrypted attrs rekt me on this one.


OAuthBringer:
This is p1, expediting.

PR #3782, opened 2022-07-01
Production Release: P1 patch for failure to assign refresh token.

OAuthBringer:
FML on that one.  The sorbet and standardrb linters got over a fight because of that line and I made the wrong choice.


OAuthBringer:
p1, merging

PR #3805, opened 2022-07-27
Make Redshift pool same size as sidekiq concurrency

tsmartt:
Our configuration is incorrect, as we need to have a pool big enough for all of the sidekiq threads we have running on a node.


OAuthBringer (approved):
Makes sense. 👍

PR #3819, opened 2022-08-08
enhancement: Enforce existing judgements

OAuthBringer:
Enforces existing suspension when user attempts to reuse a suspended uphold account.   This is interpreted as an attempt to bypass a suspension.


OAuthBringer:
I'll close this for the moment, I submitted it initially because I had not pushed up everything and the tests were all passing.

PR #3831, opened 2022-08-29
feat: exclude blocked wallet addresses from PCDN

jlbyrne:
Addresses issue [107](https://github.com/brave-intl/creators-private-issues/issues/107).  See Link for more details

The first commit has the relevant changes.  The second commit is the massive output of running `standardrb --fix`.


OAuthBringer:
@jlbyrne Something is off with the `standardrb --fix`.  There's no way it should have updated 400+ files as it is required for the build to pass, did you run it in the container or on your local? I always run it in the container.

OAuthBringer (changes requested):
- [ ] Double check your standardrb config.  Something is wrong if you had to fix 460 files as there is no way you touched 460 files.
- [ ] Move rewards client to lib
- [ ] Remove mocks from client implementation and emulate `lib/uphold` and `lib/eyeshade` and relevant spec/mocking patterns.

OAuthBringer:
@jlbyrne Let's chat about this later today at least WRT to the pattern changes I've mentioned, it'll be easier to do on a chat than through text.

app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -23,6 +23,9 @@ def perform(prefix)
   def generate_brotli_encoded_channel_response(prefix:)
     @site_banner_lookups = SiteBannerLookup.where("sha2_base16 LIKE ?", prefix + "%")
     channel_responses = PublishersPb::ChannelResponseList.new
+
+    allowed_regions = Rewards::Parameters.new.get_parameters.custodianRegions
```
OAuthBringer, line 27: @jlbyrne You should handle the failure case here.  Possible outputs are the expected response struct or the Faraday::Response object.  This will throw a no method error for any non-success request.

app/services/base_api_client.rb
```diff
@@ -138,9 +138,12 @@ def parse_response_to_struct(response, struct)
 
   def adapt_to_struct(struct, obj)
     out = {}
-
     struct.props.keys.each do |key|
-      out[key] = obj.fetch(key, nil)
+      if obj.fetch(key, nil).is_a?(Hash)
+        out[key] = adapt_to_struct(struct.props[key][:type], obj.fetch(key, nil))
```
OAuthBringer, line 143: oh my, much recursion.

OAuthBringer: @jlbyrne What is the `:type` symbol?

jlbyrne: If you take a look at the struct, each key has its type listed:
```
{:uphold=>{:immutable=>true, :type=>Rewards::Types::AllowList, :type_object=>#<T::Types::Simple:0x0000aaab0fe2a850 @raw_type=Rewards::Types::AllowList>, :accessor_key=>:@uphold, :sensitivity=>nil, :pii=>nil, :extra=>nil, :setter_proc=>#<Proc:0x0000aaab0fe299a0 /usr/local/bundle/gems/sorbet-runtime-0.5.9538/lib/types/props/private/setter_factory.rb:61>, :serialized_form=>"uphold", :fully_optional=>false, :need_nil_read_check=>nil}
```
I'm using that to set the value of struct in the recursive call.  Unfortunately, there's no good way to extract whether or not the key's type is a primitive like a string, or something custom.

The recursion is really unfortunate, but I don't see another way of handling a utility function for any kind of nested JSON. 😢

app/jobs/cache/browser_channels/responses_for_prefix.rb
jlbyrne: Good call.  Still fixing some failing test and linting issues, but I'll add that in.

app/services/base_api_client.rb
OAuthBringer: We're not going to be able to hard code the `:type` symbol in that method as it is a global handler.  It will look for a type value in all subsequent hash responses.

OAuthBringer: Let me think about it.  The response object for params is rough.

OAuthBringer:
@jlbyrne Looks like you deleted the gemfile.lock and it's not allowing me to run tests on your branch atm.

OAuthBringer:
I got around it, but man the different Gemfile based on device thing is really freaking annoying.

jlbyrne:
Crap- I can cut/paste the original gemfile from master and re-include it.  I think that has all the necessary changes to standard rb

app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -23,6 +23,17 @@ def perform(prefix)
...
     @site_banner_lookups = SiteBannerLookup.where("sha2_base16 LIKE ?", prefix + "%")
     channel_responses = PublishersPb::ChannelResponseList.new
+
+    parameters = Rewards::Parameters.new.get_parameters
+
+    case parameters
+    when Rewards::Types::ParameterResponse
+      allowed_regions = parameters.custodianRegions
+    else
+      LogException.perform(failure)
```
jlbyrne, line 33: It looks like we're generally using sentry to log errors from the jobs folder.  However, the LogException service will log errors to both sentry and new relic.

app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -23,6 +23,17 @@ def perform(prefix)
...
     @site_banner_lookups = SiteBannerLookup.where("sha2_base16 LIKE ?", prefix + "%")
     channel_responses = PublishersPb::ChannelResponseList.new
+
+    parameters = Rewards::Parameters.new.get_parameters
+
+    case parameters
+    when Rewards::Types::ParametersResponse
+      allowed_regions = parameters.custodianRegions
+    else
+      LogException.perform(failure)
```
OAuthBringer, line 33: @jlbyrne If you just raise the error it will automatically get logged to newrelic.

OAuthBringer: Other than that looks good.

OAuthBringer (approved):

app/jobs/cache/browser_channels/responses_for_prefix.rb
OAuthBringer: I've never logged into sentry, @tsmartt do we still use this?  I've only ever looked at newrelic

app/jobs/cache/browser_channels/responses_for_prefix.rb
OAuthBringer, line 33: Er, except wait holdon.  Where is `failure` defined? I would expect this to throw an unhandled exception.

PR #3836, opened 2022-09-07
Disable stats job for now

tsmartt:


OAuthBringer:
@tsmartt I don't think this is a scheduled task, it's 

`Sync::PromoRegistrationStatsJob`

PR #3837, opened 2022-09-08
Production Release: Update Gemini Oauth2 flow

OAuthBringer:
Migrates gemini oauth2 connection logic to modernized implementation.  Should resolve bad state check issue reported by @thypon


OAuthBringer:
@thypon 

State verification is performed: https://github.com/brave-intl/publishers/blob/staging/app/controllers/oauth2_controller.rb#L106-L121

Sets and encrypted, http only cookie that expires in 90 seconds.  Cookie is set on redirect to oauth2 provider.

PR #3838, opened 2022-09-09
fix: Proxies are being phased out as most services are now off of heroku

OAuthBringer:
I believe this should resolve the issue with promo/referrals.  The promo side has been fixed/confirmed in production but we are still getting timeouts in the service, I believe it is because of this.  Will test/evaluate in staging.


jlbyrne (approved):
yolo 🤣

PR #3842, opened 2022-09-12
task: Refactor to use single response type

OAuthBringer:
Consolidates type into a single structure that is compatible with the existing API client.


OAuthBringer:
Ignore me!!!!!

PR #3846, opened 2022-09-13
Production Release: Promo stats job performance enhancements

OAuthBringer:
Sniff tested this in staging, basics looked solid, will need to check in prod on a test set to get a real sense of resolve.


OAuthBringer:
`#notquiteyolo`

PR #3850, opened 2022-09-14
fix: This is why we shouldn't write services this way

OAuthBringer:
Who knows how many instances of this are present in creators.  Http clients should be clients that handle the entirety of their configuration, services should implement clients, models, and other services and mediate their interactions.

### DIFF
--- a/app/services/promo/unattached_registrar.rb
+++ b/app/services/promo/unattached_registrar.rb
@@ -49,6 +49,10 @@ class Promo::UnattachedRegistrar < BaseApiClient
 
   private
 
+  def proxy_url
+    nil
+  end
+
   def api_base_uri
     Rails.application.secrets[:api_promo_base_uri]
   end


OAuthBringer:
I will find the ven diagram of yours and my perspectives and say, "We should avoid inheritance from classes designed without respecting core OOP principles".

OAuthBringer:
I really should not have used that base_api client and should have just created a new one but I took the easy route.

OAuthBringer:
None of this is my doing however.

OAuthBringer:
Actually, no inheritance isn't the issue.  It's a failure to seperate concerns.  If all the promos http endpoints were configured in a single client and then called from the services that needed them then I would have fixed the issue in all cases the first time.  The issue here is that the author entangled a service and an http client in one, which was a pattern repeated across many of the promo services.

PR #3859, opened 2022-09-21
Replace references to sentry with LogException

jlbyrne:
Replaces sentry error logging with LogException, which logs to both sentry and new relic.  Sentry can be explicitly removed in another ticket.


OAuthBringer (approved):
Error rate is going to skyrocket.

jlbyrne:
> Error rate is going to skyrocket.

😭 True

PR #3861, opened 2022-09-27
Remove country check from bitflyer and fallbacks for gemini and uphol…

jlbyrne:
Bitflyer accounts, if valid, will always have a Japanese locale.  Since we're not storing the country code for bitflyer connections, I'm removing the country check for the pCDN and the unusable `japanese_account` method.  

I'm also removing the empty string fallback for missing addresses.  If an address is missing, we do not want the creator showing up as verified with a blank address.

### DIFF
--- a/app/jobs/cache/browser_channels/responses_for_prefix.rb
+++ b/app/jobs/cache/browser_channels/responses_for_prefix.rb
@@ -49,10 +49,10 @@ class Cache::BrowserChannels::ResponsesForPrefix
 
           if !Rails.env.production?
             if connection.country && allowed_regions[:uphold][:allow].include?(connection.country.upcase)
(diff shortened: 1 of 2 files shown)


app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -83,10 +76,10 @@ def generate_brotli_encoded_channel_response(prefix:)
 
           if !Rails.env.production?
             if connection.country && allowed_regions[:gemini][:allow].include?(connection.country.upcase)
-              gemini_wallet.address = site_banner_lookup.channel.gemini_connection&.recipient_id || connection.recipient_id || ""
```
OAuthBringer, line 86: I see.

OAuthBringer (approved):

Issue #3789, opened 2022-07-19
Remove "Deposit currency" elements in Creators dashboard


Miyayes:
## Description

We should remove "Deposit currency: " and "change" elements in the Creators dashboard, now that it is always BAT.

![image](https://user-images.githubusercontent.com/11497541/179803428-8b4d3499-36ff-48cb-9378-e935d9df7846.png)

OAuthBringer:
I realized this may be more complicated than I'd initially realized.  It's not entirely accurate that "BAT" is the only value.  I didn't migrate users who have another currency selected because of a (reasonable) fear of unknown consequences WRT data integrity/API integration issues.  Technically I'd need to make it so this is conditionally shown if the user does not have BAT set as their default.  

I have a PR that removes these elements but not a conditional. Because of this I'm going to punt on this issue for a minute and try to address the other AF/data integrity issues we discussed earlier. cc @Miyayes

jlbyrne:
Some notes:
- the `supported_currencies` method for all three connection type was only being used for the uphold currency change dropdown, so that can be deleted
- The validation I added should prevent anyone from saving a default currency to an uphold connection that isn't BAT or USD.  The validation will only run if `default_currency` has been changed, so the uphold connections with random currencies in the database should be otherwise unaffected.

PR #3867, opened 2022-10-11
remove currency select dropdown and related back end code

jlbyrne:
References: #3789


app/controllers/payment/connection/currencies_controller.rb
```diff
@@ -1,37 +0,0 @@
-# typed: ignore
-
-module Payment
-  module Connection
-    class CurrenciesController < ApplicationController
```
tsmartt, line 5: The best kind of code, code no more!

tsmartt (approved):

PR #3871, opened 2022-10-19
add in api endpoint and test

jlbyrne:
Creator's portion of https://github.com/brave-intl/eyeshade-ops/issues/109#issue-1329032579.  The endpoint will change slightly as https://github.com/brave-intl/creators-private-issues/issues/115 is added in, but this should be enough to allow development on the Eyeshade portion.

### DIFF
--- a/app/controllers/api/v3/public/channels_controller.rb
+++ b/app/controllers/api/v3/public/channels_controller.rb
@@ -11,6 +11,39 @@ class Api::V3::Public::ChannelsController < Api::V3::Public::BaseController
     render(json: statistical_totals_json, status: 200)
   end
 
+  # Takes an array of channel identifiers and retuns a dictionary of channel identifiers as keys and true/false as values
+  def allowed_countries
+    channels = params[:channel_ids].map do |id|
+      {channel: Channel.find_by_channel_identifier(id), channel_identifier: id}
+    end
+
+    parameters = Rewards::Parameters.new.get_parameters
+
+    case parameters
+    when Rewards::Types::ParametersResponse
+      allowed_regions = parameters.custodianRegions
+    else
+      LogException.perform(parameters)
+      raise StandardError.new("Could not load allowed regions")
+    end
+
+    response = {}
+    channels.each do |channel_obj|
+      publisher = channel_obj[:channel].publisher
+
+      response[channel_obj[:channel_identifier]] = if publisher.uphold_connection.present?
+        allowed_regions.include?(publisher.uphold_connection.country)
+      elsif publisher.gemini_connection.present?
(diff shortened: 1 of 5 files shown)


app/controllers/api/v3/public/channels_controller.rb
```diff
@@ -11,6 +11,39 @@ def totals
     render(json: statistical_totals_json, status: 200)
   end
 
+  # Takes an array of channel identifiers and retuns a dictionary of channel identifiers as keys and true/false as values
+  def allowed_countries
+    channels = params[:channel_ids].map do |id|
+      {channel: Channel.find_by_channel_identifier(id), channel_identifier: id}
+    end
+
+    parameters = Rewards::Parameters.new.get_parameters
```
jlbyrne, line 20: This is getting repetitive, will look to clean it up in the upcoming related tickets.

app/controllers/api/v3/public/channels_controller.rb
```diff
@@ -11,6 +11,39 @@ def totals
     render(json: statistical_totals_json, status: 200)
   end
 
+  # Takes an array of channel identifiers and retuns a dictionary of channel identifiers as keys and true/false as values
+  def allowed_countries
+    channels = params[:channel_ids].map do |id|
```
tsmartt, line 16: I would use a single query here https://stackoverflow.com/a/39035889

app/controllers/api/v3/public/channels_controller.rb
```diff
@@ -11,6 +11,39 @@ def totals
...
+    else
+      LogException.perform(parameters)
+      raise StandardError.new("Could not load allowed regions")
+    end
+
+    response = {}
+    channels.each do |channel_obj|
+      publisher = channel_obj[:channel].publisher
+
+      response[channel_obj[:channel_identifier]] = if publisher.uphold_connection.present?
```
tsmartt, line 34: In the query, make sure you preload the items you need. You can use strict loading like [here](https://github.com/brave-intl/publishers/blob/fd712a808b18ac295462ff66f2816e70a35bc560/app/services/enqueue_publishers_for_payout_service.rb#L65)

tsmartt (changes requested):

app/controllers/api/v3/public/channels_controller.rb
jlbyrne: Created issue to resolve the `channel_identifier` problem: https://github.com/orgs/brave-intl/projects/13#card-86385813

tsmartt:
Ticket for the performance issues https://github.com/brave-intl/publishers/pull/3871

PR #3900, opened 2022-11-01
Don't include Potential Payments from wallets in unallowed countries

tsmartt:
Ensure that Potential Payments don't get generated for wallets in unallowed countries.

Closes https://github.com/brave-intl/creators-private-issues/issues/108

### DIFF
--- a/app/services/payout/service.rb
+++ b/app/services/payout/service.rb
@@ -15,7 +15,14 @@ module Payout
     # Internal: Creates entries for any reason for not paying out a publisher.
     #
     # Returns a boolean
-    def skip_publisher?(payout_report:, publisher:)
+    def skip_publisher?(payout_report:, publisher:, allowed_regions: [], connection: nil)
+      if !allowed_regions.blank? && connection.present?
+        if !allowed_regions.include?(connection.country&.upcase)
+          create_message(message: "Publisher's wallet connection is in an unallowed country: #{connection.country&.upcase}. Allowlist: #{allowed_regions}", payout_report: payout_report, publisher: publisher)
+          return true
+        end
+      end
+
       if !publisher.has_verified_channel?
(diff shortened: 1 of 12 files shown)


app/services/payout/service.rb
```diff
@@ -15,7 +15,14 @@ def create_message(payout_report:, publisher:, message:)
     # Internal: Creates entries for any reason for not paying out a publisher.
     #
     # Returns a boolean
-    def skip_publisher?(payout_report:, publisher:)
+    def skip_publisher?(payout_report:, publisher:, allowed_regions: [], connection: nil)
+      if !allowed_regions.blank? && connection.present?
+        if !allowed_regions.include?(connection.country&.upcase)
```
jlbyrne, line 20: There should be a method `valid_country?` on all the connection models.  You could use that if you don't want to have to pass the allowed regions through the various services.  I think it's also fine as is, though.

test/services/enqueue_publishers_for_payout_service_test.rb
```diff
@@ -3,8 +3,14 @@
 require "test_helper"
 
 class EnqueuePublishersForPayoutServiceTest < NoTransactDBBleanupTest
+  include MockRewardsResponses
```
jlbyrne, line 6: At some point we should look into some test-suite-wide before blocks.  We're having to stub out the rewards responses in a million tests.

jlbyrne (approved):

app/services/payout/service.rb
tsmartt: For this service, we need to run the allowed countries call once so that we don't hit the API or even cache more than once. So we do that at the top of the service and pass the results down through the services. This makes for easier dependency injection testing and also allows us not to n+1 the calls. That method does seem convenient, though!

PR #3902, opened 2022-11-02
fix: allow endpoint to return values when a channel identifier is not…

jlbyrne:
Addresses https://github.com/brave-intl/creators-private-issues/issues/121


tsmartt:
If we don't recognize the channel, I vote for returning false. cc @Miyayes

tsmartt (approved):

PR #3913, opened 2022-11-09
Don't include the address in pcdn if it doesn't exist on our end

tsmartt:
Resolves [this error](https://one.newrelic.com/nr1-core/errors/overview/MjYwNDMwOXxBUE18QVBQTElDQVRJT058NTkyOTgwNzA3?account=2604309&duration=10800000&state=47bb13ec-0f8d-e62c-0b06-dc68f2a39fe9)

Resolves: https://github.com/brave-intl/publishers/issues/3914


jlbyrne:
Can we actually leave this error in for now?  it's not causing any problems besides creating notifications in new relic, which is helpful for debugging the card privacy issue I'm working on.

PR #3918, opened 2022-11-21
Create Uphold cards for those who should have them

tsmartt:
In the case that creating the Uphold Cards fails, due to a flake on the part of the Uphold API, we should ensure that these are created in a timely manner. This job runs nightly and checks for those who have verified channels and Uphold connections, but who don't have cards. It then launches a job to create them.

Closes https://github.com/brave-intl/creators-private-issues/issues/126


jlbyrne (approved):
This is a really good catch, but we might want to look at uphold connections that were updated recently rather than publishers that have signed in recently?  Publisher sign in might be a good enough grouping though.

PR #3921, opened 2022-11-22
remove account-level fallback for gemini recipient ids

jlbyrne:
Addresses https://github.com/brave-intl/creators-private-issues/issues/125


tsmartt (approved):
LG2M

PR #3926, opened 2022-11-28
We should remove the address field from the pCDN wallet if the addres…

tsmartt:
…s is nil

### DIFF
--- a/app/jobs/cache/browser_channels/responses_for_prefix.rb
+++ b/app/jobs/cache/browser_channels/responses_for_prefix.rb
@@ -40,7 +40,10 @@ class Cache::BrowserChannels::ResponsesForPrefix
           uphold_wallet.wallet_state = get_uphold_wallet_state(uphold_connection: connection)
 
           if connection.country && allowed_regions[:uphold][:allow].include?(connection.country.upcase)
-            uphold_wallet.address = site_banner_lookup.channel.uphold_connection&.address
+            uphold_address = site_banner_lookup.channel.uphold_connection&.address
+            if uphold_address
+              uphold_wallet.address = uphold_address
+            end
           end
 
           wallet.uphold_wallet = uphold_wallet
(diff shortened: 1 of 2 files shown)


app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -63,7 +66,10 @@ def generate_brotli_encoded_channel_response(prefix:)
           gemini_wallet.wallet_state = get_gemini_wallet_state(gemini_connection: connection)
 
           if connection.country && allowed_regions[:gemini][:allow].include?(connection.country.upcase)
-            gemini_wallet.address = site_banner_lookup.channel.gemini_connection&.recipient_id
+            gemini_address = site_banner_lookup.channel.gemini_connection&.recipient_id
```
OAuthBringer, line 69: @tsmartt Here as well.

OAuthBringer:
@tsmartt Why would we include the channel in the pCDN at all if it has no address?

OAuthBringer (changes requested):
I'm not sure where the change is, but we should revert to having `address` set to empty string and not dropped as it would likely break everything.

app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -40,7 +40,10 @@ def generate_brotli_encoded_channel_response(prefix:)
           uphold_wallet.wallet_state = get_uphold_wallet_state(uphold_connection: connection)
 
           if connection.country && allowed_regions[:uphold][:allow].include?(connection.country.upcase)
-            uphold_wallet.address = site_banner_lookup.channel.uphold_connection&.address
+            uphold_address = site_banner_lookup.channel.uphold_connection&.address
```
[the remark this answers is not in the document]
tsmartt: Resolved

app/jobs/cache/browser_channels/responses_for_prefix.rb
tsmartt: Resolved

tsmartt:
> I'm not sure where the change is, but we should revert to having `address` set to empty string and not dropped as it would likely break everything.

They are now included, but with a blank string

OAuthBringer (approved):

PR #3950, opened 2022-12-14
Fix/cleanup uphold card fixes

jlbyrne:
Addresses https://github.com/brave-intl/creators-private-issues/issues/142 and https://github.com/brave-intl/creators-private-issues/issues/145

### DIFF
--- a/app/controllers/admin/publishers_controller.rb
+++ b/app/controllers/admin/publishers_controller.rb
@@ -118,7 +118,6 @@ class Admin::PublishersController < AdminController
     connection = UpholdConnection.find_by(publisher: params[:publisher_id])
     if connection.present?
       connection.sync_connection!
-      connection.create_uphold_cards
     end
     redirect_to admin_publisher_path(@publisher.id)
   end
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -366,6 +366,10 @@ class Channel < ApplicationRecord
     site_banner_lookup.sync!
   end
 
+  def has_valid_uphold_connection?
+    uphold_connection&.address && uphold_connection&.card_id
+  end
+
   private
 
   def should_register_channel_for_promo?
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -391,8 +391,12 @@ class Channel < ApplicationRecord
   end
 
   def create_uphold_channel_card
-    return if !publisher&.uphold_connection&.uphold_id
+    return if !publisher&.uphold_connection&.uphold_id && !publisher&.uphold_connection&.can_create_uphold_cards?
     CreateUpholdChannelCardJob.perform_later(uphold_connection_id: publisher.uphold_connection&.id, channel_id: id)
+  end
+
+  def has_valid_uphold_connection?
+    uphold_connection && (publisher.uphold_connection&.id == uphold_connection&.uphold_connection_id) && uphold_connection&.address && uphold_connection&.card_id
   end
 
   def notify_slack
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -367,7 +367,7 @@ class Channel < ApplicationRecord
   end
 
   def has_valid_uphold_connection?
-    (publisher.uphold_connection&.id == uphold_connection&.uphold_connection_id) && uphold_connection&.address && uphold_connection&.card_id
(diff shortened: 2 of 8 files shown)


app/controllers/admin/publishers_controller.rb
```diff
@@ -118,7 +118,6 @@ def refresh_uphold
     connection = UpholdConnection.find_by(publisher: params[:publisher_id])
     if connection.present?
       connection.sync_connection!
-      connection.create_uphold_cards
```
jlbyrne, line 121: `sync_connection!` previously conditionally called `create_uphold_cards`.  Since it now calls `create_uphold_cards` whenever there's a successful result, this line is unnecessary.

app/models/gemini_connection.rb
```diff
@@ -199,11 +200,5 @@ def oauth2_config
     def encryption_key(key: Rails.application.secrets[:attr_encrypted_key])
       [key].pack("H*")
     end
-
```
jlbyrne, line 202: This is an oversight from other work.  The method was moved to the `Oauth2::AuthorizationCodeBase` parent, and is unnecessary here.

app/models/channel.rb
```diff
@@ -351,6 +351,10 @@ def update_site_banner_lookup!(skip_site_banner_info_lookup: false)
     site_banner_lookup.sync!
   end
 
+  def has_valid_uphold_connection?
```
tsmartt, line 354: What do you think of renaming this to `has_valid_uphold_connection_for_cards?`
We'll also want to make sure that the ID of the card is the same ID of the current `selected_wallet_provider` on the `publisher` If it doesn't, we should launch a `CreateUpholdChannelCardJob` job and have that create cards with the current `UpholdConnection`

app/models/gemini_connection.rb
```diff
@@ -97,6 +97,7 @@ def fetch_refresh_token
   # as this is not a valid Oauth2 spec response and thus
   # I do not want to contaminate the Oauth2::AuthenticationCodeClient
   def refresh_authorization!
+    return self if !access_token_expired?
```
tsmartt, line 100: nice!

app/models/oauth2/authorization_code_base.rb
```diff
@@ -62,30 +62,32 @@ def refresh_authorization!(&blk)
       return BFailure.new(errors: ["Connection refresh has already failed"])
     end
 
-    refresh_token = fetch_refresh_token
+    T.must(self.with_lock do
```
tsmartt: nice! A row-level pessimistic lock to prevent 2 jobs from contacting the provider at the same time. Makes this operation atomic. ACID for the win!

tsmartt (changes requested):
Nice, had a few questions.

tsmartt (approved):
Will release this separately from other items since it's a tricky change with the lock.

app/models/channel.rb
```diff
@@ -351,6 +351,10 @@ def update_site_banner_lookup!(skip_site_banner_info_lookup: false)
...
+    (publisher.uphold_connection&.id == uphold_connection&.uphold_connection_id) && uphold_connection&.address && uphold_connection&.card_id
```
tsmartt, line 355: Just saw this, what if they both are `nil`?
`publisher.uphold_connection&.id == uphold_connection&.uphold_connection_id`
We need to check that at least one of them exists

tsmartt (changes requested):

app/models/oauth2/authorization_code_base.rb
```diff
@@ -69,20 +69,28 @@ def refresh_authorization!(&blk)
       return BFailure.new(errors: ["Cannot refresh without refresh token"])
     end
 
-    result = self.class.oauth2_client.refresh_token(refresh_token)
+    # When we upgrade to rails 7, transaction blocks will silently roll back when either return or raise is called
+    result = T.let(nil, T.nilable(T.any(Oauth2::Responses::RefreshTokenResponse, Oauth2::Errors::UnknownError, Oauth2::Responses::ErrorResponse)))
+    T.must(with_lock do
+      result = self.class.oauth2_client.refresh_token(refresh_token)
+      result.is_a?(RefreshTokenResponse) ? update_access_tokens!(result) : record_refresh_failure!
```
porteron, line 76: Is there some context to when this might fail or has been failing?

porteron (approved):
👍

tsmartt (approved):

PR #3956, opened 2022-12-17
Mark wallets as not payable which are not

tsmartt:
If there is a payment failure, slurp up the failure file and mark wallets which failed payout as not payable. This will show an error message when the creator logs in and sees his wallet in a bad state, telling him to disconnect and reconnect.


app/models/uphold_connection.rb
```diff
@@ -91,6 +91,15 @@ class UpholdAccountState
     with_active_connection.with_expired_tokens
   }
 
+  scope :payable, -> {
+    joins(:publisher)
+      .with_active_connection
```
tsmartt, line 96: This puts all the scopes for all wallet providers using the same `payable` scope for consistency.

We also add `with_active_connection`, so that if we can't refresh your connection, you are not payable.

app/models/bitflyer_connection.rb
```diff
@@ -90,6 +94,11 @@ def refresh_authorization!
     end
   end
 
+  # this is a hash of the account_id
+  def wallet_provider_id
```
tsmartt, line 98: Add these methods for consistency, and rely on the same method call to determine if the passed file's provider ID matches ours.

app/services/publisher_payout_feedback_status_updater.rb
```diff
@@ -0,0 +1,29 @@
...
+    by_publisher = to_update_list.group_by { |i| i["owner"] }
+    by_publisher.each do |publisher_str, channels|
+      Rails.logger.info("Processing #{publisher_str} with channels: #{channels.map { |c| c["publisher"] }.join(";")}")
+      publisher = Publisher.where(id: publisher_str.sub("publishers#uuid:", "")).first
+      if publisher.blank?
+        Rails.logger.info("Couldn't find publisher #{publisher_str}!")
+        next
+      end
+      publisher_wp_id = publisher.selected_wallet_provider&.wallet_provider_id
+      provided_wp_id = channels[0]["walletProviderId"]
```
jlbyrne, line 18: Is this the channel's address?  I think that _should_ be different from the publisher's address, right? Just want to make sure I understand what we're checking.

jlbyrne (approved):

PR #3958, opened 2022-12-19
Fix/gemini card issues

jlbyrne:
Addresses https://github.com/brave-intl/creators-private-issues/issues/153


porteron (approved):
looks good

PR #3962, opened 2022-12-29
Allow retries for card creation jobs

jlbyrne:
Addresses https://github.com/brave-intl/creators-private-issues/issues/159


porteron (approved):
Looks good

PR #3979, opened 2023-01-25
Remove sorbet

tsmartt:
Resolves https://github.com/brave-intl/creators-private-issues/issues/172

Sorbet was introduced to provide some assurance of correctness in the codebase, and it did aid in helping us structure a few things differently. But it's quite a flawed system, doesn't have near the flexibility or maturity of typescript, and slows down development by a factor of 2 or more. The tests cover our use cases, so to improve development velocity, I am ripping this out.


app/models/channel.rb
```diff
@@ -57,32 +57,19 @@ class Channel < ApplicationRecord
...
+  validate :verified_duplicate_channels_must_be_contested, if: -> { verified? }
+
+  after_save :notify_slack, if: -> { saved_change_to_verified? && verified? }
+
+  before_save :set_derived_brave_publisher_id, if: -> { derived_brave_publisher_id.nil? }
 
   # *ChannelDetails get autosaved from above.
-  after_save :update_site_banner_lookup!, if: -> {
-                                                T.bind(self, Channel)
-                                                :saved_change_to_verified? && verified?
```
tsmartt, line 79: The `:saved_change_to_verified?` will always return true since it's a symbol. But the method is never called, so this will always update the site banner, even if no change to verification has happened.

jlbyrne (approved):

PR #3987, opened 2023-02-01
Remove communicating through proxy to internal services

tsmartt:
Closes https://github.com/brave-intl/creators-private-issues/issues/116
Attempted [here](https://github.com/brave-intl/publishers/pull/3838) as well, will make sure other services will work this time.


jlbyrne (approved):

tsmartt:
Tested in staging to be OK, but prod eyeshade needs to get this to prod first before this goes to prod https://github.com/brave-intl/bat-ledger/pull/1106

PR #4012, opened 2023-03-01
Not all connections are refreshing due to this limit

tsmartt:
We previously discovered not all Uphold wallets were regularly refreshing because of this limit. For now, remove it.


jlbyrne (approved):

porteron (commented):
looks good

PR #4014, opened 2023-03-01
Fix/missing address

tsmartt:
Also mark errors passed from Uphold/Gemini during connection as expected. If Uphold says "your account is restricted b/c XYZ", that's not an exception we care about.


porteron (approved):
👍

porteron (approved):

PR #4023, opened 2023-03-21
Add in new p2p data models and adjust channel model

jlbyrne:
Resolves https://github.com/brave-intl/creators-private-issues/issues/140

### DIFF
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -32,6 +32,7 @@ class Channel < ApplicationRecord
   has_one :promo_registration, dependent: :destroy
   has_many :uphold_connection_for_channel, dependent: :destroy
   has_many :gemini_connection_for_channel, dependent: :destroy
+  has_many :crypto_address_for_channels, dependent: :destroy
 
   has_one :contesting_channel, class_name: "Channel", foreign_key: "contested_by_channel_id"
 
@@ -57,9 +58,14 @@ class Channel < ApplicationRecord
     in: %w[domain_not_found connection_failed too_many_redirects timeout no_txt_records token_incorrect_dns token_not_found_dns token_not_found_public_file no_https]
   }, allow_nil: true
 
+  validates :public_name, uniqueness: true, allow_nil: true
+  # validates :public_identifier, uniqueness: true, presence: true
+
   validate :site_channel_details_brave_publisher_id_unique_for_publisher, if: -> { details_type == "SiteChannelDetails" }
 
   validate :verified_duplicate_channels_must_be_contested, if: -> { verified? }
+
+  before_create :set_public_identifier
 
   after_save :notify_slack, if: -> { saved_change_to_verified? && verified? }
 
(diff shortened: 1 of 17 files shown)


app/models/channel.rb
```diff
@@ -57,10 +58,15 @@ class Channel < ApplicationRecord
     in: %w[domain_not_found connection_failed too_many_redirects timeout no_txt_records token_incorrect_dns token_not_found_dns token_not_found_public_file no_https]
   }, allow_nil: true
 
+  validates :public_name, uniqueness: true, allow_nil: true
+  # validates :public_identifier, uniqueness: true, presence: true
```
tsmartt, line 62: Comment

tsmartt: In ca you were wanting to uncomment this for uniqueness, nevermind. But you can also add an index in the migration to enforce uniqueness.

db/migrate/20230320213156_create_crypto_address_for_channels.rb
```diff
@@ -0,0 +1,13 @@
+class CreateCryptoAddressForChannels < ActiveRecord::Migration[7.0]
+  def change
+    create_table :crypto_address_for_channels, id: :uuid, default: -> { "uuid_generate_v4()"} do |t|
+      t.string :chain, null: false
```
tsmartt, line 4: We should consider an enum here https://github.com/bibendi/activerecord-postgres_enum

to enforce data integrity outside of the app

test/models/channel_test.rb
```diff
@@ -94,6 +94,29 @@ class ChannelTest < ActionDispatch::IntegrationTest
...
+    channel.save!
+
+    channel_2 = channels(:new_site)
+    channel_2.public_name = channel.public_name
+    refute channel_2.valid?
+
+    assert_equal "has already been taken", channel_2.errors.messages[:public_name][0]
+  end
+
+  # test "public_identifier is added on create and must be unique" do
```
tsmartt, line 109: Comments

tsmartt (commented):
Looks pretty good, left a few comments.

app/models/channel.rb
jlbyrne: This validation will work fine once I run the rake task to backfill values.  Until I do that, though, the validation will cause a bunch of errors.  I'll uncomment it in a future PR once the task has been run.

test/models/channel_test.rb
jlbyrne: Same issue as above- once the rake task has been run, I can uncomment the validation and the test will pass.

db/migrate/20230320213156_create_crypto_address_for_channels.rb
jlbyrne: that's a good idea, I'll make some adjustments.

porteron (approved):

porteron (approved):

PR #4041, opened 2023-04-07
add crypto addresses to pCDN

jlbyrne:
resolves https://github.com/brave-intl/creators-private-issues/issues/141


app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -166,6 +180,9 @@ def get_site_banner_details(site_banner_lookup)
       details[key.underscore] = value
     end
 
+    public_id = site_banner_lookup.channel.public_identifier
+    details.web3_url = public_id ? "#{ENV["CREATORS_HOST"]}/c/#{public_id}" : ""
```
zenparsing, line 184: The `web3_url` field should only be set if the creator can accept web3 contributions (so I guess that would mean they need at least one valid p2p crypto address).

jlbyrne:
Opened a clean branch to get around git issues

PR #4052, opened 2023-05-08
add crypto addresses to pCDN

jlbyrne:
resolves https://github.com/brave-intl/creators-private-issues/issues/141

### DIFF
--- a/app/jobs/cache/browser_channels/responses_for_prefix.rb
+++ b/app/jobs/cache/browser_channels/responses_for_prefix.rb
@@ -74,20 +74,6 @@ class Cache::BrowserChannels::ResponsesForPrefix
           wallet.gemini_wallet = gemini_wallet
           channel_response.wallets.push(wallet)
         end
-        # if !site_banner_lookup.channel.crypto_address_for_channels.sol_addresses.empty?
-        #   wallet = PublishersPb::Wallet.new
-        #   solana_wallet = PublishersPb::SolanaWallet.new
-        #   solana_wallet.address = site_banner_lookup.channel.crypto_address_for_channels.sol_addresses.first.crypto_address.address
-        #   wallet.solana_wallet = solana_wallet
-        #   channel_response.wallets.push(wallet)
-        # end
-        # if !site_banner_lookup.channel.crypto_address_for_channels.eth_addresses.empty?
-        #   wallet = PublishersPb::Wallet.new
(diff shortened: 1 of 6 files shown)


app/jobs/cache/browser_channels/responses_for_prefix.rb
```diff
@@ -74,6 +74,20 @@ def generate_brotli_encoded_channel_response(prefix:)
           wallet.gemini_wallet = gemini_wallet
           channel_response.wallets.push(wallet)
         end
+        # if !site_banner_lookup.channel.crypto_address_for_channels.sol_addresses.empty?
```
tsmartt, line 77: Comments

app/models/crypto_address_for_channel.rb
```diff
@@ -4,6 +4,16 @@ class CryptoAddressForChannel < ApplicationRecord
...
   SUPPORTED_CHAINS = ["ETH", "SOL"].freeze
   validates :chain, inclusion: {in: SUPPORTED_CHAINS}
-
   validates :channel_id, uniqueness: {scope: :chain}
+
+  scope :sol_addresses, -> { where(chain: "SOL") }
+  scope :eth_addresses, -> { where(chain: "ETH") }
+
+  def eth_address
+    eth_addresses.first.crypto_address.address
```
tsmartt, line 13: What about the case of possible `nil` values for anything in the string here. Should we return `nil` or expect the data to be there? In general, if the data is expected to be there by the time this is called, we should blow up here and ensure data integrity in the caller/system. But if it's expected this will be called and might be `nil`, we'll want to use `&.` for the chain here.

tsmartt (changes requested):
Few small things

tsmartt (approved):

protos/channel_responses.proto
```diff
@@ -57,12 +58,22 @@ enum GeminiWalletState {
...
+  string address = 1;
+}
+
 message Wallet {
   oneof provider {
     UpholdWallet uphold_wallet = 1;
     PaypalWallet paypal_wallet = 2;
     BitflyerWallet bitflyer_wallet = 3;
     GeminiWallet gemini_wallet = 4;
+    EthereumWallet ethereum_wallet = 5;
```
zenparsing, line 75: The client does not include these definitions - we should probably keep them in sync?

zenparsing (approved):
Protobuf changes 👍

tsmartt (approved):

PR #4062, opened 2023-05-23
Bitflyer translation for failed wallet

tsmartt:
Also Uphold & Bitflyer E2E tests

Followup for https://github.com/brave-intl/creators-private-issues/issues/1635

Waiting on Japanese translation approval

### DIFF
--- a/app/javascript/locale/ja.ts
+++ b/app/javascript/locale/ja.ts
@@ -82,6 +82,7 @@ export default {
     brave: {
       title: "BAT受け取り用ウォレット",
       description: "ウォレットを接続してチップなどを受け取れるようにします。",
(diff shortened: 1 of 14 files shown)


app/javascript/locale/ja.ts
```diff
@@ -82,6 +82,7 @@ export default {
     brave: {
       title: "BAT受け取り用ウォレット",
       description: "ウォレットを接続してチップなどを受け取れるようにします。",
+      paymentFailedWarning: "{custodian} アカウントへの接続に問題が見つかりました。 custodian アカウントに問題がないか確認し、再度接続をお試しください。",
```
jlbyrne, line 85: Is the second "custodian" supposed to be escaped?

jlbyrne (approved):
Think I found a small oversight, but looks good!

app/javascript/locale/ja.ts
tsmartt: good catch!

PR #4067, opened 2023-06-14
Feat/p2p dashboard UI

jlbyrne:
Resolves https://github.com/brave-intl/creators-private-issues/issues/139


porteron (approved):
looks great

app/controllers/crypto_address_for_channels_controller.rb
```diff
@@ -0,0 +1,97 @@
...
+        if success
+          render(json: {crypto_address_for_channel: success}, status: 200)
+        else
+          render(json: {errors: "address could not be updated"}, status: 400)
+        end
+      }
+    end
+  end
+
+  def verify_solana_address(signature, address, message)
```
kdenhartog, line 63: If you want an example of how we're handling this for brave-talk here's some good code to take a look at for the backend service: https://github.com/brave-intl/subscriptions/pull/312/files#diff-8593b22881785b8c7084d5f77364bcf5216445c2e32306da1513056e5d6b44e0R205-R246

app/javascript/views/cryptoWalletServices/CryptoWalletServices.jsx
```diff
@@ -0,0 +1,250 @@
...
+    if (results.publicKey) {
+      const pub_key = results.publicKey
+
+      const possibleMatch = this.state.solOptions.filter(sol => sol.value.address === pub_key);
+      if (possibleMatch.length > 0) {
+        this.updateAddress(possibleMatch[0]);
+        return;
+      }
+
+      const nonce = Date.now().toString();
```
kdenhartog, line 90: Utilizing the date as the message here would result in an active MITM attack (because they'd be able to see the message being sent to the creators service) being able to claim ownership of an address that doesn't belong to them. In this case, we'd want to be able to generate the message for them server side rather than on the front end client and then compare the signed value to the originally generated message.

If you want a good example of how this is being done with brave talk in the front end, you can take a look at it here: https://github.com/brave/brave-talk/blob/main/src/components/web3/api.ts#L178-L222

app/javascript/views/cryptoWalletServices/CryptoWalletServices.jsx
```diff
@@ -0,0 +1,250 @@
...
+        this.setErrorText('Solana connection was interrupted')
+        return;
+      }
+
+      axios({
+        method: 'post',
+        url: routes.publishers.cryptoAddressForChannels.create.replace('{channel_id}', this.channel.id),
+        data: {
+          chain: 'SOL',
+          account_address: pub_key,
```
kdenhartog, line 106:
```suggestion
          account_address: signedMessage.publicKey.toBase58(),
```

app/assets/images/eth-color.png
kdenhartog, line 1: BTW, we've got decent SVGs available in the wallet codebase if those would be more useful for you

app/javascript/views/cryptoWalletServices/CryptoWalletServices.jsx
```diff
@@ -0,0 +1,267 @@
...
+        return;
+      }
+
+      const nonce = await this.getNonce();
+      if (!nonce) {
+        this.setErrorText('An error occurred. Please try again later.')
+        return;
+      }
+
+      web3.eth.sign(nonce, address)
```
kdenhartog, line 150: nit: This library tends to be quite large so if you noticed a slowdown in responsiveness after this PR is merged I can help out with a follow up PR to remove it

app/controllers/crypto_address_for_channels_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+
+  def generate_nonce
+    render json: {nonce: SecureRandom.hex(32)}
+  end
+
+  def create
+    signature = params[:transaction_signature]
+    account_address = params[:account_address]
+    chain = params[:chain]
+    message = params[:message]
```
kdenhartog, line 26: We should make sure that the nonce we're generating and sending to the front end matches on verification here as well as checking to make sure it's the same user

Gemfile
```diff
@@ -172,6 +172,10 @@ gem "sprockets-rails", "3.4.2"
 gem "sprockets", "3.7.2"
 gem "uglifier"
 
+gem "eth", "~> 0.4"
+gem "rbnacl"
```
kdenhartog, line 176: no concerns from me for the addition of these dependencies. rbnacl calls down to libsodium which is a well known implementation that we already use

package.json
```diff
@@ -95,7 +96,8 @@
     "webpack-assets-manifest": "5",
     "webpack-cli": "4",
     "webpack-dev-server": "^4.15.0",
-    "webpack-merge": "5"
+    "webpack-merge": "5",
+    "web3": "^1.10.0"
```
kdenhartog, line 100: I noticed in some cases we're version locking here and some we're not. Is there any distinguishing pattern for one is used over the other here?

package.json
kdenhartog, line 1: no concerns from me on the updated dependencies here

kdenhartog (commented):
one more slight change to make sure that the user signing the message owns the account so that we're properly protected from replay attacks here. Other than that nonce related comments rest looks good to me

app/controllers/crypto_address_for_channels_controller.rb
bcaller: As I understand it `message` is supposed to be just the nonce. However, we are not checking anything about `message` so it could even be the empty string as long as that is signed correctly.

We should verify that the nonce (`message`) was generated by our backend and that it is not being reused.
Brave Talk, for instance, used Redis. The nonce generation endpoint SETs a key for the nonce and verification does the equivalent of GETDEL to check that it is a valid nonce and delete it so it cannot be reused.

app/controllers/crypto_address_for_channels_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  def index
+    current_channel = current_publisher.channels.find(params[:channel_id])
+    @crypto_addresses_for_channel = CryptoAddressForChannel.where(channel: current_channel)
+    render(json: @crypto_addresses_for_channel)
+  end
+
+  def generate_nonce
+    set = [('a'..'z'), ('A'..'Z'), (0..9)].map(&:to_a).flatten
+    nonce = (0...32).map { set[rand(set.length)] }.join
```
bcaller, line 17: Abusing the random number generation here is very high complexity, but still whenever we are generating random numbers for security purposes we should use a cryptographically secure random number generator. @kdenhartog do we want to recommend [Ruby's securerandom](https://ruby-doc.org/stdlib-2.4.0/libdoc/securerandom/rdoc/SecureRandom.html)? e.g. `SecureRandom.hex(16)`.

app/controllers/crypto_address_for_channels_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+    signature = params[:transaction_signature]
+    account_address = params[:account_address]
+    chain = params[:chain]
+    message = params[:message]
+    current_channel = current_publisher.channels.find(params[:channel_id])
+    @errors = []
+
+    # check that the message is a valid nonce and delete after use
+    valid_message = if Rails.cache.read(message)
+      !!Rails.cache.delete(message)
```
bcaller, line 32: Probably a race condition allowing multiple use here, but no real impact so I'd ignore it.

app/controllers/crypto_address_for_channels_controller.rb
jlbyrne: The Eth library I'm using for ruby on the back end seems to do something to hex strings internally, and will always return false on verification when I use `SecureRandom`.  I talked about this with @kdenhartog, and I think this is random enough for our purposes.  The ruby libraries have just been a nightmare.

kdenhartog: When I was looking at that, I was thinking we'd do something like `nonce = Base58Encode(SecureRandom(16))`. This is just pseudocode and I'm not sure of the specific syntax to achieve this in Ruby off the top of my head. 

Ultimately, we do want to make sure we're using something sufficiently random with a CSPRG like SecureRandom still (at least 128 bits of entropy is ideal), but it's fine to use a different encoding other than hex encoding.

app/controllers/crypto_address_for_channels_controller.rb
kdenhartog: If this cache is reused across all channels then this could be a problem here where a replay attack could be used to perform an authentication bypass that allows one creator to steal another creators signed message and replay it before they are able to submit it which would then link the address to the wrong creator.

This can be avoided by ideally linking the nonce to a specific creator such that even if the message were relayed we'd only link the address to the original account that requested the nonce.

app/controllers/crypto_address_for_channels_controller.rb
kdenhartog: Looks like `SecureRandom.uuid` would be an ideal option here if you want to remove any sort of encoding confusion.

package.json
tsmartt: We should version lock whenever possible

app/controllers/crypto_address_for_channels_controller.rb
```diff
@@ -0,0 +1,115 @@
...
+
+  def index
+    current_channel = current_publisher.channels.find(params[:channel_id])
+    @crypto_addresses_for_channel = CryptoAddressForChannel.where(channel: current_channel)
+    render(json: @crypto_addresses_for_channel)
+  end
+
+  def generate_nonce
+    set = [("a".."z"), ("A".."Z"), (0..9)].map(&:to_a).flatten
+    nonce = (0...32).map { set[rand(set.length)] }.join
```
kdenhartog, line 17:
```suggestion
    nonce = SecureRandom.uuid()
```

kdenhartog: this should address the one above @jlbyrne

app/controllers/crypto_address_for_channels_controller.rb
```diff
@@ -0,0 +1,115 @@
...
+    Rails.cache.write(nonce, true)
```
kdenhartog, line 18:
```suggestion
    current_channel = current_publisher.channels.find(params[:channel_id])
    Rails.cache.write(nonce + ":" + current_channel, true)
```

app/controllers/crypto_address_for_channels_controller.rb
```diff
@@ -0,0 +1,115 @@
...
+  def create
+    signature = params[:transaction_signature]
+    account_address = params[:account_address]
+    chain = params[:chain]
+    message = params[:message]
+    current_channel = current_publisher.channels.find(params[:channel_id])
+    @errors = []
+
+    # check that the message is a valid nonce and delete after use
+    valid_message = if Rails.cache.read(message)
```
kdenhartog, line 31:
```suggestion
    valid_message = if Rails.cache.read(message + ":" + current_channel)
```

app/controllers/crypto_address_for_channels_controller.rb
jlbyrne: I've updated the secureRandom to use uuid instead of hex.  Rather than adding the channel id to the nonce, which is visible from the URL, I'm storing the publisher ID with the nonce on the back end, so no one but the correct logged in publisher can use the nonce.

kdenhartog (approved):
LGTM thanks for fixing it all up

PR #4068, opened 2023-06-14
Updates image path for OG image

mattwilliams85:
Relative paths won't work for an OG tag, so I updated the reference to point to a image url instead

Closes [Issue #201](https://github.com/brave-intl/creators-private-issues/issues/201)


jlbyrne (approved):
LGTM!

PR #4069, opened 2023-06-14
Handle for missing promo registration elements 

mattwilliams85:
Not sure this code is relevant anymore, but this prevents an errors being throw if the elements don't exist on the page.

Closes [Issue #391](https://github.com/brave-intl/creators-private-issues/issues/391)


jlbyrne (approved):
LGTM!

PR #4071, opened 2023-06-16
Fixes styling on swoop to be more responsive

mattwilliams85:
Resolves https://github.com/brave-intl/creators-private-issues/issues/448

Updates swoop to be more responsive on all screen-sizes

# 1200px Breakpoint
## Before
Screenshot 2023-06-16 at 1 43 51 PM

## After
Screenshot 2023-06-16 at 1 44 03 PM

# 1200px Breakpoint
## Before
Screenshot 2023-06-16 at 1 44 35 PM

## After
Screenshot 2023-06-16 at 1 44 22 PM


mattwilliams85:
@jlbyrne just wondering, should we `.gitignore` the static minified css files? Usually those are created during the build phase so we don't have to check them in.

jlbyrne:
> @jlbyrne just wondering, should we `.gitignore` the static minified css files? Usually those are created during the build phase so we don't have to check them in.

Yes! That sounds like a good idea.

jlbyrne (approved):

PR #4073, opened 2023-06-16
Removes/ignores the public/static directory

mattwilliams85:
Resolves https://github.com/brave-intl/creators-private-issues/issues/1644

### Summary
Removes `public/static`  directory that is created during build-time 
Adds `public/static` to `.gitignore`
Adds `yarn build` command to `test.yml`

**Bonus:** I also updated several files in the `creators-landing` app to fix lint error warnings during the `yarn build` process

### Reason
To prevent unnecessary files from being committed when making changes to the React App


mattwilliams85:
@tsmartt would like to get your thoughts on this PR when you get back from PTO

tsmartt (approved):

PR #4077, opened 2023-06-21
Replaces static count with verified channels count to landing page

mattwilliams85:
## Summary 
Replace's the static `1,000,000` count in the sign-off section of the landing page with a count of verified channels from the database

Removes original code  for the PublicChannelsController since it is missing the associated file to work correctly

**Note:** The test DB only has 54 verified channels
Screenshot 2023-06-21 at 1 14 24 PM


jlbyrne (approved):
Yay, test coverage!

PR #4100, opened 2023-07-10
Exclude suspended publisher channels

mattwilliams85:
Exclude suspended channels from the CDN in the same way we do for users who are not in a supported region (i.e., do not include "wallet info").

Resolves: https://github.com/brave-intl/creators-private-issues/issues/1599


app/models/publisher.rb
```diff
@@ -473,7 +473,7 @@ def last_supported_login_locale
   end
 
   def brave_payable?
-    selected_wallet_provider&.payable?
+    selected_wallet_provider&.payable? && active? && !only_user_funds?
```
tsmartt, line 476: Does `brave_payable?` get called in the prefix response? Can we add a test to ensure that these values don't make it through?

tsmartt (changes requested):

app/models/publisher.rb
mattwilliams85: looks like `brave_payable?` isn't really called anywhere except for promo registration...

Would it make more sense to put this logic in the wallet state checks?
```
  def get_uphold_wallet_state(uphold_connection:)
    if uphold_connection.is_member && uphold_connection.address.present?
      PublishersPb::UpholdWalletState::UPHOLD_ACCOUNT_KYC
    else
      PublishersPb::UpholdWalletState::UPHOLD_ACCOUNT_NO_KYC
    end
  end

  def get_bitflyer_wallet_state(bitflyer_connection:)
    PublishersPb::BitflyerWalletState::BITFLYER_ACCOUNT_KYC
  end

  def get_gemini_wallet_state(gemini_connection:)
    if gemini_connection.payable?
      PublishersPb::GeminiWalletState::GEMINI_ACCOUNT_KYC
    else
      PublishersPb::GeminiWalletState::GEMINI_ACCOUNT_NO_KYC
    end
  end
```

Or maybe we could call `brave_payable?` in the `generate_brotli_encoded_channel_response` method and skip the entry if it comes back `false`

tsmartt (approved):

PR #4165, opened 2023-08-03
Add background job to create site banner for all channels, remove ref…

jlbyrne:
Add background job to create site banner for all channels, remove references to default banner for publisher, remove banner toggle on the front end

### DIFF
--- a/lib/tasks/database_updates/create_unique_site_banner_details_for_each_channel.rake
+++ b/lib/tasks/database_updates/create_unique_site_banner_details_for_each_channel.rake
@@ -1,0 +1,20 @@
+namespace :database_updates do
+  desc "Create a unique site banner detail row for each channel that shares one on the publisher level"
+  task create_unique_site_banner_details_for_each_channel: :environment do
+    could_not_update = []
+    channels = Channel.where.missing(:site_banner).pluck(:id)
+    channel_count = channels.count
+    puts "#{channels.count} channels have no site banner associated."
+
+    channels.each_slice do |channel_chunk|
+      CreateChannelBannersFromDefaultSiteBanners.perform_later(channel_chunk)
+    end
+
+    new_channel_count = Channel.where.missing(:site_banner).count
+
+    puts "Attempted to update #{channel_count} channels without a site banner. #{could_not_update.length} channels could not be updated."
+    puts "#{new_channel_count} channels are left with no site banner associated."
+    puts "Channel ids:"
+    puts could_not_update
+  end
+end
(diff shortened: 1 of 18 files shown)


lib/tasks/database_updates/create_unique_site_banner_details_for_each_channel.rake
```diff
@@ -0,0 +1,43 @@
+namespace :database_updates do
+  desc "Create a unique site banner detail row for each channel that shares one on the publisher level"
+  task create_unique_site_banner_details_for_each_channel: :environment do
+    could_not_update = []
+    channels = Channel.where.missing(:site_banner)
+    channel_count = channels.count
+    puts "#{channels.count} channels have no site banner associated."
+
+    channels.each do |channel|
```
tsmartt, line 9: Might want to do this in chunks to not run out of memory, since there could be a lot of channels. See [here](https://github.com/brave-intl/publishers/blob/staging/lib/tasks/database_updates/backfill_publisher_selected_wallet.rake#L43)

lib/tasks/database_updates/create_unique_site_banner_details_for_each_channel.rake
```diff
@@ -0,0 +1,43 @@
...
+      if default_site_banner
+        new_site_banner = SiteBanner.new(default_site_banner.attributes.except("id"))
+        new_site_banner.channel_id = channel.id
+
+        # Copy the attached images (logo and background_image) if they exist
+        if default_site_banner.logo.attached?
+          new_site_banner.logo.attach(default_site_banner.logo.blob)
+        end
+        if default_site_banner.background_image.attached?
+          new_site_banner.background_image.attach(default_site_banner.background_image.blob)
```
tsmartt, line 22: This is going to potentially do a lot of traffic to s3. Perhaps the core logic here can be extracted to a sidekiq job that we can launch async so all of that can be done independently and then this job can execute fairly quickly.

Dockerfile.dev
```diff
@@ -58,5 +58,7 @@ RUN apt-get install chromium --fix-missing -y
 COPY . .
 RUN cd public/creators-landing && yarn install && yarn build
 
+RUN NODE_OPTIONS=--openssl-legacy-provider RAILS_ENV=development CREATORS_FULL_HOST="1" SECRET_KEY_BASE="1" bundle exec rails assets:precompile
```
tsmartt, line 61: I think we'll want to leave the env production here for compiling, even in the dev docker since the prod env is set up to correctly compile and that's the only ENV we actually compile in.

tsmartt: So then this just becomes a sanity check when developing.

tsmartt (changes requested):
Thanks for getting this done!

tsmartt (approved):

PR #4183, opened 2023-08-16
Remove Legacy Youtube login

tsmartt:
Closes https://github.com/brave-intl/creators-private-issues/issues/1686


mattwilliams85:
We should probably remove the translations 
Here https://github.com/brave-intl/publishers/blob/cf8c5b8d234b040c7015224c9bffa923d9391e40/config/locales/en.yml#L598
and here https://github.com/brave-intl/publishers/blob/cf8c5b8d234b040c7015224c9bffa923d9391e40/config/locales/ja.yml#L568

mattwilliams85 (approved):

PR #4191, opened 2023-08-23
Feat/next js integration

tsmartt:
Initial integration of NextJS app and Publisher Rails.

Inspired by https://github.com/naofumi/rails-nextjs-erb/wiki/How-It-Works

We can't use that exact method since we're using the new Next routing framework, but we can have all fetch requests pass along the session cookie until we finally move login over to the SPA

First few commits non-SSL, final SSL enabled.


mattwilliams85 (approved):
Great work! LGTM

PR #4199, opened 2023-09-06
Only keep gemini enabled for existing users

tsmartt:
Closes https://github.com/brave-intl/creators-private-issues/issues/1689


tsmartt:
No longer needed

PR #4225, opened 2023-09-28
Reintroduce express to proxy to rails app

tsmartt:
This express proxy will take care of rewriting redirects coming back from rails which will redirect to itself. We want instead to redirect back to our Next proxy.


thypon:
Is there any flow diagram of the next/rails forwarding?

Cc @bcaller

mattwilliams85 (approved):
image
Looks good to me!

PR #4236, opened 2023-10-04
Fix/webauthn verify with proxy

tsmartt:
Using our proxy, we can still have webauthn requests verified with no need to update yubikeys.


app/views/static/index.html
mattwilliams85, line 1: We've been having to rebuild this file a lot lately, not sure whats going on 🤔

mattwilliams85 (approved):
LGTM

jlbyrne (approved):

PR #4241, opened 2023-10-06
Add BAT as a payment option for the channel page

jlbyrne:


tsmartt (approved):
We'll need to make it pluggable in a follow-up PR

PR #4321, opened 2024-01-23
Hardcode meta tags for now, until GH action fixed

tsmartt:
Closes https://github.com/brave-intl/creators-private-issues/issues/1746


jlbyrne (approved):

bcaller (approved):
Yeah I don't think we need subresource integrity here

PR #4364, opened 2024-02-15
Fix/next ssl in dev only

tsmartt:
Tweaks needed to get the Next app to relay correctly in the dev environment.

### DIFF
--- a/nextjs/scripts/create-local-server.js
+++ b/nextjs/scripts/create-local-server.js
@@ -45,7 +45,7 @@ app
         logger: console,
         target: pubHost,
         changeOrigin: true,
-        secure: !isDevelopment,
+        secure: !dev,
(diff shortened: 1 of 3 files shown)


nextjs/scripts/create-local-server.js
```diff
@@ -45,8 +44,10 @@ app
         logger: console,
         target: pubHost,
         changeOrigin: true,
-        secure: !isDevelopment,
+        secure: !dev,
         onProxyReq: (proxyReq, request, response) => {
+          const ip = (request.headers['x-forwarded-for'] || request.socket.remoteAddress).split(':').pop()
+          proxyReq.setHeader('originalIP', ip );
```
tsmartt, line 50: This is so the Next app will tell the rails app the original IP for admin verification.

jlbyrne (approved):

PR #4375, opened 2024-03-01
fix dev server updates and icon asset paths

jlbyrne:


tsmartt (approved):

tsmartt (changes requested):
Given recent PRs, lets' discuss

PR #4387, opened 2024-03-14
Use the FULL host for U2F since the base_url is different now that ra…

tsmartt:
Closes https://github.com/brave-intl/creators-private-issues/issues/1794

### DIFF
--- a/.audit-ci.json
+++ b/.audit-ci.json
@@ -37,6 +37,7 @@
     "GHSA-m95q-7qp3-xv42",
     "GHSA-7fh5-64p2-3v2j",
     "GHSA-78xj-cgh5-2h22",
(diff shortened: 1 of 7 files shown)


jlbyrne (approved):

diracdeltas:
You can ignore the Socket error but please fix the `yarn audit` problem since that should be easy to do by upgrading follow-redirects: https://github.com/advisories/GHSA-cxjh-pqwp-8mfp

tsmartt:
Fixed in main app, for the landing page, since it came from the webpack-dev-server, ignored since that whole build pipeline is ancient and about to be replaced in our UI rewrite with a NextJS modern system.

PR #4393, opened 2024-03-19
Feat/migrate dashboard

jlbyrne:


nextjs/src/lib/api.ts
```diff
@@ -11,7 +11,7 @@ export async function apiRequest(
     const response = await axios({ method, url, data });
 
     // verify request had 2xx status code
-    if (response.status !== 200) {
+    if (!(response.status >= 200 && response.status <= 299)) {
```
thypon, line 14: taka a second dynamic look

app/controllers/publishers/omniauth_callbacks_controller.rb
```diff
@@ -297,5 +297,9 @@ def require_publisher_not_created_through_youtube_auth
         redirect_to(home_publishers_path)
       end
     end
+
+    def redirect_options
```
thypon, line 301: verify this dynamically

app/controllers/api/nextv1/connection/bitflyer_connections_controller.rb
```diff
@@ -0,0 +1,57 @@
+require "uri"
+require "net/http"
+require "json"
+require "digest"
+require "base64"
+
+class Api::Nextv1::Connection::BitflyerConnectionsController < Api::Nextv1::Oauth2Controller
```
thypon, line 7: is this new or moved?

app/controllers/api/nextv1/oauth2_controller.rb
```diff
@@ -0,0 +1,114 @@
+class Api::Nextv1::Oauth2Controller < Api::Nextv1::BaseController
```
thypon, line 1: moved?

nextjs/src/lib/api.ts
jlbyrne: This change was necessary because the less flexible code would kick 201 and 204 responses and the like to the error handler.

app/controllers/api/nextv1/connection/bitflyer_connections_controller.rb
jlbyrne: This is all moved.  The originals can be found here: [app/controllers/payment/connection/bitflyer_connections_controller.rb](https://github.com/brave-intl/publishers/blob/staging/app/controllers/payment/connection/bitflyer_connections_controller.rb)

app/controllers/api/nextv1/oauth2_controller.rb
jlbyrne: yes, original here: [app/controllers/oauth2_controller.rb](https://github.com/brave-intl/publishers/blob/staging/app/controllers/oauth2_controller.rb)

app/controllers/publishers/omniauth_callbacks_controller.rb
jlbyrne: removed- it was a mistake that this change was included initially.

thypon (approved):

PR #4434, opened 2024-05-08
add usdc options

jlbyrne:
Resolves https://github.com/brave-intl/creators-private-issues/issues/1804


tsmartt (commented):
Look good but tests are failing. And is it possible to add a unit test here?

tsmartt (approved):

PR #4487, opened 2024-07-25
First pass at getting new UI testing down

tsmartt:
Resolves https://github.com/brave-intl/creators-private-issues/issues/1859

This PR sets up a separate set of Capybara tests to hit the new UI written in NextJS. 
We add a test runner script that will run the old tests as well as the new UI tests. 
They must be run separately as we redirect Capybara to an external host which breaks a bunch of the old tests.
Using this separate runner we can piecemeal port the old capybara tests over to the new UI.


tsmartt:
This is ready to go except for an elliptic check which just popped up

See https://github.com/indutny/elliptic/pull/317 for more

Will make a separate PR for that before merging this

jlbyrne (approved):

kdenhartog:
From what I can tell, this is occurring because we need crypto-browserify>browserify-sign>elliptic as a webpack replacement for node crypto. However, it's not clear where this is actually in use when I did a quick search through the various JS code.

AFAICT we're only producing signatures on the JS side and that's done by calling into the wallet scripts injected by the wallet (which means the wallet code of the user is doing the signing not our frontend code), but not verifying them with JS. I did check through where we're actually doing verification of signatures and it appears to be using ruby instead so I don't believe we're hitting the vulnerable code paths. If for some reason you find that we are using `crypto-browserify` in someway and that is performing signature verification then it would present a small issue but not one that I think is feasibly exploitable. 

So, I'm alright if we just ignore this update until the patch to the elliptic library is accepted and propagated through to crypto-browserify. Alternatively, since it doesn't seem like crypto-browserify is actually in use we could also look at dropping the dependency all together.

tsmartt:
I looked at dropping crypto-browserify, and to do that we need to move to solana's new v2 webjs library. It's currently in RC so let's keep an eye on that. 

https://github.com/solana-labs/solana-web3.js/releases

For now I'll ignore this particular CVE. Thanks!

kdenhartog:
Ok, if only solana's web3.js is using crypto-browersify I'm confident this doesn't affect us.

PR #4494, opened 2024-08-16
Add contribution banner customizer, migrate public channel page files

jlbyrne:
Resolves: https://github.com/brave-intl/creators-private-issues/issues/1603

image

image


Miyayes:
@jlbyrne Should hide the "Add account" on GitHub, since we no longer support adding GH.

tsmartt (approved):

PR #4502, opened 2024-09-12
get rid of typo in image upload and unnecessary file extensions

jlbyrne:
gets rid of additional `.` character in the content type, gets rid of the `.padded` extension in the filename, and removes `.jpg.pad` from the s3 request.


porteron (approved):
The code looks fine to me, but we should get more information on the purposes of having padded images here before merging.

porteron (approved):

PR #4518, opened 2024-10-16
Fix/image filetype

jlbyrne:
Resolves https://github.com/brave-intl/creators-private-issues/issues/1873

Changes outside of `app/controllers/api/nextv1/contribution_page_controller.rb` are all for dependency updates and knock on changes.  Several security updates required me to update node, which created some linting issues.  I changed some tsx files to jsx files to avoid needing to fix those.  I also deleted the crypto-browserify package, since its dependency, elliptic, doesn't have a patch and we aren't using the main package anymore.  

Updating the minor version of rails also created more linting and upgrade issues, so some random ruby files were changed and included in this PR as well.


porteron (approved):
lgtm

PR #4709, opened 2025-04-14
Feat/move to context for widget

jlbyrne:
Resolves https://github.com/brave-intl/creators-private-issues/issues/1897
Resolves https://github.com/brave-intl/creators-private-issues/issues/1900


nextjs/src/app/[locale]/c/[public_identifier]/CryptoWidgetAmountSelect.jsx
```diff
@@ -0,0 +1,303 @@
...
+
+  return (
+    <>
+      {/*Chain dropdown*/}
+      <Select
+        options={dropdownOptions}
+        onChange={changeChain}
+        components={{
+          SingleValue: ({ children, ...rest }) => (
+            <components.SingleValue {...rest} className="flex">
```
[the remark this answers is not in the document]
Comment (jlbyrne) [reply nextjs/src/app/[locale]/c/[public_identifier]/CryptoWidgetAmountSelect.jsx:173]: This is just cutting and pasting existing code into a new file as part of the refactor.  Also, the `...rest` is coming from the select dropdown library's docs for customizing its components.  It has a bunch of internal props that I'm not sure are explicitly listed and would be tricky and annoying to find and include here.

nextjs/src/app/[locale]/c/[public_identifier]/CryptoWidgetPaymentButton.jsx
```diff
@@ -0,0 +1,357 @@
...
+    setIsSuccessView,
+    setErrorMsg,
+    setErrorTitle,
+    setIsTryBraveModalOpen,
+  } = useContext(CryptoWidgetContext);
+
+  async function sendPayment() {
+    clearError();
+    switch (currentChain) {
+      case "ETH":
```
Comment (porteron) [inline nextjs/src/app/[locale]/c/[public_identifier]/CryptoWidgetPaymentButton.jsx:59]: The switch statement is fine but you might consider storing this in a data structure as a mapping. e.g.) `{"ETH": sendEthPayment, "SOL": sendSolPayment, ...}`

Comment (jlbyrne) [reply nextjs/src/app/[locale]/c/[public_identifier]/CryptoWidgetPaymentButton.jsx]: I like it!  Much shorter.

PR #4819, opened 2025-05-14
Feat/admin OFAC button

jlbyrne:
Resolves https://github.com/brave-intl/creators-private-issues/issues/1904

Includes changes needed to test on staging for https://github.com/brave-intl/creators-private-issues/issues/1903

### DIFF
--- a/app/controllers/api/nextv1/crypto_address_for_channels_controller.rb
+++ b/app/controllers/api/nextv1/crypto_address_for_channels_controller.rb
@@ -42,8 +42,11 @@ class Api::Nextv1::CryptoAddressForChannelsController < Api::Nextv1::BaseControl
         false
       end
 
+    # check to make sure the address is not sanctioned
+    sanctioned = sanctioned?(account_address)
+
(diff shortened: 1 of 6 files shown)


app/controllers/api/nextv1/crypto_address_for_channels_controller.rb
```diff
@@ -94,4 +99,22 @@ def replace_crypto_address_for_channel(account_address, chain, channel)
     LogException.perform(e, publisher: current_publisher)
     false
   end
+
+  def sanctioned?(address)
+    bucket_name = 'brave-production-ofac-addresses'
```
porteron, line 104: maybe move this to an env var

docker-compose.yml
```diff
@@ -24,6 +29,7 @@ services:
     volumes:
       - .:/var/www
       - /var/www/node_modules
+      - $HOME/.aws:/root/.aws
```
porteron, line 32: Same, with the iam auth you would not need to do this.

porteron (approved):

PR #4820, opened 2025-05-14
specify rp_id

jlbyrne:
Resolves https://github.com/brave-intl/creators-private-issues/issues/1914

Fixes a bug introduced in https://github.com/brave-intl/publishers/pull/4753.  Reading through the webauthn code, we shouldn't need to specify an rp_id (it should read it from the `domain` variable we are already supplying), but apparently it's necessary.  Also removing the u2f check, since there's no longer a difference in the verify methods.

### DIFF
--- a/app/services/two_factor_auth/webauthn_verify_service.rb
+++ b/app/services/two_factor_auth/webauthn_verify_service.rb
@@ -26,22 +26,13 @@ module TwoFactorAuth
       )
 
       begin
-        if registration.u2f? # old registration format
-          assertion_response.verify(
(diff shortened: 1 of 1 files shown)


porteron (approved):

jlbyrne:
> The security team is monitoring all repositories for certain keywords. This PR includes the word(s) "authn" and so security team members have been added as reviewers to take a look. No need to request a full security review at this stage, the security team will take a look shortly and either clear the label or request more information/changes. Notifications have already been sent, but if this is blocking your merge feel free to reach out directly to the security team on Slack so that we can expedite this check.

This isn't expanding or changing security features, this is just a bugfix for an existing feature.

PR #5299, opened 2026-03-16
update change_address endpoint and tests

jlbyrne:
update change_address endpoint to check for publisher ownership of address.
Resolves https://github.com/brave-intl/creators-private-issues/issues/1945


porteron (approved):
lgtm

PR #5595, opened 2026-06-25
fix omniauth issues and add fallback for promo registrations error

jlbyrne:


jlbyrne:
@SocketSecurity ignore gem/json@2.20.0

DJAndries (approved):

PR #5603, opened 2026-06-25
remove references to github channels

jlbyrne:
Resolves https://github.com/brave-intl/creators-private-issues/issues/1915


app/controllers/publishers/omniauth_callbacks_controller.rb
```diff
@@ -181,31 +181,6 @@ def register_reddit_channel
       redirect_to home_publishers_path, notice: t("shared.channel_created")
     end
 
-    # def register_github_channel
```
DJAndries, line 184: nit: any reason why we can't remove instead of commenting it out? consider adding a TODO if you wish to keep it as is

DJAndries (approved):

app/controllers/publishers/omniauth_callbacks_controller.rb
jlbyrne: It was commented out, now removed

PR #5716, opened 2026-07-29
override sharp and postcss

jlbyrne:
Addresses https://github.com/brave-intl/publishers/security/dependabot/573
https://github.com/brave-intl/publishers/security/dependabot/572
https://github.com/brave-intl/publishers/security/dependabot/559


porteron (approved):
lgtm

PR #5750, opened 2026-08-09
update packages and remove tslint

jlbyrne:
Fixes:
https://github.com/brave-intl/publishers/security/dependabot/591
https://github.com/brave-intl/publishers/security/dependabot/593
https://github.com/brave-intl/publishers/security/dependabot/590
https://github.com/brave-intl/publishers/security/dependabot/589
https://github.com/brave-intl/publishers/security/dependabot/588


config/shakapacker.yml
```diff
@@ -115,17 +66,44 @@ development:
...
+  # Extra module lookup paths; must be an array even if empty because
+  # shakapacker spreads it without an undefined guard (rules/jscommon.js).
+  additional_paths: []
+  compiler_strategy: digest
+  compile: false
 
 production:
-  <<: *default
-
+  source_path: app/javascript
```
porteron, line 90: I'm not super familiar with this config setup but seems ok to me

porteron (approved):

config/shakapacker.yml
jlbyrne: in previous versions of js-yaml the `<<: *{variable}` setup was allowed, in later ones it's not.  these are the same settings, just more verbose because you can't just stick all the common ones in *default.

PR #5753, opened 2026-08-10
fix: make reserved public name lookup case-insensitive

kdenhartog:
Closes https://github.com/brave-intl/creators-private-issues/issues/1958

## Summary

`Channel#validate_public_url_unique` checked the `reserved_public_names` table with a case-sensitive `find_by(public_name:)` (`app/models/channel.rb:546`), while every other check uses `LOWER()`.

This mismatch let a verified creator bypass a reservation by flipping one letter's cases leading to redirecting tips to a different wallet address.

## Fix

Use `LOWER()` in the `find_by`, matching the case-insensitive behavior of the rest of the method. This handles both existing mixed-case rows and new ones without a data migration.

## Test

Adds `test/models/reserved_public_name_case_test.rb` asserting that both the exact-case and case-flipped forms of a reserved name are rejected. All 86 existing model tests continue to pass.

## Checklist

- [x] Production code fix (logic correction, not defense-in-depth)
- [x] Test added covering the bypass
- [x] Existing tests pass
- [ ] No migration required (lookup fix handles existing data)

### DIFF
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -543,7 +543,7 @@ class Channel < ApplicationRecord
       errors.add(:public_name, "must only contain letters, numbers, dashes, and underscores")
     end
 
-    reserved = ReservedPublicName.find_by(public_name: public_name)
+    reserved = ReservedPublicName.find_by("LOWER(public_name) = ?", public_name.downcase)
     # names previously in use are released after 1 year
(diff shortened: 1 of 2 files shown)


jlbyrne (approved):

jlbyrne:
Audit is failing because of vulnerabilities addressed in another PR

PR #5756, opened 2026-08-10
Chore/update redis and redis session store

jlbyrne:
Rehash of https://github.com/brave-intl/publishers/pull/5746

redis-session-store will only work with redis version 5 or lower and doesn't appear to be actively maintained.  It was introduced here (https://github.com/brave-intl/publishers/pull/4025) as a replacement for activerecord-session_store.  redis-actionpack is actively maintained and seems closest in functionality to redis-session-store.


porteron (approved):
lgtm

PR #5765, opened 2026-08-11
Revert "Prod Release 07-29-26"

jlbyrne:
Reverts brave-intl/publishers#5717

Release was squashed and merged rather than just being merged


porteron (approved):

jlbyrne:
Ignoring security alerts because I'm reverting, then immediately re-adding all the changes here to clean up the master branch

